PR #25, opened 2015-07-17
Add XKB_CONFIG_ROOT environment

mvo5:
Trivial patch that allows overriding the DFLT_XKB_CONFIG_ROOT path via a new XKB_CONFIG_ROOT environment to make it possible to set the path at runtime instead of build-time (to make  binaries build with libxkbcommon easier relocatable).

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -75,12 +75,16 @@ err:
 XKB_EXPORT int
 xkb_context_include_path_append_default(struct xkb_context *ctx)
 {
-    const char *home;
+    const char *home, *root;
     char *user_path;
     int err;
     int ret = 0;
 
-    ret |= xkb_context_include_path_append(ctx, DFLT_XKB_CONFIG_ROOT);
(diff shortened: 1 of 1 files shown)


bluetech:
I'm not a big fan of environment variables, but we have precedent, and this seems logical, so I'm OK with it. Just two comments:

1. Please use `secure_getenv`.
2. Please add a note about the effect of this envvar. Hmm actually, the docs could use some expanding about the default include path anyway, so you can leave this to me.

Thanks.

mvo5:
Thanks a lot for you quick reply! I updated the code to use secure_getenv() and also moved the *root declaration up next to *home to match the existing code better.

I'm also happy to help with the documentation for the new environment, I poked a bit around and could not quite find a place to start so I'm happy to leave it to you. But please let me know if I can help in any way.

bluetech:
Thanks!

I'll add a section to the documentation about the include path handling soon, and include a note about `XKB_CONFIG_ROOT`.

PR #27, opened 2015-08-24
Modifications to make 'make check' test complete for Mac OS X

nuko8:
This series of patches consists of two commits.

Although each of the commits may be handled independently, I'm placing them in a single PR because I think they are closely related in terms of the `make check` test on a specific platform, as suggested in the title.

On Mac OS X, the `make check` test cannot be carried out because

(1) Since `clock_gettime()` is not available on the OS, all the programs `bench/*.c` won't compile.

(2) When XQuartz's `Xvfb` is given the option `-displayfd`, it is observed that it always returns `:0`.   As a consequence, `x11comp` uses that (default) display number to connect to the X server and closes it when it's finished.  Obviously, that leaves the window system in an undesirable state.

Commit 4627892 handles the first issue and Commit 18e6612 the second.

With the patches proposed, `make check` can build the programs `bench/*.c` successfully and carry out the test suite with clean exit state.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -250,8 +250,12 @@ check_PROGRAMS += $(TESTS)
 ##
 # Benchmarks
 ##
-
-BENCH_LDADD = $(TESTS_LDADD) $(RT_LIBS)
+check_LTLIBRARIES += libbench.la
+libbench_la_SOURCES = \
+	bench/bench.c \
+	bench/bench.h
+BENCH_LDADD  = libbench.la
+BENCH_LDADD += $(TESTS_LDADD) $(RT_LIBS)
 
 check_PROGRAMS += \
 	bench/key-proc \
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -1,0 +1,148 @@
+/*
+ * Copyright © 2015 Kazunobu Kuriyama <kazunobu.kuriyama@nifty.com>
+ *                  Ran Benita <ran234@gmail.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+#if defined(HAVE_CLOCK_GETTIME)
+#define USE_CLOCK_GETTIME
+#include <time.h>
+#elif defined(__MACH__) && __MACH__ == 1
+#define USE_MACH_ABSOLUTE_TIME
+#include <mach/mach_time.h>
+#else
+/* gettimeofday() - a last resort */
+#include <sys/time.h>
+#endif
+
+#include <stdlib.h>
+#include <string.h>
+#include <stdio.h>
+
+#include "bench.h"
+
+static void
+set_bench_time(struct bench_time *dest, long seconds, long milliseconds)
+{
+    dest->seconds = seconds;
+    dest->milliseconds = milliseconds;
+}
+
+static void
+normalize_bench_time(struct bench_time *obj)
+{
+    if (obj->milliseconds >= 0) {
+        return;
+    }
+    obj->milliseconds += 1000000;
+    obj->seconds--;
+}
+
+void
+bench_timer_reset(struct bench_timer *self)
+{
+#if defined(USE_MACH_ABSOLUTE_TIME)
+    mach_timebase_info_data_t info;
+    if (mach_timebase_info(&info) == 0) {
+        self->scaling_factor = info.numer / info.denom;
+    }
+#endif
+    self->start.seconds = 0L;
+    self->start.milliseconds = 0L;
+    self->stop.seconds = 0L;
+    self->stop.milliseconds = 0L;
+}
+
+void
+bench_timer_start(struct bench_timer *self)
+{
+#if defined(USE_CLOCK_GETTIME)
+    struct timespec val;
+
+    (void)clock_gettime(CLOCK_MONOTONIC, &val);
+
+    /* With conversion from nanosecond to millisecond */
+    set_bench_time(&self->start, val.tv_sec, val.tv_nsec / 1000);
+#elif defined(USE_MACH_ABSOLUTE_TIME)
+    uint64_t val;
+
+    val = mach_absolute_time();
+
+    /* With conversion from nanosecond to millisecond */
+    set_bench_time(&self->start,
+                   self->scaling_factor * val / 1000000000,
+                   self->scaling_factor * val % 1000000000 / 1000);
+#else
+    struct timeval val;
+
+    (void)gettimeofday(&val, NULL);
+
+    set_bench_time(&self->start, val.tv_sec, val.tv_usec);
+#endif
+}
+
+void
+bench_timer_stop(struct bench_timer *self)
+{
+#if defined(USE_CLOCK_GETTIME)
+    struct timespec val;
+
+    (void)clock_gettime(CLOCK_MONOTONIC, &val);
+
+    /* With conversion from nanosecond to millisecond */
+    set_bench_time(&self->stop, val.tv_sec, val.tv_nsec / 1000);
+#elif defined(USE_MACH_ABSOLUTE_TIME)
+    uint64_t val;
+
+    val = mach_absolute_time();
+
+    /* With conversion from nanosecond to millisecond */
+    set_bench_time(&self->stop,
+                   self->scaling_factor * val / 1000000000,
+                   self->scaling_factor * val % 1000000000 / 1000);
+#else
+    struct timeval val;
+
+    (void)gettimeofday(&val, NULL);
+
+    set_bench_time(&self->stop, val.tv_sec, val.tv_usec);
+#endif
+}
+
+void
+bench_timer_get_elapsed_time(struct bench_timer *self, struct bench_time *result)
+{
+    result->seconds = self->stop.seconds - self->start.seconds;
+    result->milliseconds = self->stop.milliseconds - self->start.milliseconds;
+}
+
+char *
+bench_timer_get_elapsed_time_str(struct bench_timer *self)
+{
+    struct bench_time elapsed;
+    char *buf;
+
+    bench_timer_get_elapsed_time(self, &elapsed);
+    normalize_bench_time(&elapsed);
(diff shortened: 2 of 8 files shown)


nuko8:
> I do not like all of the ifdefs this adds.
> But again I don't like all of the ifdefs.

Neither do I :-)  So, I'm rather happy to hear those comments.  

I was wrongly assuming you might prefer to keeping the original code as possible so that the peculiarities of Mac OS X would be enclosed and localized.  That's why there are a bunch of ifdefs in the code I proposed.  Now I know you accept another way.

So, let me code again in such a way that resulting code is easier to read and maintain than the current one.  I'll do that based on your suggestions noted in the comments above.

Thank you for the valuable comments!

nuko8:
I've just pushed an updated version of the PR.  I'd be happy if you'd take a look at it.

bluetech:
I am not that familiar with the github workflow (haven't used much), but I'd be nice if you can squash the fixups into the 2 commits, that way we have a clean log after merging.

nuko8:
> I am not that familiar with the github workflow (haven't used much), but I'd be nice if you can squash the fixups into the 2 commits, that way we have a clean log after merging.

Neither, I am.  I am a new refugee all the way from google.code :-)

I also prefer to squashing the existing commits eventually into a single or two commits before merging, if it should happen.

That said, I want to make sure whether you mean I should squash commits every time I update the PR, or only once after you approve the PR and just before you really carry out the merging, though I’m not sure I can do that with `git push -f <my remote> <squashed branch>` or alike (I’m such a novice), and don’t know what are possible side-effects by that.

Though I’m ready to update the PR, it looks safer to put it off till I hear from you.

bluetech:
I guess either works. If the diff between versions is small, I guess an
additional commit is easier to review. But if the implementation changes
a lot, an incremental diff is harder.

Anyway, I'm mostly used to mailing-list patch review, where the series
is resent every time. So let's just squash and force push every time. But
if you want to do it at the end that's also cool.

nuko8:
> Why do we care about TCP though? I think the default for recent X now is not to listen on TCP, but if that's not the case on Mac OSX than we should pass -nolisten tcp to Xvfb; that's a good idea anyway I guess. Then we can use still use some high range where there's no "contention".

Oh, I should have mentioned UDP as well.  The port number registry by IANA for X clients is not only for TCP but also for UDP: The port range [6000, 6063] is assigned for both TCP and UDP connections.

By the vendor’s settings, XQuartz on my Mac is invoked with `-noliesten tcp` and a socket is created in the hidden directory `/tmp/.X11-unix`.  Its name is usually `X0`.  So, a similar argument I did based on TCP holds here.

I think the situation is more or less similar on linux boxes, and that the range [0, 63] is still reasonable as a range of allowable `DISPLAY` numbers. 

I do think that conservative range [0, 63] is sufficient because, simply, it's working.  Actually,  contention between `X11` and  `Xvfb` that was seen in the original test suite is now gone with the patched test suite.

bluetech:
I didn't know X had a UDP transport (`Xvfb -listen udp` fails for me).

I still don't understand why the ports or IANA are relevant here, since `Xvfb :<any 32bit number>` works fine here. But the conservative approach works, so OK by me.

nuko8:
> So let's just squash and force push every time. 

I did an experiment and confirmed that I could use `push -f` to replace a series of commits at `origin` with a single, squashed commit at a cloned repository.  So I guess it's technically possible to do a similar kind of push from a local to the public remote.

But if I did that, this page's links to each commit would almost surely be broken, and I'm uncertain what would happen to the page.

It looks github's issue tracking system is not friendly to history revision.   Certainly, without referring to each commit, the conversation pages are little or no use.

So, there's no way for me other than simply adding commits.  I hope you understand that.

nuko8:
> We usually use defines for constants. If it needs the type for the calculations we add an L to the literal.

Done.

> I don't see where UNDEF is used, so can be dropped?

No.  The value of UNDEF is implicitly used and `subsecond_unit` is set to that value in `bench_timer_init()`, and helps us tell not-used objects from used ones.

> Also, we don't use typedefs for structs/enums (besides for some inherited code or for descriptive purposes); instead we spell the whole name (e.g. struct xkb_keymap).
> Since this type is only used internally, there is no need for any encapsulation here, so it'd be better to just move the definition to the header file and use stack allocation.
> Also, we don't use typedefs for structs/enum.

The `struct` typedef is used to make the struct opaque, which purpose is to hide implementation details.  

Owing to that, you can write benchmark programs without paying any attention to platform dependency and other messy stuff such as the names of timer functions, their synopses, the definitions of `struct`s used, the names and types of `sturct`’s fields.  You only remember to use `*_new()’, `*_start()`, `*_stop()`, `*_get_elapsed_time()` and `*_release()`.

On the other hand, if you find something wrong or get an interesting idea with the timer module, all you need to do is to touch `bench.c` and `bench.h`, and there should be no need to examine the code of benchmark programs which uses the module.

So I think encapsulation of internal objects is still helpful even for the developers of libxkbcommon, in particular, non-member contributors.

By giving the `enum` a name, the valid value ranges of `subsecond_unit` and the second parameter of `normalize_bench_time()` are made clear.  If they were simply declared `int` or alike, those who read the code would scroll up and down to figure out what are the valid ranges of the entities.

So, I don’t see anything wrong with the `typedef` of `struct`s and `enum`s if they definitely have their own purposes and are used to fulfill them.

> Since this type is only used internally, there is no need for any encapsulation here, so it'd be better to just move the definition to the header file and use stack allocation.

Breaking the encapsulation deteriorates the purpose of the timer module.  All platform dependency is enclosed and hidden there.  This means, if someone wants to add something platform-dependent, she or he is naturally led to the module.  If the encapsulation were broken, no one could stop her or him to put platform-dependent code outside the module.  That deteriorates maintainability and makes future extensions hard.

> Should this say `self->stop`?

You're right.  Fixed.

> Maybe multiply this by 1000, and then just assume NANO and drop the enum?

My concerns here are the sizes of the integers contained in `struct`s used with time functions, format specifiers used with `printf()` family, and how to relate them.

`NANO` and `MICRO` are used to specify integer sizes and format specifiers and relate them each other by their names.  Multiplication won’t be helpful for those purposes.

> Missing space before the =.

Fixed.

> Hiding pointers behind is generally a bad idea IMO.

I have to admit that’s my fault, and I cannot remember why I did such a silly thing :-)  Fixed.

> I had a few comments, but besides that this is much better than the current situation!

Thank you for reading the code intensively and giving comments from various points of view!

nuko8:
I've just done force push to reduce the number of the existing commits to two.   Looks it's ok...

Sorry about the inconvenience so far and thank you for your patience.

nuko8:
Finally, I managed to find time to fulfill your requirements.

Currently, the number of the remaining issues is only one.

Below, progress of work follows each of your comments.

> We usually use defines for constants. If it needs the type for the calculations we add an L to the literal.

Done.

> I don't see where UNDEF is used, so can be dropped?

It was used implicitly in the previous implementation.  But, to make the intention clearer, it is now used explicitly, e.g., `bench_timer_reset()`.

> Since this type is only used internally, there is no need for any encapsulation here, so it'd be better to just move the definition to the header file and use stack allocation.

Done.

> Should this say self->stop?

Fixed through a new auxiliary function suggested by

> Maybe there can be a single function that takes the time, and then both start() and stop() call that?

It’s `set_bench_time()`.

> Maybe multiply this by 1000, and then just assume NANO and drop the enum?

Pending.

If the enum has to be dropped, I think there’re two options:

1. Use `#ifdef` in `bench_timer_get_elapsed_time_str()` to choose a path at compile time to select a format specifier among `“%ld.%09ld”` for nano and `“%ld.%06ld”` for micro.

2. Use the fixed number of decimal digits for representing elapsed time independently of its granularity, and determine what the fixed number is, e.g., 3, 6, or 9.

If you still want to drop the `enum`, let me know which option you prefer to.  I’ll code accordingly.  Other suggestions are also welcome.

> We can also go the other way; nanosecond granularity is really not statistically significant. (We should also add min/max/mean/median/stddev someday...).

Sorry, but I’m not sure of what to do with that for carrying out my current work here, because I haven’t heard anything from you about the test suite’s feature plan on the statistics functions mentioned.

If that is part of your requirement, I’d be happy if you would clarify it more explicitly.

N.B.  Because the `enum` goes to the header, the names `NANO` and `MICRO` are replaced with `BENCH_TIMER_NANO` and `BENCH_TIMER_MICRO`, respectively.

> Missing space before the =.

Fixed.

> Hiding pointers behind is generally a bad idea IMO. Pointer and non-pointer have different semantics when assigned or passed as parameter and can be unexpected when using const.

Void.

This issue was gone because the heap allocation for the `structs` is abandoned.

> I had a few comments, but besides that this is much better than the current situation!

Currently, “dropping the enum” is an only outstanding issue.  As I wrote above, your decisions, prefererences, and/or suggestions are needed to resolve the issue.

Additional feedbacks are also welcome, of course.

bluetech:
Hi @nuko8, sorry for the delay.

> So, I don’t see anything wrong with the typedef of structs and enums if they definitely have their own purposes and are used to fulfill them.

It is mostly a style issue, I suppose. In xkbcommon, we do not typedef structs and enums (you can see that in `xkbcommon/xkbcommon.h` we say `struct xkb_context *context`, `enum xkb_log_level level`, etc. The structs are all opaque BTW, a typedef is not needed for that). See also [here](http://yarchive.net/comp/linux/typedefs.html). 

So we should use `struct bench_timer`, without the typedef, in order to keep the same style everywhere.

> If you still want to drop the enum, let me know which option you prefer to. I’ll code accordingly. Other suggestions are also welcome.

Since the extra precision given by nanoseconds over microseconds is completely statistically insignificant when running ~20000 runs over ~10 seconds, my suggestion was to just change `subseconds` to `microseconds` (since it is the lowest common denominator), and make e.g. the `clock_gettime` path do `microseconds = val.tv_nsec / 1000`.  Once we always use microseconds, there is no need for the enum then. Does that make sense?

> Sorry, but I’m not sure of what to do with that for carrying out my current work here, because I haven’t heard anything from you about the test suite’s feature plan on the statistics functions mentioned.

Sorry, that was just some off-topic wishlist comment, inspired by [this](http://zedshaw.com/archive/programmers-need-to-learn-statistics-or-i-will-kill-them-all/) :) You don't need to do any of that (though that would be very welcome for a future PR :))

nuko8:
I've just pushed a new version which was revised along the line of:

> Since the extra precision given by nanoseconds over microseconds is completely statistically insignificant when running ~20000 runs over ~10 seconds, my suggestion was to just change subseconds to microseconds (since it is the lowest common denominator), and make e.g. the clock_gettime path do microseconds = val.tv_nsec / 1000. Once we always use microseconds, there is no need for the enum then. Does that make sense?

> that would be very welcome for a future PR :)

Oh, I'm very happy to hear that!

bluetech:
This looks great now. I only have one tiny comment, I can fix it up myself if you prefer: the "_t" is usually reserved for typedefs, so the type should be just `struct bench_time` instead of `struct bench_time_t`.

And while I'm here, a couple more suggestions which I forgot, but feel free to ignore them:

1. `bench_timer_get_elapsed_time_str` returns an allocated formatted string, but maybe we can have a `bench_timer_get_elapsed_time` which returns a `struct bench_time`, and let the caller format it however they want.

2. We usually prefix commit messages with the area of the code they change; this makes it easier to browse the log quickly. So that would be "bench: Modify benchmarks for a wider range of platforms" and "test/x11comp: Fix contention between X11 and Xvfb on Mac OS X".

Thanks for bearing with me :)

bluetech:
Oh, and one other thing: you should add a copyright header with your your name to bench.{c,h} files. For the preferred format see the file LICENSE.

nuko8:
> I can fix it up myself if you prefer: the "_t" is usually reserved for typedefs, so the type should be just struct bench_time instead of struct bench_time_t.

Done.  I should have done it when I moved the declaration to the header.

> bench_timer_get_elapsed_time_str returns an allocated formatted string, but maybe we can have a bench_timer_get_elapsed_time which returns a struct bench_time, and let the caller format it however they want.

Good idea.  Done.

> We usually prefix commit messages with the area of the code they change; this makes it easier to browse the log quickly. So that would be "bench: Modify benchmarks for a wider range of platforms" and "test/x11comp: Fix contention between X11 and Xvfb on Mac OS X".

Done.

> Oh, and one other thing: you should add a copyright header with your your name to bench.{c,h} files. For the preferred format see the file LICENSE.

Done.

One thing I must apologize to you.  I added your name there as coauthor without your permission in advance.  But I couldn’t help doing so because I do believe you deserve it.  I wish you wouldn’t mind it.

> Thanks for bearing with me :)

That’s just the word I have to say to you.  Thanks a lot!

bluetech:
Thanks for the contribution! Portability fixes which improve the code - good outcome.

We haven't had a release in a while, but we've accumulated enough that one is in order. I'll do one some time next month.

Issue #17, opened 2015-01-13
"Control" modifier incorrectly consumed on keypad


wm4:
If you press e.g. the '+' key on the keypad, and the ctrl key is pressed, then the xkb API will report the ctrl modifier as consumed, even though ctrl doesn't change the interpretation of the key (i.e. same keysym).

I'd expect that ctrl is not considered consumed in this case.

The APIs I was looking at were the following:

```
xkb_keysym_t sym = xkb_state_key_get_one_sym(state, code);
xkb_mod_index_t index = xkb_keymap_mod_get_index(keymap, XKB_MOD_NAME_CTRL);
int active = xkb_state_mod_index_is_active(state,  index, XKB_STATE_MODS_DEPRESSED);
int consumed = xkb_state_mod_index_is_consumed(state, code, index);
```

active and consumed can both be 1 for `code==XKB_KEY_KP_Add`.

bluetech:
The file `/usr/share/X11/xkb/symbols/keypad` has this in the `operator` section, which is included by default:

```
     key <KPAD> {
         type="CTRL+ALT",»       // +VMode switches to the next video mode
         symbols[Group1]= [ KP_Add, KP_Add, KP_Add, KP_Add, XF86_Next_VMode ]
     };
```

The type `CTRL+ALT` is defined as follows in `types/basic`:

```
    type "CTRL+ALT" {
        modifiers = Control+Alt+Shift+LevelThree;
        map[None] = Level1;
        map[Shift] = Level2;
        map[LevelThree] = Level3;
        map[Shift+LevelThree] = Level4;
        map[Control+Alt] = Level5;
        preserve[Shift] = Shift;
        preserve[Shift+LevelThree] = Shift;
        level_name[Level1] = "Base";
        level_name[Level2] = "Shift";
        level_name[Level3] = "Alt Base";
        level_name[Level4] = "Shift Alt";
        level_name[Level5] = "Ctrl+Alt";
    };
```

So the type indeed does not explicitly assign a level for just `Control`. However the way this is interpreted is that there an *implicit* assignment to Level0. As there is not `preserve` statement for this combination, the `Control` modifier is consumed in this case.

[...]

PR #29, opened 2016-01-20
keymap: add xkb_keymap_keyname_get_keycode()

zmike:
this function allows finding a keycode from a given keyname and
is useful for generating keyboard events to use in regression tests
during CI

Signed-off-by: Mike Blumenkrantz <zmike@osg.samsung.com>

### DIFF
--- a/src/keymap.c
+++ b/src/keymap.c
@@ -470,6 +470,40 @@ xkb_keymap_key_for_each(struct xkb_keymap *keymap, xkb_keymap_key_iter_t iter,
         iter(keymap, key->keycode, data);
 }
 
+XKB_EXPORT const char *
+xkb_keymap_key_get_name(struct xkb_keymap *keymap, xkb_keycode_t kc)
+{
+    const struct xkb_key *key = XkbKey(keymap, kc);
+
+    if (!key)
+        return NULL;
+
+    return xkb_atom_text(keymap->ctx, key->name);
+}
+
+XKB_EXPORT xkb_keycode_t
+xkb_keymap_key_by_name(struct xkb_keymap *keymap, const char *name)
+{
+    struct xkb_key *key;
+    xkb_atom_t atom;
+
+    atom = xkb_atom_lookup(keymap->ctx, name);
+    if (atom) {
+        xkb_atom_t ratom = XkbResolveKeyAlias(keymap, atom);
+        if (ratom)
+            atom = ratom;
+    }
+    if (!atom)
+        return XKB_KEYCODE_INVALID;
+
+    xkb_keys_foreach(key, keymap) {
+        if (key->name == atom)
+            return key->keycode;
+    }
+
+    return XKB_KEYCODE_INVALID;
+}
+
 /**
  * Simple boolean specifying whether or not the key should repeat.
  */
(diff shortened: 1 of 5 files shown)


zmike:
Would it be possible to do a release after merging this?

bluetech:
This is indeed a lacuna in the API, so makes sense to add. Some comments:

1. I do not particularly like hiding the loop in the function; it suggests we actually want to export the inverse, i.e. keycode -> name. However, due to the aliases, the function is not invertible, so we do need the name -> keycode function. But please do add the inverse, to be consistent with the similar functions we have for mods, layouts, leds. It should return the canonical name (please add a comment to this effect!). 
2. Please call the functions `xkb_keymap_key_get_name`, `xkb_keymap_key_by_name` - more consistent with existing names.
3. Please add a simple test just to exercise the code. I can do it after merging if you prefer.

Thanks.

src/keymap.c
```diff
@@ -470,6 +470,33 @@ xkb_keymap_key_for_each(struct xkb_keymap *keymap, xkb_keymap_key_iter_t iter,
...
+    name = xkb_atom_lookup(keymap->ctx, keyname);
+    if (!name)
+        return XKB_KEYCODE_INVALID;
+
+    xkb_keys_foreach(key, keymap) {
+        if (key->name == name)
+            return key->keycode;
+    }
+
+    name = XkbResolveKeyAlias(keymap, name);
```
bluetech: Would be better to canonicalize the name first, then do just one loop. An alias cannot conflict with a canonical name, so the result is the same.

xkbcommon/xkbcommon.h
```diff
@@ -941,6 +941,18 @@ xkb_keymap_key_for_each(struct xkb_keymap *keymap, xkb_keymap_key_iter_t iter,
                         void *data);
 
 /**
+ * Find the keycode for a given keyname in a keymap.
+ *
+ * If no corresponding key is found in the keymap, returns 0.
```
bluetech: Should say `XKB_KEYCODE_INVALID`.

Also, please write it like so, again for consistency:

```
@returns The keycode.  If no key with this name exists, returns
XKB_KEYCODE_INVALID.
```

zmike:
I tried adding a simple test (https://gist.github.com/zmike/88daf2a29a7b6d5d4987), but the first call fails and XKB_KEYCODE_INVALID is returned. Seems unexpected; did I do something wrong in the implementation?

bluetech:
The problem with the test is that there is no key with the name "9". You can check `/usr/share/X11/xkb/keycodes/evdev` to see the usual names (surrounded by `<..>`).

And two more comments :)

1. Please squash to one commit.
2. I don't think `test/state.c` is appropriate for this, as that file is meant for `xkb_state` tests. I don't see an existing appropriate file, so please add a new one `test/keymap.c` or so.

Thanks!

zmike:
Okay, I've added a separate test using a keyname from the evdev keycodes file, but it's still failing.

zmike:
Ah, and I'll squash after everything is resolved.

zmike:
Nevermind, user error. Squashing...

zmike:
Should be all done now.

bluetech:
I made some tiny tweaks and pushed, thanks.

A release is definitely overdue, I'll prepare one. @fooishbar what do you about making this one 1.0.0? I think it's about time :)

fooishbar:
@bluetech sadly I think we need to resolve #17 first. :( It's on my TODO ...

PR #31, opened 2016-02-27
Consumed modifiers modes

bluetech:
This is an implementation of what I suggested [here](https://bugzilla.gnome.org/show_bug.cgi?id=754110#c30).

There are no tests, so the implementation can have some bug. But I'll add them after we agree on the approach.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -83,6 +83,7 @@ AS_IF([test "x$ac_cv_func_secure_getenv" = xno -a \
 ])
 
 AX_GCC_BUILTIN(__builtin_expect)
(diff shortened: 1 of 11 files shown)


bluetech:
In the docstring I wrote this, which explains the popcount==1:

```
+     * - Additionally, a single modifier is considered consumed if, were it the
+     *   only active modifier affecting the key (as described above), it would
+     *   result in different keysyms being produced than when no modifiers are
+     *   active.
```

The sentence is a bit hard to parse, but I tried to be precise.

PR #32, opened 2016-03-15
doc: Fix Doxygen include for srcdir != builddir

fooishbar:
Instead of giving Doxygen a series of absolute paths to the source
files and a relative path to the output directory, run it from the
source directory with purely relative paths to the source files, and
give it an absolute path to the build directory.

This fixes the parsing of README.md with a separate build directory,
since the relative includes for doc/quick-guide.md and doc/compat.md
don't resolve otherwise. Doxygen's INCLUDE_PATH turns out not to fix
this either, since that's just a set of paths to open and parse, rather
than an analogue to cpp's -I.

Signed-off-by: Daniel Stone <daniels@collabora.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: c
 
 before_install:
     - sudo apt-get update -qq
-    - sudo apt-get install xutils-dev
+    - sudo apt-get install xutils-dev doxygen
 
 compiler:
     - gcc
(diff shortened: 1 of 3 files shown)


fooishbar:
Hm, I didn't realise that updating a source branch would also update a PR. Oops. Anyway, the first commit (paths vs. Doxyfile) seems fine, but the second commit seems to require unbounded time to install doxygen/texlive on Travis infrastructure. :(

bluetech:
Right, this eliminates that distcheck warning I continuously ignored :)

PR #33, opened 2016-03-15
travis: remove --disable-x11

bluetech:
Looks like the CI machines were updated to use Ubuntu 14.04, which has
libxcb 1.10, which is new enough. Lets see if it works.

Signed-off-by: Ran Benita <ran234@gmail.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,13 @@
+sudo: required
+dist: trusty
 language: c
 
 before_install:
     - sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


bluetech:
Doesn't seem to work!

fooishbar:
@bluetech Through no fault of your own - looks like the mirrors were broken at the time.

fooishbar:
@bluetech And look at that, it works now! World's unluckiest transient failure.

bluetech:
Huh, nice. So I'll merge this.

PR #35, opened 2016-04-14
build: Make sure to create the root dir at install phase

jwendell:
It might not exist, so context creation will fail with the message
"failed to add default include path".

Signed-off-by: Jonh Wendell <jonh.wendell@redhat.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -140,6 +140,9 @@ CLEANFILES = $(BUILT_SOURCES)
 src/xkbcomp/parser.c: $(top_builddir)/src/$(am__dirstamp) $(top_builddir)/src/xkbcomp/$(am__dirstamp)
 src/xkbcomp/parser.h: $(top_builddir)/src/$(am__dirstamp) $(top_builddir)/src/xkbcomp/$(am__dirstamp)
 
+install-data-local::
+	$(MKDIR_P) $(XKBCONFIGROOT)
+
 ##
 # Documentation
 ##


fooishbar:
If it's empty, then we won't be able to compile anything anyway, since there will be no rules in there. This will also break ``make install DESTDIR=foo``, if the user isn't installing as root, or doesn't expect the installation to have side effects.

Do you have a particular case this fixes?

jwendell:
I'm using jhbuild to compile and run gnome directly from git.

Then, taken from config.log:
XKBCONFIGROOT='${datarootdir}/X11/xkb'

Where datarootdir is something like /home/jwendell/jhbuild/install/share

At this point, this directory, i.e /home/jwendell/jhbuild/install/share/X11/xkb doesn't exist.

Then, gnome-shell (or other piece of software, I'm not sure) fails to startup due to xkb_context_new() refusing to complete.

fooishbar:
@jwendell Creating an empty directory won't make keymap (rather than context) creation succeed though: it needs that directory to in fact be populated with valid XKB data. If keymap creation suddenly succeeds with that directory created, something else has gone quite horribly wrong.

If you can verify that the presence of an empty directory is enough to change things, can you please provide the output of running with XKB_LOG_LEVEL=debug both with and without that directory created?

Also, as an aside, jhbuild should either build and deploy its own copy of xkeyboard-config, or more usefully, just take it as a sysdep and build xkbcommon using --with-xkb-config-root=/usr/share/X11/xkb to use the system keyboard data.

jwendell:
You are right. I think changing jhbuild to use system's config s the way to go. I'll cook a patch there.

Sorry for the noise, I never had a contact with this library before, and did a quick fix without understanding the issue as a whole.

Thanks!

fooishbar:
@jwendell No problem! I'd still be interested to see why that fixed it (as keymap creation should fail even if context creation succeeds, and that should be fatal) though, even with jhbuild being fixed.

PR #45, opened 2017-04-11
Wayland: xdg-shell v6 support

fooishbar:
One small bugfix, and one feature for newer versions of Wayland protocols.


bluetech:
Right, master indeed no longer works with gnome-shell, but works great after this.

PR #46, opened 2017-05-12
Sync Keysyms with recent xproto additions

jwrdegoede:
xproto recently has been extended with 4 new keysyms:
XF86XK_Keyboard
XF86XK_WWAN
XF86XK_RFKill
XF86XK_AudioPreset

This commit is the result of running "make update-keysyms" on a system
with the updated xproto installed.

Signed-off-by: Hans de Goede <hdegoede@redhat.com>


bluetech:
Thanks!

If you need this released quickly, please prod me here when xproto is released, and I'll do an xkbcommon release as well.

PR #48, opened 2017-05-27
Fixed a minor bug in error detection in Wayland test

namandixit:
Fixed test for checking failure to create xkb_state in test/interactive-wayland.c


bluetech:
Thanks!

PR #49, opened 2017-07-31
build: add meson build system

bluetech:
I wanted to see what the hype was about. The hype is true - meson is great.

@fooishbar Would be great if you could spare some time to look this over. As preparation I removed a couple of things I think no one uses: the Android build support and the `-uninstalled` pkg-config files. If you think we should keep them I'll try to do that.


meson.build
```diff
@@ -0,0 +1,484 @@
...
+test_dep = declare_dependency(
+    include_directories: include_directories('src'),
+    link_with: libxkbcommon_test_internal,
+)
+test(
+    'keysym',
+    executable('test-keysym', 'test/keysym.c', dependencies: test_dep),
+    env: test_env,
+)
+test(
```
fooishbar: I'd use a for loop for these, like https://git.collabora.com/cgit/user/daniels/wayland.git/tree/meson.build?h=wip/2016-11/meson&id=15a93a7701ecb4991f5cda35aa0acb85ccdceb60#n325

meson.build
```diff
@@ -0,0 +1,484 @@
...
+    executable('interactive-evdev', 'test/interactive-evdev.c', dependencies: test_dep)
+endif
+if get_option('enable-x11')
+    executable('interactive-x11', 'test/interactive-x11.c', dependencies: test_dep, link_with: libxkbcommon_x11_internal)
+endif
+if get_option('enable-wayland')
+    wayland_client_dep = dependency('wayland-client', version: '>=1.2.0')
+    wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.7')
+    wayland_scanner_dep = dependency('wayland-scanner')
+    if not wayland_client_dep.found() or not wayland_protocols_dep.found() or not wayland_scanner_dep.found()
```
fooishbar: Same comment as on the X11 deps, except these ones don't actually have required: false, so you'll never hit this branch.

meson.build
```diff
@@ -0,0 +1,484 @@
...
+    'rules',
+    executable('bench-rules', 'bench/rules.c', dependencies: bench_dep),
+    env: bench_env,
+)
+benchmark(
+    'rulescomp',
+    executable('bench-rulescomp', 'bench/rulescomp.c', dependencies: bench_dep),
+    env: bench_env,
+)
+benchmark(
```
fooishbar: More for loops.

meson.build
```diff
@@ -0,0 +1,484 @@
...
+    else
+        XKBCONFIGROOT = join_paths(get_option('prefix'), get_option('datadir'), 'X11', 'xkb')
+  endif
+endif
+
+
+# The X locale directory for compose.
+XLOCALEDIR = get_option('x-locale-root')
+if XLOCALEDIR == ''
+    XLOCALEDIR = join_paths(get_option('prefix'), get_option('datadir'), 'X11', 'locale')
```
bluetech: I think I tried it originally (when adding compose). But I don't remember what was the problem.

Ideally libX11 would have a special variable for this path in its pc file, but it doesn't... So I'll try this.

meson.build
```diff
@@ -0,0 +1,484 @@
...
+    'src/keymap-priv.c',
+    'src/scanner-utils.h',
+    'src/state.c',
+    'src/text.c',
+    'src/text.h',
+    'src/utf8.c',
+    'src/utf8.h',
+    'src/utils.c',
+    'src/utils.h',
+    include_directories: include_directories('src'),
```
bluetech: Yes, from my tests it works as it's supposed to. I assume the source file's own directory comes before the `include_directories` in the search path.

meson.build
```diff
@@ -0,0 +1,484 @@
...
+    version: meson.project_version(),
+    description: 'XKB API common to servers and clients',
+)
+
+
+# libxkbcommon-x11.
+if get_option('enable-x11')
+    xcb_dep = dependency('xcb', version: '>=1.10', required: false)
+    xcb_xkb_dep = dependency('xcb-xkb', version: '>=1.10', required: false)
+    if not xcb_dep.found() or not xcb_xkb_dep.found()
```
bluetech: Yes, it's for the error message.

The component idea definitely looks like it would fit here.

fooishbar:
Consider all my concerns addressed. Ship it? :)

meson.build
bluetech: I started out with a for loop over an array of arrays but I didn't like it much. You lose the flexibility to modify a specific test, and it makes things just a little more opaque.

 I think the repetition is not so bad, I'm trained by automake to accept that :)

meson.build
bluetech: Right, they are missing the `required: false`. I'll add them.

bluetech:
@fooishbar thanks for the review!

The CI build fails, for some reason I naively assumed it would just have `meson` installed. I'll try to install it in the CI, it would probably take several attempts though.

fooishbar:
It'd probably be ancient Meson anyway. If you add `git clone https://github.com/mesonbuild/meson && pushd meson && python3 setup.py build && popd`, then exec `../meson/meson.py` instead of a global `meson` from `$PATH`, it should just work. You'll still need to install ninja tho.

fooishbar:
> - Meson uses `-g` by default, autotools uses `-g -O2`.
As far as this one goes, see [Meson's guide](http://mesonbuild.com/Running-Meson.html): there are various standardised build types you can use to control debug & optimisation levels.

PR #53, opened 2017-09-29
keysym-utf: Add missing codes for signifblank and permille keysyms

whitslack:
While working on a custom keyboard layout, I noticed that the `permille` keysym was not being translated to the U+2030 (PER MILLE SIGN) character. Evidently this keysym was omitted from `keysymtab`, presumably by mistake. It is even absent from the original [keysym2ucs](http://www.cl.cam.ac.uk/~mgk25/ucs/keysym2ucs.c) table from which xkbcommon's `keysym-utf.c` is derived.

This pull request adds the missing `permille` table entry and also fills in the placeholder for `signifblank` U+2423 (OPEN BOX).

### DIFF
--- a/src/keysym-utf.c
+++ b/src/keysym-utf.c
@@ -521,7 +521,7 @@ static const struct codepair keysymtab[] = {
     { 0x0aa8, 0x200a }, /*                   hairspace   HAIR SPACE */
     { 0x0aa9, 0x2014 }, /*                      emdash — EM DASH */
     { 0x0aaa, 0x2013 }, /*                      endash – EN DASH */
-    /*  0x0aac                               signifblank ? ??? */
+    { 0x0aac, 0x2423 }, /*                 signifblank ␣ OPEN BOX */
     { 0x0aae, 0x2026 }, /*                    ellipsis … HORIZONTAL ELLIPSIS */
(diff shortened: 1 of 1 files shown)


bluetech:
The PER MILLE seems obviously correct and also matches [libx11](https://cgit.freedesktop.org/xorg/lib/libX11/tree/src/xlibi18n/imKStoUCS.c#n146).

The open box one seems a little bit more unclear (looks like the keysym2ucs author wasn't sure about it, and also libx11 doesn't match it). However, looking [here](https://en.wikipedia.org/wiki/Whitespace_character#Unicode) it does seems like the only clear choice for "significant space" so let's go with it.

Thanks!

whitslack:
> The open box one seems a little bit more unclear (looks like the keysym2ucs author wasn't sure about it, and also libx11 doesn't match it).

I didn't arbitrarily choose a code point for this keysym. I got it right out of [`<X11/keysymdef.h>`](https://cgit.freedesktop.org/xorg/proto/x11proto/plain/keysymdef.h). Perhaps whatever automated script originally generated the keysym2ucs table got confused by the parentheses in the comment on the definition of `XK_signifblank`. Regardless, it is unambiguous that U+2423 is the intended code point.

bluetech:
I said that because they put parentheses around it, `(U+2423 OPEN BOX)`, which I took to convey uncertainty. Though I see now the parentheses are in many other lines in this file.

PR #63, opened 2018-07-16
meson: Fix xkbcommon-x11.pc Requires

heftig:
The meson-generated pkgconfig file was missing Requires and
Requires.private.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -216,6 +216,11 @@ if get_option('enable-x11')
 You can disable X11 support with -Denable-x11=false.''')
     endif
 
+    libxkbcommon_x11_deps = [
+        xcb_dep,
+        xcb_xkb_dep,
+    ]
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks, these are definitely missing.

The CI failure is legitimate though:

```
Meson encountered an error in file meson.build, line 257, column 14:
Item must be a list or one of <class 'str'>
```

heftig:
Hmm, apparently support for passing libs and deps like that was added in meson 0.46.0.

heftig:
A different fix was applied (a3c907d3891bb275b9d2592d7a94d94020866b3c) but resulted in issue #64, fixed by PR #65.

bluetech:
Sorry about that! I made the commit but got distracted and forgot to report back.

PR #65, opened 2018-07-17
meson: Fix xkbcommon-x11.pc Requires versioning

### LINKED ISSUE #64: Build failure in GNOME Continuous
GNOME components that depend on xkbcommon-x11 started to fail to build after commit a3c907d3891bb275b9d2592d7a94d94020866b3c with this error:

```
Package 'xcb-xkb>=1.10', required by 'xkbcommon-x11', not found
```

See full build log for Mutter, available [here](http://build.gnome.org/continuous/buildmaster/builds/2018/07/17/28/build/log-mutter.txt).

heftig:
Old meson expects an array with one dependency per element. Providing a
string containing multiple deps results in only the first dep getting
its whitespace properly applied. As a result, the output was:

    Requires.private: xcb >= 1.10 xcb-xkb>=1.10

And downstream projects failed to find a package named 'xcb-xkb>=1.10'.

Specifying an array of versioned deps results in correct output:

    Requires.private: xcb >= 1.10, xcb-xkb >= 1.10

Fixes #64.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -258,8 +258,8 @@ You can disable X11 support with -Denable-x11=false.''')
         libraries: libxkbcommon_x11,
         version: meson.project_version(),
         description: 'XKB API common to servers and clients - X11 support',
-        requires: 'xkbcommon',
-        requires_private: 'xcb>=1.10 xcb-xkb>=1.10',
+        requires: ['xkbcommon'],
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks!

PR #67, opened 2018-08-03
Fix more fuzz-testing fallout

fooishbar:
Most of this was found quite a while ago but got lost.


bluetech:
Thanks! They all look good to me. I'll make another release with these soon.

Issue #72, opened 2018-10-27
How to get and set the currently active layout?


Hi-Angel:
I'm trying [to allow sway to store a layout per window](https://github.com/swaywm/sway/issues/2361#issuecomment-433642415). I figured that currently active layout is stored in `xkb_state`.

As I understand, libxkbcommon doesn't have any "layout-changed" hook, so I need to manually check active layout every time `xkb_state` being touched *(is that correct?)*.

Then questions come down to:

1. How to get an id of currently active layout?
2. How to set the currently active layout with an id.

The closest I could find was [this issue](https://github.com/xkbcommon/libxkbcommon/issues/20), but they're getting layout name with `xkb_keymap_layout_get_name()`, whereas I guess I gonna need an index; and I couldn't find in `xkbcommon.h` anything with infix `layout_set` anyway.

bluetech:
Before I answer your questions, I'd like to understand how are you planning to architect things, since that would affect the answer. Will each window have it's own `xkb_state`? What would you like to happen when a new window is created?

Hi-Angel:
> Will each window have it's own xkb_state?

Per my understanding `xkb_state` contains not only layout, but also lots of other stuff. I want only to have layout per window, so I'm guessing instead I need to keep with each window a `xkb_layout_index_t` which is an ID of a layout in libxkbcommon.

> What would you like to happen when a new window is created?

I imagine, this window should have "default layout", which I'm going to get just after compositor was started, and then assign it to every new window.

Hi-Angel:
Currently `xkb_state` is stored per keyboard device *(the `struct wlr_keyboard` from wlroots)*. I think of keeping it that way.

Hi-Angel:
Anyway, thank you for the tips, I guess the question can be closed.

Hi-Angel:
I'm a bit confused though: the `xkb_state_serialize_layout()` accepts a non-const `xkb_state` . Does it modify `xkb_state` somehow? Don't see anything in description.

bluetech:
No, it doesn't modify the `xkb_state`. None of the method receivers are marked `const`. I guess it didn't sufficiently bother anyone. Also, in a few cases, interior mutability is used (ref counts, caches, scratch buffers etc.), that's another reason.

Hi-Angel:
> Let us know how it goes.

[...]

PR #73, opened 2018-11-05
RFC: Constify functions that doesn't mutate state

Hi-Angel:
This PR is pretty small, I wanted to get some feedback before proceeding.

As part of [somewhat irrelevant issue I found](https://github.com/xkbcommon/libxkbcommon/issues/72#issuecomment-435698478) that libxkbcommon has a lot of functions that do not mutate any state, yet they accept non-const arguments.

Having such arguments declared as `const`:

1. Improves readability for library users
2. Allows compiler optimizations *(which sounds important, given xkb protocol used a lot by applications)*
3. Enables better design for end-users of the library, as they, in turn, can declare arguments they don't intend to modify as `const`.

So I wanted to constify whatever possible without breaking existing interface.

My questions:

1. Is it okay with library maintainers?
2. How shall I design patches to suit better for review: e.g. one big commit, or a commit per change, or a commit per `.h` file?

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -129,7 +129,7 @@ xkb_context_include_path_reset_defaults(struct xkb_context *ctx)
  * Returns the number of entries in the context's include path.
  */
 XKB_EXPORT unsigned int
-xkb_context_num_include_paths(struct xkb_context *ctx)
+xkb_context_num_include_paths(const struct xkb_context *ctx)
 {
     return darray_size(ctx->includes);
 }
@@ -293,7 +293,7 @@ xkb_context_set_log_fn(struct xkb_context *ctx,
 }
 
 XKB_EXPORT enum xkb_log_level
-xkb_context_get_log_level(struct xkb_context *ctx)
+xkb_context_get_log_level(const struct xkb_context *ctx)
 {
     return ctx->log_level;
 }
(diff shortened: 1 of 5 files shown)


whot:
IMO making the context argument `const` can bite back in the future, exactly because it prevents any changes to the context. While there are many cases where this seems like a nonissue (e.g. `xkb_context_get_log_level`) it prevents future improvements like lazy fetching of data and sometimes even logging debug messages (because you may have to fetch extra state to do so).

The context pointer itself is already "magic", you don't know what it does short of the public API. I don't think adding the consts here makes much of a difference for readability and it's hard to judge the compiler optimization impact.

Hi-Angel:
You might be right… The compiler optimization that compiler can do on the user side given const args is not to reload a pointer to a field after returning from a function call. However libxkbcommon gives a user opaque structs, so such optimization wouldn't make sense, i.e. because user can't access fields of a struct.

I also tried building the lib with and without my PR in release-mode, and compared the content of .text section *(output of `objdump -d libxkbcommon.so`)*, and I see no difference.

There's still a point I don't like that users of the library can't declare pointers as `const`, but libxkbcommon haven't got many functions, even less so ones that can be constified. So it probably won't affect design of end-aplications much, I guess I can live with it.

PR #85, opened 2019-02-08
meson.build: manually link all sources into the library

### LINKED ISSUE #84: Can not build static library
Hi! 

I'm trying to build libxkbcommon using Meson + Ninja, but when I set to build as static library, the symbols are not exported, however for shared library it works fine.

#### Steps

    git clone https://github.com/xkbcommon/libxkbcommon.git
    cd libxkbcommon
    git checkout xkbcommon-0.8.2
    mkdir build
    meson ${PWD} build --backend=ninja --default-library=static
    ninja -C build/
    nm build/libxkbcommon.a

You could execute the same steps using `--default-library=shared` instead.

#### Environment
OS: Linux Ubuntu 18.10 amd64
Meson: 0.49.0
Ninja: 1.8.2
Compiler: GCC 7.3.0

/cc @ericLemanissier

Regards!

whot:
This works around meson bug 3937, 'link_whole' arguments don't get added into
the final static library and we end up with a virtually empty 8-byte
libxkbcommon.a file, see https://github.com/mesonbuild/meson/issues/3937

Workaround is simply to add all sources to both libraries we need them in.
This obviously compiles them twice but this year's winter was cold and
bit of extra warmth will be appreciated.

Fixes #84

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -122,8 +122,7 @@ yacc_gen = generator(
     output: ['@BASENAME@.c', '@BASENAME@.h'],
     arguments: ['@INPUT@', '--defines=@OUTPUT1@', '--output=@OUTPUT0@', '-p _xkbcommon_'],
 )
-libxkbcommon_internal = static_library(
-    'xkbcommon-internal',
+libxkbcommon_sources = [
     'src/compose/parser.c',
     'src/compose/parser.h',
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks!

PR #90, opened 2019-02-23
meson.build: use program from build machine not host or target.

maxice8:
We can't always execute binaries from the host or target machine,
as is the case in cross compilation.

This instead uses native : true to search the build system for wayland-scanner.pc and get it from there.

Fixes cross compilation on Void Linux

Side-note: Meson is so nice and makes this so trivially easy.


maxice8:
?

PR #95, opened 2019-03-23
Fix some of LGTM warnings

Hi-Angel:
source: https://lgtm.com/projects/g/xkbcommon/libxkbcommon/alerts/?mode=list

### DIFF
--- a/src/xkbcomp/symbols.c
+++ b/src/xkbcomp/symbols.c
@@ -227,6 +227,7 @@ MergeGroups(SymbolsInfo *info, GroupInfo *into, GroupInfo *from, bool clobber,
     /* First find the type of the merged group. */
     if (into->type != from->type) {
         if (from->type == XKB_ATOM_NONE) {
+            /* it's empty for consistency with other comparisons */
         }
         else if (into->type == XKB_ATOM_NONE) {
             into->type = from->type;
@@ -269,6 +270,7 @@ MergeGroups(SymbolsInfo *info, GroupInfo *into, GroupInfo *from, bool clobber,
         struct xkb_level *fromLevel = &darray_item(from->levels, i);
 
         if (fromLevel->action.type == ACTION_TYPE_NONE) {
+            /* it's empty for consistency with other comparisons */
         }
         else if (intoLevel->action.type == ACTION_TYPE_NONE) {
             intoLevel->action = fromLevel->action;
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks! That's an interesting tool.

Regarding the `UCSConvertCase` commit, this function is copy/pasted from libX11 (see comment at the top of the function), and I wouldn't want to diverge from it, unless it fixes a bug.

Regarding the `symbols` changes, from what I remember the code is intentionally written this way: {first easy case} else {second easy case} else {complicated case}. Here the first easy case has nothing to do. I think the tool wants us to add a `/* Nothing to do. */` comment in there, to acknowledge that the empty body is intentional. I'd be happy to take a patch for that.

Hi-Angel:
> Regarding the `symbols` changes, from what I remember the code is intentionally written this way: {first easy case} else {second easy case} else {complicated case}. Here the first easy case has nothing to do. I think the tool wants us to add a `/* Nothing to do. */` comment in there, to acknowledge that the empty body is intentional. I'd be happy to take a patch for that.

I don't think anybody would see this pattern besides the original writer. I makes the code more complex, because you have to remember that "Aha, in this another `if-else` branch the `from->type` can't be equal to `XKB_ATOM_NONE`. Whereas in the new version this made explicit by indentation. Besides, the newer version has less code to read.

Anyway, these are subtle things, but having an empty `if ()` block doesn't sound right.

Hi-Angel:
> Regarding the `UCSConvertCase` commit, this function is copy/pasted from libX11 (see comment at the top of the function), and I wouldn't want to diverge from it, unless it fixes a bug.

So, if I make a commit to X11, will it appear here as well? How often the code here is updated?

bluetech:
> I don't think anybody would see this pattern besides the original writer.

There are 3 distinct cases. In this existing code it looks like

```
if (...) {
   // Case 1
} else if (...) {
   // Case 2
} else {
   // Case 3
}
```

whereas with the diff it's

```
if (...) {
    if (...) {
        // Case 2
    } else {
        // Case 3
    }
}
// Case 1 implicitly empty
```

I like the symmetry, reduced indentation and explicitness of the former, this is probably why I wrote it this way.

> So, if I make a commit to X11, will it appear here as well? How often the code here is updated?

If you change libX11 I'd be happy to sync up. The code is not updated regularly, only when someone prompts it. libX11 doesn't change much so I don't remember there was ever a need to.

Hi-Angel:
I don't like it, but you're a maintainer, so I did the changes. I put a comment `/* it's empty for consistency with other comparisons */`, see if it's okay with you.

bluetech:
Thanks!

PR #96, opened 2019-06-04
Fix a few doc typos

Jjagg:


bluetech:
Thanks @Jjagg and @whot.

PR #98, opened 2019-06-07
Use bitwise test for power-of-2 instead of popcount

michaelforney:
We don't need to determine the total number of bits set to determine if
at most one is set.

Additionally, on x86_64 without any -march=* flag, __builtin_popcount
will get compiled to a function call to the compiler runtime (on gcc),
or a long sequence of bit operations (on clang).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -77,7 +77,6 @@ AS_IF([test "x$ac_cv_func_secure_getenv" = xno -a \
 ])
 
 AX_GCC_BUILTIN(__builtin_expect)
-AX_GCC_BUILTIN(__builtin_popcount)
 
 # Some tests use Linux-specific headers
 AC_CHECK_HEADER([linux/input.h])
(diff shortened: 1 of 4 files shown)


bluetech:
Removing the intrinsic makes sense, very much premature optimization.

The test you added is not equivalent for the `entry->mods.mask == 0`. I prefer to keep the code reading `my_popcount == 1` because that's how the logic is specified. Just removing the fast-path from `my_popcount` would be best I think.

michaelforney:
It's not equivalent to popcount for the `== 0` case, but in that case the bitwise and with anything is 0, and the result is the same.

Instead, maybe we could rename the helper to `multiple_bits_set`, and change the condition to

```C
if (entry == matching_entry || !multiple_bits_set(entry->mods.mask))
```

which I think would capture the intent just as well.

bluetech:
The desired check is "exactly one bit set" (false for 0), not "is power of two"/"!(multiple bits set)" (true for 0). I think `popcount() == 1` expressed this well.

michaelforney:
I updated it to use a helper that checks if exactly one bit is set.

bluetech:
Thanks. The commit message needs an update. I can do this if you prefer.

michaelforney:
Done. Feel free to tweak as you want.

bluetech:
Thanks!

PR #100, opened 2019-07-02
interactive-wayland: Port to stable xdg-shell

fooishbar:
xdg_shell v6 was pretty close to the finalised stable version of
xdg-shell. We can now just use the stable version, which is supported
everywhere (Enlightenment, KWin, Mutter, Weston, wlroots).

This requires bumping the wayland-protocols dependency.

Signed-off-by: Daniel Stone <daniels@collabora.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -299,8 +299,8 @@ endef
 	$(AM_V_GEN)$(MKDIR_P) $(dir $@) && $(wayland_scanner) client-header < $< > $@
 
 INTERACTIVE_WL_XDG_SHELL_SRCS = \
-	xdg-shell-unstable-v6-protocol.c \
-	xdg-shell-unstable-v6-client-protocol.h
+	xdg-shell-protocol.c \
+	xdg-shell-client-protocol.h
 
 test_interactive_wayland_SOURCES = \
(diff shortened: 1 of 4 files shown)


bluetech:
I tried it and it works as expected.

When building (also before this PR), this message is issued by `wayland-scanner`:

```
Using "code" is deprecated - use private-code or public-code.
See the help page for details.
```

Would be nice to fix that too if possible.

fooishbar:
> When building (also before this PR), this message is issued by `wayland-scanner`:
> 
> ```
> Using "code" is deprecated - use private-code or public-code.
> See the help page for details.
> ```
> 
> Would be nice to fix that too if possible.

`private-code` and `public-code` are new since wayland-scanner 1.15.0, which isn't available in the current Debian stable version (though coming in Buster), nor is it available in the current Ubuntu LTS (but available in the non-LTS release).

We could add a version check which uses one or the other depending on the available version, but the thought of doing that in autotools made me lose the will to live.

bluetech:
Got it. Let's defer that for now...

Merging, thanks!

fooishbar:
Thanks Ran! :)

PR #103, opened 2019-08-05
[RFC] Make it possible to build using Meson and MSVC

aperezdc:
This PR includes a series of commits which make it possible to build libxkbcommon on Windows using Microsoft Visual C (a.k.a. MSVC). Most of the commits are quite small and deal with providing small implementations of functions commonly available on “Unixy” systems (GNU/Linux, the BSDs, etc.) which are not available in Windows when using MSVC: in some cases functions are renamed (e.g. `strdup` → `_strdup`), small wrappers are needed over native functions (`gettimeofday`), or different headers need to be used.

The biggest change by itself is the commit that attempts to use backslashes as directory separators where appropriate. I had the hope that this would make `test-filecomp` pass successfully, but sadly I am probably missing something—my lack of Windows development knowledge and not having a Windows box around may be to blame.

A few interspersed commits are used to enable CI builds on AppVeyor, which are currently running for my `msvc` branch ([sample build](https://ci.appveyor.com/project/aperezdc/libxkbcommon/builds/26483305)), so that's how I know things are building and most tests passing.

I have no idea why `test-filecomp` is broken, and some help to understand and figure out what it is supposed to do and how the output should look would be very appreciated—you can check the output of the test in the build linked above.

### DIFF
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -24,9 +24,40 @@
 
 #include <assert.h>
 #include <stdio.h>
-#include <sys/time.h>
 
 #include "bench.h"
+#include "../src/utils.h"
+
+#ifndef _MSC_VER
+#include <sys/time.h>
+#else
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <stdint.h>
+
+struct timeval {
+    long tv_sec, tv_usec;
+};
+
+static int
+gettimeofday(struct timeval *tv, void *unused)
+{
+    static const uint64_t EPOCH = ((uint64_t) 116444736000000000ULL);
+
+    SYSTEMTIME system_time;
+    FILETIME file_time;
+    uint64_t t;
+
+    GetSystemTime(&system_time);
+    SystemTimeToFileTime(&system_time, &file_time);
+    t = (uint64_t) file_time.dwLowDateTime;
+    t += ((uint64_t) file_time.dwHighDateTime) << 32;
+
+    tv->tv_sec  = (long) ((t - EPOCH) / 10000000L);
+    tv->tv_usec = (long) (system_time.wMilliseconds * 1000);
+    return 0;
+}
+#endif
 
 void
 bench_start(struct bench *bench)
--- a/src/utils.c
+++ b/src/utils.c
@@ -161,3 +161,48 @@ istrncmp(const char *a, const char *b, size_t n)
     }
     return 0;
 }
+
+#if !(defined(HAVE_ASPRINTF) && HAVE_ASPRINTF)
+int
+asprintf(char **strp, const char *fmt, ...)
+{
+    int ret;
+    va_list ap;
+    va_start(ap, fmt);
+    ret = vasprintf(strp, fmt, ap);
+    va_end(ap);
+    return ret;
+}
+
+# if !(defined(HAVE_VASPRINTF) && HAVE_VASPRINTF)
+int
+vasprintf(char **strp, const char *fmt, va_list ap)
+{
+    int ret;
+    char *buf;
+    va_list ap_copy;
+
+    /*
+     * The value of the va_list parameter is undefined after the call to
+     * vsnprintf() returns: pass a copy to make sure "ap" remains valid.
+     */
+    va_copy(ap_copy, ap);
+    ret = vsnprintf(NULL, 0, fmt, ap_copy);
+    va_end(ap_copy);
+
+    if (ret < 0)
+        return ret;
+
+    if (!(buf = malloc(ret + 1)))
+        return -1;
+
+    if ((ret = vsnprintf(buf, ret + 1, fmt, ap)) < 0) {
+        free(buf);
+        return ret;
+    }
+
+    *strp = buf;
+    return ret;
+}
+# endif /* !HAVE_VASPRINTF */
+#endif /* !HAVE_ASPRINTF */
--- a/src/utils.h
+++ b/src/utils.h
@@ -278,4 +278,12 @@ unmap_file(char *string, size_t size);
 #define ATTR_PACKED
 #endif
 
+#if !(defined(HAVE_ASPRINTF) && HAVE_ASPRINTF)
+int asprintf(char **strp, const char *fmt, ...) ATTR_PRINTF(2, 3);
+# if !(defined(HAVE_VASPRINTF) && HAVE_VASPRINTF)
+#  include <stdarg.h>
+int vasprintf(char **strp, const char *fmt, va_list ap);
+# endif /* !HAVE_VASPRINTF */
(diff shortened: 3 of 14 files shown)


bluetech:
This sounds great -- I'd like the library to be as portable as possible. I will hopefully review it on Friday, if no one beats me to it.

A quick note on CI: first, thanks for adding one. Second, I've heard that Travis CI, which we currently use for Linux and Darwin, was sold off and might be going away or some such. The main alternative seems to be [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) (from Microsoft...) which according to the link supports all of Linux, Darwin and Windows. So if we switch to that, we could keep using just one CI service.

Do you think you'll be able to use Azure instead of Appveyor? 

- If not, that's perfectly fine -- we'll keep the Azure as a separate issue.
- If yes, you can do Windows only. But if you also manage to get the Linux and Darwin builds going, that'd be perfect :)

aperezdc:
> This sounds great -- I'd like the library to be as portable as possible. I will hopefully review it on Friday, if no one beats me to it.

Thanks :blush: 
 
> A quick note on CI: first, thanks for adding one. Second, I've heard that Travis CI, which we currently use for Linux and Darwin, was sold off and might be going away or some such. The main alternative seems to be [Azure Pipelines](https://azure.microsoft.com/en-us/services/devops/pipelines/) (from Microsoft...) which according to the link supports all of Linux, Darwin and Windows. So if we switch to that, we could keep using just one CI service.
> 
> Do you think you'll be able to use Azure instead of Appveyor?
>
> * If not, that's perfectly fine -- we'll keep the Azure as a separate issue.
> * If yes, you can do Windows only. But if you also manage to get the Linux and Darwin builds going, that'd be perfect :)

After trying out Azure for a while today it looks to me that it should be possible to use that for Windows builds as well. I will look into rebasing this patch set later during the week on top of PR #104, then add the Windows build configuration to the Azure configuration.

src/utils.h
```diff
@@ -122,6 +122,19 @@ memdup(const void *mem, size_t nmemb, size_t size)
     return p;
 }
 
+#ifdef _MSC_VER
+static inline char *
+strndup(const char *str, size_t len)
+{
+    char *p = calloc(len + 1, sizeof(char));
```
bluetech: Would be better to do a `strlen(s)` and take the min instead of always allocating `len+1` bytes. E.g. `strndup("foo", 10000)` should only allocate 4 bytes.

src/utils.h
```diff
@@ -122,6 +122,19 @@ memdup(const void *mem, size_t nmemb, size_t size)
     return p;
 }
 
+#ifdef _MSC_VER
```
bluetech: Let's use a configure check here instead of checking for MSVC, this way it'd work for any platform which might be missing `strndup`.

src/utils.c
```diff
@@ -161,3 +161,41 @@ istrncmp(const char *a, const char *b, size_t n)
...
+    int r;
+    va_list ap;
+    va_start(ap, fmt);
+    r = vasprintf(strp, fmt, ap);
+    va_end(ap);
+    return r;
+}
+
+int
+vasprintf(char **strp, const char *fmt, va_list ap)
```
bluetech: It seems like MSVC does have [`vsnprintf`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-vsnprintf-vsnprintf-l-vsnwprintf-vsnwprintf-l?view=vs-2019), so `vasprintf` could be written a bit more generally like this (written quickly):

```c
int vasprintf(char **strp, const char *fmt, va_list ap) {
    int ret;
    char *buf;

    ret = vsnprintf(NULL, 0, fmt, ap);
    if (ret < 0) {
        return ret;
    }

    buf = malloc(len + 1);
    if (!buf) {
        return -1;
    }

    ret = vsnprintf(buf, ret + 1, fmt, ap);
    if (ret < 0) {
        free(buf);
        return ret;
    }

    *strp = buf;
    return ret;
}
```

What do you think about using a configure check for `vasprintf` and using the above as the fallback implementation, instead of using an MSVC-specific implementation?

bench/bench.c
```diff
@@ -24,9 +24,40 @@
...
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <stdint.h>
+
+struct timeval {
+    long tv_sec, tv_usec;
+};
+
+static int
+gettimeofday(struct timeval *tv, void *unused)
```
bluetech: Hmm, maybe we should just use `clock()`, which is not great but is standard C, instead?

This is fine, anyway.

aperezdc:
I'll rebase this onto `master` and apply the suggested changes in the next days. Thanks a lot for the review!

src/compose/paths.c
```diff
@@ -61,7 +61,7 @@ resolve_name(const char *filename, enum resolve_name_direction direction,
 
     xlocaledir = get_xlocaledir_path();
 
-    ret = snprintf(path, sizeof(path), "%s/%s", xlocaledir, filename);
+    ret = snprintf(path, sizeof(path), "%s" PATHSEPS "%s", xlocaledir, filename);
```
bluetech: According to the [Windows `fopen` documentation](https://docs.microsoft.com/en-us/previous-versions/yeby3zcb(v%3Dvs.140)#remarks), it does accept forward slash as path separator. Is it not so?

bluetech:
> I have no idea why test-filecomp is broken

Looking at [this log](https://ci.appveyor.com/project/aperezdc/libxkbcommon/builds/26483305/job/05r521nhbg9hfxx6/tests), it's indeed not very informative. Is it possible nothing is printed when an assert fails?

src/utils.c
aperezdc: Good idea, thanks for the tip!

src/utils.h
aperezdc: Indeed, thanks for the tip!

src/utils.c
aperezdc: The `va_list` in this case is used twice, and its value after the first call to `vsnprintf()` is undefined, so I will need to use `va_copy()` to pass a copy. Shouldn't be a problem :thinking:

src/compose/paths.c
aperezdc: True! I missed that part of the documentation, and I will be more than happy to see the hacky commit that tried to use the backslashes as path separators go away :smile:

bluetech:
@aperezdc Please let me know when this is ready -- github doesn't notify on pushes so I'm not sure.

I wonder why the CI doesn't run on the PR, maybe because it was created before we enabled it?

bluetech:
I have cherry-picked some general commits from here to master, and will try to get the rest to pass CI in a separate PR.

PR #104, opened 2019-08-06
CI: Use the Azure Pipelines service

aperezdc:
As requested by @bluetech in #103 I have been investigating on using Azure Pipelines for the continuous integration, and it looks like I more or less have been able to replicate what Travis is currently doing:

- Building on Ubuntu:
  - Clang+Autotools
  - Clang+Meson+Valgrind
  - GCC+Autotools
  - GCC+Meson+Valgrind
- Building on Mac OS:
  - Clang+Meson

The only differences with the current Travis configuration are:

- The most recent Ubuntu version Azure has is Xenial, so I had to pass `-Denable-wayland=false` for the Meson builds because the Wayland development packages are too old.
- Mac OS gets an update from 10.2 to 10.13
- Output from the tests is gathered from Meson's JSON log files an converted to JUnit XML, which Azure can ingest to show the test results in the Web interface.

Note that I have been learning to use the Azure Pipelines service while working on this patch set, so it could be as well that some things could be done better. I kept multiple commits to show the evolution of the configuration files as features were being added and improved upon, hopefully that will help out reviewing the PR. The conversion script for the unit test output was thrown together in a rush, so it might be possible to improve on it—the result is already quite nice, though! :eyes: 

You can see a [sample build on Azure Pipelines](https://dev.azure.com/aperezdc/libxkbcommon-ci/_build/results?buildId=50) running on my copy of the repository.

### DIFF
--- a/.azure-pipelines/steps/autotools.yml
+++ b/.azure-pipelines/steps/autotools.yml
@@ -1,0 +1,23 @@
+---
+parameters:
+  options: ""
+  workdir: "autotools-build"
+
+steps:
+  - bash: |
+      mkdir '${{ parameters.workdir }}' && cd "$_"
+      ../autogen.sh ${{ parameters.options }}
+    displayName: 'Configuration (Autotools)'
+  - bash: make -j$(nproc)
+    displayName: 'Build (Autotools)'
+    workingDirectory: ${{ parameters.workdir }}
+  - bash: make check
+    displayName: 'Tests (Autotools)'
+    workingDirectory: ${{ parameters.workdir }}
+  - bash: |
+      shopt -s nullglob
+      for file in "$(pwd)"/*.log ; do
+        echo "##vso[task.uploadfile]${file}"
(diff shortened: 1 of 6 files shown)


bluetech:
This is excellent. I looked over everything and I don't have any comments really. The setup is a bit more complex than the Travis one, but also cleaner with the template construct and more capable with the junit reporting.

I feel bad that I forgot to mention that the autotools build is deprecated and I intend to remove maybe even for the next release -- would have saved you some work. But it's good that it's working.

Regarding the junit XML script, do you think that it's something meson itself would like to support? Though maybe it's better for it to just output one format (JSON) and let each project adapt it to their wishes.

I'll try to register this repo to Azure Pipelines now. I might need to poke the branch to trigger it, let's see.

bluetech:
I'll be damned if I can understand how to make it work. I think it might starting working from master, I'll merge and hope for the best.

PR #106, opened 2019-08-09
Set up CI with Azure Pipelines

bluetech:


fooishbar:
I don't speak Azure, but this looks good to me - I've been hoping to get off Travis for a while. The cross-OS coverage is really nice too. Thanks both!

fooishbar:
Would be nice to do test-suite runs with ASan and Valgrind too - Meson has easy support for both.

bluetech:
OK, I think removing this trigger does the trick (this PR was generated from the azure website).

@fooishbar all thanks go to  @aperezdc  :)

I think the tests do run under Valgrind on Linux. Does ASan (and the other sanitizers) catch something Valgrind doesn't?

azure-pipelines.yml
```diff
@@ -1,7 +1,4 @@
 ---
-trigger:
```
aperezdc: Right, I forgot to mention that this would have been to be removed to build all branches, sorry for the omission—and thanks for merging the patch series :smile:

PR #108, opened 2019-09-03
Allow for custom rulesets through include files.

### LINKED ISSUE #99: Add wildcard parsing for options
A common use-case (I think) for keyboard layouts is to want to override one or two symbols only. i.e. what `xmodmap` provides, but we don't really have an answer to that. 

libxkbcommon parses from `$HOME/.xkb` already and `setxkbmap`/`xkbcommon` can do the same with `-I`.  Alas, this basically only works for layouts/variants thanks to the right wildcards but not for options. So this works:
```
$ cat $HOME/.xkb/symbols/custom 
partial alphanumeric_keys 
xkb_symbols "foo" {
    include "us(basic)"
    name[Group1]= "English (US)";
    key.type[group1]="ONE_LEVEL";
    key <TLDE> {        [      VoidSymbol ]       };
};
```
Together with:
```
$ setxkbmap -layout custom -variant foo -print | xkbcomp -I$HOME/.xkb - $DISPLAY
```
But that's limited in that only one can be selected at a time, you can't pick-and-match like you do with options. It would be better to be able to set this through:

```
$ setxkbmap -layout us -option "custom:foo"
```
But that doesn't work because we don't have a wildcard in the rules file for options. So what is needed in xkeyboard-config is something like this:
```
! option        =       symbols  
  .... all existing entries
  * = +%o
```
This requires parsing of `%o` in libxkbfile and libxkbcommon and the correct expansion. I've dabbled around with this in libxkbfile and the options are basically:

1. `* = +custom(%o)`, specified as `-option foo`. This would force users to have all options in `$HOME/.xkb/symbols/custom`.
1. `* = +%o`, specified as `-option custom(foo)`
1. `* = +%o` with auto-replacing `custom:foo` to `custom(foo)`. Nicest for integration, but highest risk of misparsing something.

If we go with this, we need to synchronise this well enough that we ship xkeyboard-config with the new rules file after we shipped all other related packages that shouldn't fail on the new wildcard match.

Opinions? Anything I forgot?

cc @svu, @fooishbar 

Somewhat related to #40

whot:
This adds an `! include /path/to/rulesfile` directive for rules files. The simplest case would be to ship `xkeyboard-config` with an extra line at the end in the form of:
```
! include %h/.xkb/rules/myrules
```

Then that file can contain something like this:
```
! option      = symbols
  custom:foo  = +custom(foo)
```
and resolve accordingly. This allows for simple additions of local configurations without having to modify the parsers too much. Conveniently, `xkbcomp` ignores the `! include` directive though this is still an API break for the rules files so distributions need to figure out what to do here. We do need to get this into xkbcomp/libxkbfile too though to have the same behaviour for wayland and xorg.

The directive is simple and currently only allows for `%h` to resolve to `$HOME`. Failures are basically ignored.

I've tested this under Wayland on a F30 and it works as expected: 
```gsettings set org.gnome.desktop.input-sources xkb-options "['terminate:ctrl_alt_bksp', 'compose:caps', 'custom:foo', 'custom:bar']"```

There's a bit of cleanup left and there may be some other state that I'm not aware of that needs to be considered. Needs more tests etc. but overall I think this is the best solution for #99 since it doesn't touch any of the hairer code and immediately gives us access to all of kcgcst.

This doesn't yet handle the `evdev.xml`/`evdev.lst` situation, not sure what to do here.

Fixes #99 

cc @svu, @bluetech, @fooishbar

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -285,6 +285,11 @@ test_env.set('XKB_LOG_LEVEL', 'debug')
 test_env.set('XKB_LOG_VERBOSITY', '10')
 test_env.set('top_srcdir', meson.source_root())
 test_env.set('top_builddir', meson.build_root())
+
+test_configh_data = configuration_data()
+test_configh_data.set_quoted('TEST_XKB_CONFIG_ROOT', join_paths(meson.source_root(), 'test', 'data'))
+configure_file(output: 'test-config.h', configuration: test_configh_data)
+
 # Some tests need to use unexported symbols, so we link them against
 # an internal copy of libxkbcommon with all symbols exposed.
 libxkbcommon_test_internal = static_library(
@@ -332,6 +337,11 @@ test(
 test(
     'rules-file',
     executable('test-rules-file', 'test/rules-file.c', dependencies: test_dep),
+    env: test_env,
+)
+test(
+    'rules-file-includes',
+    executable('test-rules-file-includes', 'test/rules-file-includes.c', dependencies: test_dep),
     env: test_env,
 )
 test(
--- a/src/context.c
+++ b/src/context.c
@@ -69,6 +69,13 @@ err:
     return 0;
 }
 
+const char *
+xkb_context_include_path_get_system_path(struct xkb_context *ctx)
+{
+    const char *root = secure_getenv("XKB_CONFIG_ROOT");
+    return root ? root : DFLT_XKB_CONFIG_ROOT;
+}
+
 /**
  * Append the default include directories to the context.
  */
@@ -106,11 +113,8 @@ xkb_context_include_path_append_default(struct xkb_context *ctx)
         }
     }
 
-    root = secure_getenv("XKB_CONFIG_ROOT");
-    if (root != NULL)
-       ret |= xkb_context_include_path_append(ctx, root);
-    else
-       ret |= xkb_context_include_path_append(ctx, DFLT_XKB_CONFIG_ROOT);
+    root = xkb_context_include_path_get_system_path(ctx);
+    ret |= xkb_context_include_path_append(ctx, root);
 
     return ret;
 }
--- a/src/xkbcomp/rules.c
+++ b/src/xkbcomp/rules.c
@@ -243,7 +243,6 @@ struct matcher {
     /* Input.*/
     struct rule_names rmlvo;
     union lvalue val;
-    struct scanner scanner;
     darray(struct group) groups;
     /* Current mapping. */
     struct mapping mapping;
@@ -332,7 +331,8 @@ matcher_group_start_new(struct matcher *m, struct sval name)
 }
 
 static void
-matcher_group_add_element(struct matcher *m, struct sval element)
+matcher_group_add_element(struct matcher *m, struct scanner *s,
+                          struct sval element)
 {
     darray_append(darray_item(m->groups, darray_size(m->groups) - 1).elements,
                   element);
@@ -369,7 +369,8 @@ extract_layout_index(const char *s, size_t max_len, xkb_layout_index_t *out)
 }
 
 static void
-matcher_mapping_set_mlvo(struct matcher *m, struct sval ident)
+matcher_mapping_set_mlvo(struct matcher *m, struct scanner *s,
+                         struct sval ident)
 {
     enum rules_mlvo mlvo;
     struct sval mlvo_sval;
@@ -383,14 +384,14 @@ matcher_mapping_set_mlvo(struct matcher *m, struct sval ident)
 
     /* Not found. */
     if (mlvo >= _MLVO_NUM_ENTRIES) {
-        scanner_err(&m->scanner, "invalid mapping: %.*s is not a valid value here; ignoring rule set",
+        scanner_err(s, "invalid mapping: %.*s is not a valid value here; ignoring rule set",
                     ident.len, ident.start);
         m->mapping.skip = true;
         return;
     }
 
     if (m->mapping.defined_mlvo_mask & (1u << mlvo)) {
-        scanner_err(&m->scanner, "invalid mapping: %.*s appears twice on the same line; ignoring rule set",
+        scanner_err(s, "invalid mapping: %.*s appears twice on the same line; ignoring rule set",
                     mlvo_sval.len, mlvo_sval.start);
         m->mapping.skip = true;
         return;
@@ -402,7 +403,7 @@ matcher_mapping_set_mlvo(struct matcher *m, struct sval ident)
         int consumed = extract_layout_index(ident.start + mlvo_sval.len,
                                             ident.len - mlvo_sval.len, &idx);
         if ((int) (ident.len - mlvo_sval.len) != consumed) {
-            scanner_err(&m->scanner, "invalid mapping: \"%.*s\" may only be followed by a valid group index; ignoring rule set",
+            scanner_err(s, "invalid mapping: \"%.*s\" may only be followed by a valid group index; ignoring rule set",
                         mlvo_sval.len, mlvo_sval.start);
             m->mapping.skip = true;
             return;
@@ -415,7 +416,7 @@ matcher_mapping_set_mlvo(struct matcher *m, struct sval ident)
             m->mapping.variant_idx = idx;
         }
         else {
-            scanner_err(&m->scanner, "invalid mapping: \"%.*s\" cannot be followed by a group index; ignoring rule set",
+            scanner_err(s, "invalid mapping: \"%.*s\" cannot be followed by a group index; ignoring rule set",
                         mlvo_sval.len, mlvo_sval.start);
             m->mapping.skip = true;
             return;
@@ -428,7 +429,7 @@ matcher_mapping_set_mlvo(struct matcher *m, struct sval ident)
 }
 
 static void
-matcher_mapping_set_kccgst(struct matcher *m, struct sval ident)
+matcher_mapping_set_kccgst(struct matcher *m, struct scanner *s, struct sval ident)
 {
     enum rules_kccgst kccgst;
     struct sval kccgst_sval;
@@ -442,14 +443,14 @@ matcher_mapping_set_kccgst(struct matcher *m, struct sval ident)
 
     /* Not found. */
     if (kccgst >= _KCCGST_NUM_ENTRIES) {
-        scanner_err(&m->scanner, "invalid mapping: %.*s is not a valid value here; ignoring rule set",
+        scanner_err(s, "invalid mapping: %.*s is not a valid value here; ignoring rule set",
                     ident.len, ident.start);
         m->mapping.skip = true;
         return;
     }
 
     if (m->mapping.defined_kccgst_mask & (1u << kccgst)) {
-        scanner_err(&m->scanner, "invalid mapping: %.*s appears twice on the same line; ignoring rule set",
+        scanner_err(s, "invalid mapping: %.*s appears twice on the same line; ignoring rule set",
                     kccgst_sval.len, kccgst_sval.start);
         m->mapping.skip = true;
         return;
@@ -461,15 +462,15 @@ matcher_mapping_set_kccgst(struct matcher *m, struct sval ident)
 }
 
 static void
-matcher_mapping_verify(struct matcher *m)
+matcher_mapping_verify(struct matcher *m, struct scanner *s)
 {
     if (m->mapping.num_mlvo == 0) {
-        scanner_err(&m->scanner, "invalid mapping: must have at least one value on the left hand side; ignoring rule set");
+        scanner_err(s, "invalid mapping: must have at least one value on the left hand side; ignoring rule set");
         goto skip;
     }
 
     if (m->mapping.num_kccgst == 0) {
-        scanner_err(&m->scanner, "invalid mapping: must have at least one value on the right hand side; ignoring rule set");
+        scanner_err(s, "invalid mapping: must have at least one value on the right hand side; ignoring rule set");
         goto skip;
     }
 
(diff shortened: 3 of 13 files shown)

### CODE REVIEW

**Comment by whot** [comment]
Follow-up thought: it may be interesting to do this the other way round, i.e. look up `$HOME/.xkb/`  first and have the `evdev` rules file there, including the system done. i.e. that file would look like this:
```
! include /usr/share/X11/xkb/rules/evdev

! option      = symbols
  custom:foo  = +custom(foo)
```
The obvious advantage here would be that the system file remains unchanged and backwards compatible. The problem is that xkbcommon loads the system ones first so they'll always win. The only way out of this is by using a different ruleset but `evdev` is basically hardcoded everywhere these days.

Changing the order may have unintended side effects, doubly so where the admin doesn't want to allow for user-specific XKB files.

**Comment by whot** [comment]
I had a look at how `evdev.xml` would be handled. The smartest choice would likely be to use XInclude in the `evdev.xml` file, so at the end of the file have something like:
```
  <xi:include href="$HOME/.xkb/rules/myrules.xml" />
```

with some handwaving regarding expanding `$HOME`. How's this going to work:

- GNOME: `libgnome-desktop` uses a [simple SAX parser](https://developer.gnome.org/glib/stable/glib-Simple-XML-Subset-Parser.html). No support for namespaces, with some hackery we could fake namespace support for XInclude though. Or just switch to a better parser that handles this.
- XFCE4: uses libxklavier to get the rules resolved.
- libxklavier: uses libxml's SAX2 parser which after a quick peek supports namespaces. But libxklavier hasn't seen changes since 2013.
- Plasma uses QXmlSimpleReader (a SAX parser).

AFAICT all of these would simply ignore any unsupported `<xi:include>` statements, so at least we don't have to worry about backwards compatibility.


As for `evdev.lst`/`base.lst`, I'm not sure who if anything uses that (other than the old keyboard driver's man page)

File: src/xkbcomp/rules.c [hunk 1]
```diff
@@ -343,6 +350,39 @@ matcher_group_add_element(struct matcher *m, struct sval element)
...
+                struct xkb_component_names *out);
+static void
+matcher_include(struct matcher *m, struct sval inc,
+                struct xkb_component_names *out)
+{
+    char *path = strndup(inc.start, inc.len);
+    bool ret;
+
+    /* %h at the beginning expands to $HOME */
+    if (strlen(path) > 2 && strncmp(path, "%h", 2) == 0) {
```
bluetech: If we interpret `%`-expansions, I'd prefer that `%` be matched separately, and if the following character is not recognized, a warning be issued. We should also allow a `%%` to expand to `%`, I suppose.

The Compose parser [does this](https://github.com/xkbcommon/libxkbcommon/blob/934d5741727886fe90e9c2a4d90e63bbd2cb2226/src/compose/parser.c#L258-L296).

whot:
Thanks for the comments, much appreciated.

> Isn't this backwards? What I would expect is, that given an evdev RMLVO, that ~/.xkb/rules/evdev would be considered before the system one, 

[I agree](https://github.com/xkbcommon/libxkbcommon/pull/108#issuecomment-527319887), but it's a more aggressive change than the other one. There's one argument against that: an administrator that wants to prevent users from loading custom keymaps can `! include` from the system files. That is not as easy when we load `$HOME` first, we'd need some config that specifies the load order. I don't know how common that scenario would be, realistically three are only two, three xkb options that are dangerous (terminate server and grab breaking/debugging. And terminate-server at least has a xorg.conf option to stop it).

> The user would use a different R

I think evdev is hardcoded in too many places that we're years off from having this work reliably. 

But it would be useful changing to `! include %h/.xkb` (i.e. just the basedir) and then loading the rules file from $filename. This allows for changing rules in the future.

> >  Also note the bar/baz above the rule doesn't have to match the component names.

>  I couldn't figure out what this refers to - can you clarify?

Oh, simply that your rule could be this:

```
! option = symbols
  bar:baz = +us(dvorak)
```
So unlike the wildcards we talked about earlier there is no hard requirement for matching names between the user-visible rules option and the actual components being loaded. This is true for anything in rules anyway, just see this as comparison against the wildcard approach from #99.

> my blah blah regarding `evdev.xml`

random thought: why do we have a separate file here instead of relying on libxkbcommon? does it make sense to add the right API to xkbcommon to provide that info so we don't have different xml parsers everywhere?

src/xkbcomp/rules.c
```diff
@@ -243,7 +243,9 @@ struct matcher {
     /* Input.*/
     struct rule_names rmlvo;
     union lvalue val;
-    struct scanner scanner;
+    struct scanner *scanner; /* points to one in the scanner_stack */
+    struct scanner scanner_stack[16];
```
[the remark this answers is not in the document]
whot: No particular preference, whatever's your preference and can be implemented. The scanner stack was just the easiest one to write :)

src/xkbcomp/rules.c
```diff
@@ -153,6 +158,7 @@ enum rules_mlvo {
     MLVO_LAYOUT,
     MLVO_VARIANT,
     MLVO_OPTION,
+    MLVO_INCLUDE,
```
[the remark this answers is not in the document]
bluetech: Hmm, I am not sure myself why this is added here?

whot:
> Hmm is this still possible? I vaguely remember some CVE about this, after which it was no longer possible.

It used to be in the default keymap and now it's behind an option, but still accessible (at least grab debugging, haven't tested grab breaking recently).

> I've thought about it -- there are actually two files, evdev.lst and evdev.xml.

Well... `evdev.lst` predates fancy things like XML. [one of the first xkeyboard-config commits](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/commit/9bbda6bfbcc29a575ea3b5cb7bd2553767e80d53) describes them as  `The formerly used descriptive files were named <rules>.lst [...]`. And that was 15 years ago. In xkeyboard-config, the .lst files are generated from .xml files.

I assume that `localectl` reads the `.lst` files because they're easier to parse than the xml equivalents. ping @poettering who wrote that bit originally.

I don't think we have much of a choice for libxkbcommon though, we need to use the xml files because they're the upstream-sanctioned sources. We're not just pure consumers of it, with the change suggested here we're starting down the road of expecting users to write their own XKB configuration, including the `evdev.xml` snippet if they want it to show up in the GUI. And it has to be the XML snippet because that's the upstream data format (which also means we need a converter tool so the lst file is generated accordingly). If we pick `lst` as format for libxkbcommon we're at odds with xkeyboard-config.

> Unless anyone objects (@whot, @fooishbar), I will make the change to move ~/.xkb to be first in the default search path in master.

ACK from me, thanks.

src/xkbcomp/rules.c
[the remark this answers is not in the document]
whot: detritus leftover from an earlier version, will be removed in the next version of the patch. Sorry about the noise.

src/xkbcomp/rules.c
```diff
@@ -897,6 +939,15 @@ matcher_match(struct matcher *m, const char *string, size_t len,
         goto unexpected;
     }
 
+include_statement:
+    switch (tok = gettok(m)) {
+        case TOK_IDENTIFIER:
```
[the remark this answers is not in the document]
whot: fixed locally, thanks.

src/xkbcomp/rules.c
```diff
@@ -999,6 +999,37 @@ matcher_match(struct matcher *m, const char *string, size_t len,
     return false;
 }
 
+static int
```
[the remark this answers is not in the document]
whot: fixed locally, thanks.

fooishbar:
> Unless anyone objects (@whot, @fooishbar), I will make the change to move ~/.xkb to be first in the default search path in master.

No objection here, thanks.

bluetech:
I moved ~/.xkb to be first in the search path in 3a91788d9254b4571ff883439931f497d8a0663f.

whot:
cool, thanks. fwiw, I talked about this with a few people at XDC and the first thing that came up: this shouldn't be in `$HOME/.xkb` but `$XDG_CONFIG_HOME/xkb` - and tbh I agree. Is there some historical requirement for `$HOME/.xkb`? can we make it the second thing after the xdg dir to look into?

bluetech:
I think XDG would be nice, but I'd like to deprecate or remove ~/.xkb if we start using XDG. (I don't believe ~/.xkb has a lot of users, so won't be too bad).

Would we consider only XDG_CONFIG_HOME or also XDG_CONFIG_DIRS?

Does XDG work on MacOS and Windows?

Maybe we should discuss this in another issue.

whot:
ok, I've stared at this long enough that I don't see any bugs anymore, even if they're present. Any review would be much appreciated.

Quite a few changes but all address previous review comments. A few cleanup patches first with little or no functional changes. The scanner is on the stack now instead of the previous push/pop approach and we have test cases. Nesting limit is 5 (previously: 16) because you really shouldn't ever need more than 2 anyway.

Since we now have #112 the whole thing turns on its head: rather than shipping xkeyboard-config changes that may break other clients we can just tell users to put a `$XDG_CONFIG_HOME/xkb/rules/evdev` files in place with their custom bits and an `! include %S/evdev` line and done. 

That works but not for the primary use-case: gnome-shell. It is suid root, `secure_getenv()` returns NULL for `$HOME`. So we can't actually use this in the shell. I've verified including things work in the system files but that's not really the goal of the game. Not sure how to deal with that situation now tbh.

whot:
I guess the only real choice we have here is to get mutter/gnome-shell to set the search path manually. mutter PR submitted: https://gitlab.gnome.org/GNOME/mutter/merge_requests/936

fooishbar:
> That works but not for the primary use-case: gnome-shell. It is suid root

Why? Using logind means you don't have to be root.

whot:
true, thanks for pointing this out. Clearly I need glasses, the `ls` colours were almost the same to a setuid executable but not exactly (white on red vs yellow on red...) and I didn't check.

Anyway, `gnome-shell` has `cap_sys_nice`, needed for the [mutter rt-scheduler](https://gitlab.gnome.org/GNOME/mutter/commit/dae2c1d420ed272710ac55b7a00f6787e5c0e762). Not enabled by default upstream according to that commit but Fedora 31 sets the cap in its `.spec` file. I've updated my comment above accordingly.

Overall, same outcome here, if this become standard upstream `secure_getenv()` is worthless to us a common use-case.

src/xkbcomp/rules.c
```diff
@@ -1059,15 +1065,14 @@ static bool
...
                 const char *path)
 {
     bool ret = false;
-    FILE *file;
     char *string;
     size_t size;
     struct scanner scanner;
 
-    file = fopen(path, "r");
     if (!file)
```
bluetech: I think it should be up to the caller to guarantee this, so should remove the check here.

src/xkbcomp/rules.c
```diff
@@ -408,9 +409,14 @@ matcher_include(struct matcher *m, struct scanner *parent_scanner,
         return;
     }
 
-    ret = read_rules_file(m->ctx, m, include_depth + 1, s.buf);
-    if (!ret)
-        log_err(m->ctx, "No components returned from included XKB rules \"%s\"\n", s.buf);
+    file = fopen(s.buf, "r");
```
bluetech: Looks like this file is not closed anywhere now?

meson.build
```diff
@@ -288,6 +288,11 @@ test_env.set('top_builddir', meson.build_root())
 test_env.set('MALLOC_PERTURB_', '15')
 test_env.set('MallocPreScribble', '1')
 test_env.set('MallocScribble', '1')
+
```
bluetech: I introduced a conflict here in master (sorry) - needs a rebase.

src/xkbcomp/rules.c
```diff
@@ -131,6 +134,10 @@ lex(struct scanner *s, union lvalue *val)
         return TOK_GROUP_NAME;
     }
 
+    /* Include statement. */
+    if (lit(s, "include"))
```
bluetech: In the commit message you wrote "The file path must be without quotes and may not start with the
literal "include"." - is it because we now treat it a a token?

src/context.c
```diff
@@ -69,6 +69,13 @@ xkb_context_include_path_append(struct xkb_context *ctx, const char *path)
     return 0;
 }
 
+const char *
+xkb_context_include_path_get_system_default(struct xkb_context *ctx)
```
bluetech: I think we should remove the "default" from the name - it is the system path itself, the default system path would be `DFLT_XKB_CONFIG_ROOT` I suppose.

bluetech:
Regarding the `secure_gettenv`, I don't think it should block this PR. We can discuss separately if needed. Maybe reconsider the threat model and change back to `getenv`? (Not sure).

src/xkbcomp/rules.c
whot: that and - I freely admit - laziness. We could allow for quotes, but then we'd have to worry about possible escaping, etc. Just easier to treat anything after the literal `include` as-is.

meson.build
whot: easy fix

src/xkbcomp/rules.c
whot: ack, removed.

src/context.c
whot: renamed to `xkb_context_include_path_get_system_path()` because I can't think of a better name. Maybe santa brings me an epiphany for christmas ;)

whot:
Marked as ready now, all comments addressed except for the "quotes around paths". As for the `secure_getenv()`, yeah, we can revisit that separately. Filed as #129 now. Thanks for all the reviews btw!

bluetech:
Merged, thanks @whot!

PR #109, opened 2019-09-10
build: Solaris needs __EXTENSIONS__ instead of _GNU_SOURCE

alanc:
Fix meson build on Solaris by using __EXTENSIONS__ where Linux & other
platforms use _GNU_SOURCE.  Without this the build fails due to missing
prototypes for functions like strdup & getopt not defined in the C99
standard.  (In autoconf, this was handled by AC_USE_SYSTEM_EXTENSIONS.)

Signed-off-by: Alan Coopersmith <alan.coopersmith@oracle.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -61,7 +61,15 @@ endif
 
 # config.h.
 configh_data = configuration_data()
-configh_data.set('_GNU_SOURCE', 1)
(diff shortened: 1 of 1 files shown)


bluetech:
Thanks. We should do a release soon.

PR #115, opened 2019-10-20
context: Don't fail to create the context if HOME isn't available

heftig:
E.g. when Mutter has `CAP_SYS_NICE` and thus `secure_getenv` returns `NULL`.

Fixes https://bugs.archlinux.org/task/64191

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -81,13 +81,13 @@ xkb_context_include_path_append_default(struct xkb_context *ctx)
     int ret = 0;
 
     home = secure_getenv("HOME");
-    if (!home)
-        return ret;
-    err = asprintf(&user_path, "%s/.xkb", home);
-    if (err <= 0)
-        return ret;
-    ret |= xkb_context_include_path_append(ctx, user_path);
-    free(user_path);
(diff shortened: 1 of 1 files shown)


src/context.c
```diff
@@ -81,13 +81,13 @@ xkb_context_include_path_append_default(struct xkb_context *ctx)
...
-        return ret;
-    err = asprintf(&user_path, "%s/.xkb", home);
-    if (err <= 0)
-        return ret;
-    ret |= xkb_context_include_path_append(ctx, user_path);
-    free(user_path);
+    if (home != NULL) {
+        err = asprintf(&user_path, "%s/.xkb", home);
+        if (err <= 0)
+            return ret;
```
bluetech: Let's not return here, but ignore. It doesn't make sense to return 0 (success) without doing anything.

heftig: Should I make a new commit or will you fix it?

bluetech:
Committed to master with the fix.

bluetech:
0.9.1 is released with this fix.

PR #116, opened 2019-10-21
Add meson option to specify a native wayland-scanner

giordano:
When Wayland is enabled, it's not possible to cross-compile this library because it requires a native `wayland-scanner` that can be run on the build system.

Instead of looking for `wayland-scanner` in the pkgconfig file, this PR allows setting from the command line the path (or just the name if it's in the `PATH`) to a native `wayland-scanner` that can be used for the build.

I've used this patch to successfully cross-compile this library for 9 different platforms using Julia's [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl):  https://github.com/JuliaPackaging/Yggdrasil/pull/147

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -423,7 +423,7 @@ if get_option('enable-wayland')
 You can disable the Wayland demo programs with -Denable-wayland=false.''')
     endif
 
-    wayland_scanner = find_program(wayland_scanner_dep.get_pkgconfig_variable('wayland_scanner'))
+    wayland_scanner = find_program(get_option('wayland-scanner'))
     wayland_scanner_code_gen = generator(
         wayland_scanner,
         output: '@BASENAME@-protocol.c',
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -56,3 +56,9 @@ option(
     value: true,
     description: 'Enable support for Wayland utility programs',
 )
+option(
+  'wayland-scanner',
+  type: 'string',
+  value: 'wayland-scanner',
+  description: 'Path to native wayland-scanner',
+)


meson.build
```diff
@@ -423,7 +423,7 @@ if get_option('enable-wayland')
 You can disable the Wayland demo programs with -Denable-wayland=false.''')
     endif
 
-    wayland_scanner = find_program(wayland_scanner_dep.get_pkgconfig_variable('wayland_scanner'))
+    wayland_scanner = find_program(get_option('wayland-scanner'))
```
whot: doesn't this break the previous build as the option now **must** be specified for all builds? This should fallback to the pkgconfig variable since that's the most common use-case.

giordano: > doesn't this break the previous build as the option now must be specified for all builds?

There is a default value, so you don't have to specify for all builds.  In native builds, if `wayland-scanner` is available it's likely in the `PATH`, without having to fiddle with the pkgconfig file.  BTW, with my pkgconfig file I couldn't find `wayland-scanner` anyway, so that looking into that file was quite useless

fooishbar: Wayland upstream installs a `wayland-scanner.pc` which very definitely has the literal path to the pkg-config path. That works just fine.

There is more discussion on the issue [in weston!157](https://gitlab.freedesktop.org/wayland/weston/merge_requests/157).

bluetech:
Truth be told, I am quite confused by cross builds. But, if you do have a native wayland install, and our meson file specifies `native: true`,

https://github.com/xkbcommon/libxkbcommon/blob/90497b84f88656ec151076644c5c69c3507ca3a9/meson.build#L420

then why is it not found? (Basically how the Weston link @fooishbar provided starts out).

BTW, wayland is only used for a test program, libxbkcommon doesn't actually depend on it in any way. So if it is causing you trouble, you shouldn't feel bad about just disabling it.

giordano:
I hadn't the time to get back to this PR, hopefully in the next few days, for the time being let me reply to some comments.

> then why is it not found? 

`pkg-config` is not looking where the native tools are, as I want it to pick-up the cross-build environment during cross-compilation.  I don't want it to mess up with the native environment.

> BTW, wayland is only used for a test program, libxbkcommon doesn't actually depend on it in any way. So if it is causing you trouble, you shouldn't feel bad about just disabling it.

Sure, but there is currently no way to skip tests without patching `meson.build`.  Adding a meson option to skip tests was my other idea, but this is slightly orthogonal.

For the record, wayland itself has a configure option to indicate a native `wayland-scanner` for the build: https://gitlab.freedesktop.org/wayland/wayland/blob/8e0513410dd16a3f3b1d07745b613e5ae6141bd0/configure.ac#L80

bluetech:
> pkg-config is not looking where the native tools are, as I want it to pick-up the cross-build environment during cross-compilation. I don't want it to mess up with the native environment.

Right, but since we are passing `native: true` then meson should look at the native pkg-config path, not the cross one, shouldn't it? It'd be interesting to see the meson log.

> Sure, but there is currently no way to skip tests without patching meson.build.

Right, I didn't mean disable the tests, I meant disable wayland (`meson setup -Denable-wayland=false`). The options are defined in [`meson_options.txt`](https://github.com/xkbcommon/libxkbcommon/blob/master/meson_options.txt).

bluetech:
I'm going to close this now, but feel free to reopen if you want to continue.

PR #117, opened 2019-10-25
Use XDG_CONFIG_HOME as first XKB search path

### LINKED ISSUE #112: Use XDG_CONFIG_HOME for user-specific files
This came up in https://github.com/xkbcommon/libxkbcommon/pull/108#issuecomment-538717498, this issue is to keep track of that part.

And to answer the question there: I think we can ignore `$XDG_CONFIG_DIRS` for now and add that when someone requires it, using `$XDG_CONFIG_HOME` is enough for now.

whot:
Use $XDG_CONFIG_HOME/xkb as the primary lookup path for XKB rules. Same
motivation as in 3a91788d9254b, however the XDG directories are more standard
and recommended these days than application-specific dotfiles.

The XDG spec says to fall back to $HOME/.config where XDG_CONFIG_HOME is not
set so we implement that behavior as well.

Fixes #112

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -75,12 +75,29 @@ err:
 XKB_EXPORT int
 xkb_context_include_path_append_default(struct xkb_context *ctx)
 {
-    const char *home, *root;
+    const char *home, *xdg, *root;
     char *user_path;
     int err;
     int ret = 0;
 
     home = secure_getenv("HOME");
+
+    xdg = secure_getenv("XDG_CONFIG_HOME");
+    if (xdg != NULL) {
+        err = asprintf(&user_path, "%s/xkb", xdg);
+        if (err >= 0) {
+            ret |= xkb_context_include_path_append(ctx, user_path);
+            free(user_path);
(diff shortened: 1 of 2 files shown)


bluetech:
Thanks @whot.

The implementation is a bit confused though, it calls `home = secure_getenv("HOME");` twice and adds `~/.xkb` to the search path twice.

I also think that in terms of reading the code, it will be clearer if the `~/.config/` fallback is in an `else` block of the `XDG_CONFIG_HOME` check. If it works out.

whot:
> adds ~/.xkb to the search path twice.

doh, that whole block should've been removed, copy/paste error. Changed and updated as requested with a follow-up patch that checks for the various include paths to be correct/present.

whot:
Ok, test errors are fixed, cause ended up that on macos the default xkb path does not exist - but it appears to exist on the others. Something to keep in mind maybe.

bluetech:
Thanks!

PR #119, opened 2019-10-31
scripts/makeheader: allow overriding the prefix path of the X11 headers

swick:
with X11_HEADERS_PREFIX

Signed-off-by: Sebastian Wick <sebastian@sebastianwick.net>


bluetech:
Thanks!

PR #121, opened 2019-11-07
two coverity fixes

whot:
Coverity actually complains about a few others but many of those are false positives. And others are largely in the test code - these two are the only ones that I could justify writing a patch for (i.e. where it doesn't get more convoluted afterwards). And IMO only the memleak patch is really one that's worth merging.


bluetech:
Thanks!

PR #125, opened 2019-11-18
test: let rmlvo-to-kccgst take long options like rmlvo-to-keymap

whot:
The short options were left for backwards compatibility.


bluetech:
Thanks!

PR #132, opened 2020-02-06
A new API to query modifier set required to type a keysym

j-kubik:
The new API is useful to implement features like auto-type and
desktop automation. Since the inputs for these features is usually
specified in terms of the symbols that need to be typed, the
implementation needs to be able to invert the keycode->keysym->text
transformation and produce a sequence of keycodes that can be used
to type the requested character(s).

### DIFF
--- a/src/keymap.c
+++ b/src/keymap.c
@@ -409,6 +409,39 @@ xkb_keymap_led_get_index(struct xkb_keymap *keymap, const char *name)
     return XKB_LED_INVALID;
 }
 
+XKB_EXPORT size_t
+xkb_keymap_key_get_mods_for_level(struct xkb_keymap *keymap,
+                                  xkb_keycode_t kc,
+                                  xkb_layout_index_t layout,
+                                  xkb_level_index_t level,
+                                  xkb_mod_mask_t *masks_out,
+                                  size_t masks_size)
+{
+    const struct xkb_key *key = XkbKey(keymap, kc);
+    if (!key)
+        return 0;
+
+    layout = XkbWrapGroupIntoRange(layout, key->num_groups,
+                                   key->out_of_range_group_action,
+                                   key->out_of_range_group_number);
+    if (layout == XKB_LAYOUT_INVALID)
+        return 0;
+
+    if (level >= XkbKeyNumLevels(key, layout))
+        return 0;
+
+    const struct xkb_key_type *type = key->groups[layout].type;
+
+    size_t count = 0;
+    for (unsigned i = 0; i < type->num_entries && count < masks_size; i++)
+        if (entry_is_active(&type->entries[i]) &&
+            type->entries[i].level == level) {
+            masks_out[count] = type->entries[i].mods.mask;
+            ++count;
+        }
+    return count;
+}
+
 /**
  * As below, but takes an explicit layout/level rather than state.
  */
--- a/test/keymap.c
+++ b/test/keymap.c
@@ -38,6 +38,11 @@ main(void)
     struct xkb_keymap *keymap;
     xkb_keycode_t kc;
     const char *keyname;
+    xkb_mod_mask_t masks_out[4] = { 0, 0, 0, 0 };
+    size_t mask_count;
+    xkb_mod_mask_t shift_mask;
+    xkb_mod_mask_t lock_mask;
+    xkb_mod_mask_t mod2_mask;
 
     assert(context);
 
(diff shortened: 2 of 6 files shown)


bluetech:
Hi @j-kubik,

Thanks for the patches.

First, please separate the 2nd patch (`xkb_utf32_to_keysym`) to another PR. GitHub PRs don't support any sort of threading, so using different PRs keeps things organized. I'll only discuss the first patch here.

The functionality seems useful. It is a hole in the information currently exposed by the library. The only way to get it with the current API in a future-proof manner would be to go over all 2**32 modifier combinations and try each one, which is not reasonable.

> Since the inputs for these features is usually specified in terms of the symbols that need to be typed, the implementation needs to be able to invert the keycode->keysym->text transformation and produce a sequence of keycodes that can be used to type the requested character(s).

I can see the use case here. Of course, it is much better to bypass the keyboard entirely and avoid simulating key presses and such, and send the text directly instead. Most UI frameworks allow that. But I assume it is not possible in your case?

Another thing is, once you get the required mod-mask that would activate some specific level, how do you know which keys to hold to generate the mask itself?

src/keymap.c
```diff
@@ -409,6 +409,39 @@ xkb_keymap_led_get_index(struct xkb_keymap *keymap, const char *name)
...
+    if (level >= XkbKeyNumLevels(key, layout))
+        return 0;
+
+    const struct xkb_key_type *type = key->groups[layout].type;
+
+    size_t count = 0;
+    for (unsigned i = 0; i < type->num_entries && count < masks_size; i++)
+        if (entry_is_active(&type->entries[i]) &&
+            type->entries[i].level == level) {
+            masks_out[count]=type->entries[i].mods.mask;
```
bluetech:
```suggestion
            masks_out[count] = type->entries[i].mods.mask;
```

xkbcommon/xkbcommon.h
```diff
@@ -1140,6 +1155,48 @@ xkb_level_index_t
 xkb_keymap_num_levels_for_key(struct xkb_keymap *keymap, xkb_keycode_t key,
                               xkb_layout_index_t layout);
 
+/**
+ * Retrieves every possible modifier mask that produces the specified
+ * shift level for a specific key and layout.
+ *
+ * This API is useful for inverse key transformation; ie. finding out
```
bluetech:
```suggestion
 * This API is useful for inverse key transformation; i.e. finding out
```

xkbcommon/xkbcommon.h
```diff
@@ -1140,6 +1155,48 @@ xkb_level_index_t
...
+ *
+ * @warning It returns only up to masks_size modifier masks. If the
+ * buffer passed is too small, some of the possible modifier combinations
+ * will not be returned.
+ *
+ * @param[in] keymap      The keymap.
+ * @param[in] key         The keycode of the key.
+ * @param[in] layout      The layout for which to get modifiers.
+ * @param[in] level       The shift level in the layout for which to get the
+ * modifiers. This must be smaller than:
```
bluetech:
```suggestion
 * modifiers. This should be smaller than:
```

src/keymap.c
bluetech: I wonder if it would be beneficial to guarantee some particular order, "shortest first" or such?

j-kubik: Not sure whether shortest mod-mask is actually interesting. What would be interesting is the shortest sequence of key-presses that produce the modmask for the specified level, but that cannot be computed here.

j-kubik:
Our project is part of a wayland compositor and the core wayland protocol allows sending keycodes only. Alternatively, we could propose a wayland protocol extension that would allow sending such inputs directly, but that would depend on adoption and implementation by all possible UI toolkits, which is unlikely to happen. So sending the correct key-codes is the only way to go.

As for the required mod-mask: at the moment we iterate the entire keymap using `xkb_keymap_key_for_each` on initialization. We create a `xkb_state`, and for each key we reset it,
send the key press and release and observe the mod-masks changes. This is rather sub-optimal and assumes that modifiers work completely independently - that is the most common configuration. I was thinking about another PR to add a feature that would do better, but I am not yet sure what would be the best form for such a query. The question boils down to: using the xkb_state state machine, compute any (or maybe even the shortest) path of state transitions that leads to a particular sub-set of states - because the full state consists of depressed, latched and locked mod-masks but for the final keypress that needs that state only the effective mod-mask is important. Such an algorithm is definitely possible, but I am not sure whether `libxkbcommon` is the right place for it. Preferably I would just expose the state transition table in some usable form and allow the library users deal with it as required. What is your opinion?

I will split the PR and re-submit it - preferably through an account linked to my company.

xkbcommon/xkbcommon.h
j-kubik: The doc text was copied from the doc for the nearby `xkb_keymap_key_get_syms_by_level` function. Should it not remain consistent?

fooishbar:
@j-jubik I disagree that it's the only way to go. As I said to dcz in the parallel Wayland PR, either every compositor implements this support for faking key events, or all the clients implement support for doing it properly. The latter is obviously a lot preferable, and they both seem like about the same amount of effort.

xkbcommon/xkbcommon.h
bluetech: I pushed a commit to master which changes it there. It's allowed to try an out of range level (not an error), just not very sensible.

j-kubik:
It is the only way if I do not have any control over the clients. When clients provide a way to deliver input in a direct way, we already utilize it. But some clients just wont play ball. So sending keycodes through wayland core protocol is the only solution that works without the need to modify every single client out there.

Furthermore, since our project is all about software testing, it stands to a reason that it should send the inputs in the exact same way a human user would. Relying on a side channel to feed the input directly to the client/UI toolkit leaves parts of the tested software stack unused.

bluetech:
Two more comments (sorry I didn't mention before):

1. I prefer the name `xkb_keymap_key_get_mods_for_level` (by -> of). The "by" in e.g. `xkb_keymap_key_get_syms_by_level` really do get the syms *of* the level, while in this case the mods are what *triggers* the level.

2. It'd be great to have some tests for this new functionality. Can you write some? There's not much to our test suite - just `assert`s in a file, run by `ninja test`. Can add the checks in a new function in the `test/keymap.c` file.

test/keymap.c
```diff
@@ -38,6 +38,11 @@ main(void)
     struct xkb_keymap *keymap;
     xkb_keycode_t kc;
     const char *keyname;
+    xkb_mod_mask_t masksOut[4] = { 0, 0, 0, 0 };
```
bluetech:
```suggestion
    xkb_mod_mask_t masks_out[4] = { 0, 0, 0, 0 };
```

test/keymap.c
```diff
@@ -38,6 +38,11 @@ main(void)
...
+    size_t maskCount;
```
bluetech:
```suggestion
    size_t mask_count;
```

test/keymap.c
```diff
@@ -59,6 +64,42 @@ main(void)
     keyname = xkb_keymap_key_get_name(keymap, kc);
     assert(streq(keyname, "COMP"));
 
+    kc = xkb_keymap_key_by_name(keymap, "AC01");
+
+    // AC01 level 0 ('a') requires no modifiers on us-pc104
+    maskCount = xkb_keymap_key_get_mods_for_level(keymap, kc, 0, 0, masksOut, 10);
```
bluetech: The size of `masksOut` is 4, so the 10 should be 4, no?

test/keymap.c
```diff
@@ -59,6 +64,42 @@ main(void)
     keyname = xkb_keymap_key_get_name(keymap, kc);
     assert(streq(keyname, "COMP"));
 
+    kc = xkb_keymap_key_by_name(keymap, "AC01");
```
bluetech: Add an assert that it's not `XKB_KEYCODE_INVALID`.

bluetech:
Thanks for adding tests @j-kubik. The code looks good to me now.

@fooishbar WDYT?

fooishbar:
To be honest, I still think the usecase is somewhat broken, but the code at least seems fine.

> Another thing is, once you get the required mod-mask that would activate some specific level, how do you know which keys to hold to generate the mask itself?

I assume it would be easier to just forcibly reset the modmask by hand.

bluetech:
> To be honest, I still think the usecase is somewhat broken, but the code at least seems fine.

I am not sure about the particular use case myself, but I do think it is something that should be *possible* to do. Currently there is no way to reverse engineer programmatically which modifiers can/should be set to activate a certain level. When using some unfamiliar keymaps school, I personally would have found it useful to have a tool which, given an RMLVO + a Unicode character, tells me which layout, keycode, level and modifiers I can use to generate it. That's why I'm inclined to merge it.

In fact, before merging, I should actually write that tool to verify the API.

bluetech:
Thanks @j-kubik! I rebased to fix a conflict in the symbol map file and merged.

I also added a demo program which uses this (as discussed in previous comment): #137.

PR #134, opened 2020-02-16
Support translation Unicode codepoints to keysyms

j-kubik:
In order to support features like auto-type and UI automation, the
relevant tools need to be able to invert the keycode->keysym->text
transformation. In order to facilitate that, a new API was added.
It allows querying the keysyms that correspond to particular Unicode
codepoints. For all practical purposes, it can be thought of as an
inverse of xkb_keysym_to_utf32().

### DIFF
--- a/src/keysym-utf.c
+++ b/src/keysym-utf.c
@@ -897,6 +897,35 @@ xkb_keysym_to_utf32(xkb_keysym_t keysym)
     return bin_search(keysymtab, ARRAY_SIZE(keysymtab) - 1, keysym);
 }
 
+XKB_EXPORT xkb_keysym_t
+xkb_utf32_to_keysym(uint32_t ucs)
+{
+    /* first check for Latin-1 characters (1:1 mapping) */
+    if ((ucs >= 0x0020 && ucs <= 0x007e) ||
+        (ucs >= 0x00a0 && ucs <= 0x00ff))
+        return ucs;
+
+    /* special keysyms */
+    if ((ucs >= (XKB_KEY_BackSpace & 0x7f) && ucs <= (XKB_KEY_Clear & 0x7f)) ||
+        ucs == (XKB_KEY_Return & 0x7f) || ucs == (XKB_KEY_Escape & 0x7f))
+        return ucs | 0xff00;
+    if (ucs == (XKB_KEY_Delete & 0x7f))
+        return XKB_KEY_Delete;
+
+    /* Unicode non-symbols and code points outside Unicode planes */
+    if ((ucs >= 0xfdd0 && ucs <= 0xfdef) ||
+        ucs > 0x10ffff || (ucs & 0xfffe) == 0xfffe)
+        return XKB_KEY_NoSymbol;
+
+    /* search main table */
+    for (size_t i = 0; i < ARRAY_SIZE(keysymtab); i++)
+        if (keysymtab[i].ucs == ucs)
+            return keysymtab[i].keysym;
+
+    /* Use direct encoding if everything else fails */
+    return ucs | 0x01000000;
+}
+
 /*
  * Copyright © 2012 Intel Corporation
  *
--- a/test/keysym.c
+++ b/test/keysym.c
@@ -104,6 +104,29 @@ test_github_issue_42(void)
     setlocale(LC_CTYPE, "C");
 }
 
+static void
+get_keysym_name(xkb_keysym_t keysym, char *buffer, size_t size)
+{
+    int name_length = xkb_keysym_get_name(keysym, buffer, size);
+    if (name_length < 0) {
+        snprintf(buffer, size, "(unknown: 0x%lx)", (unsigned long)keysym);
+    }
+}
+
+static int
+test_utf32_to_keysym(uint32_t ucs, xkb_keysym_t expected)
+{
+    char expected_name[64];
+    char actual_name[64];
+    xkb_keysym_t actual = xkb_utf32_to_keysym(ucs);
+    get_keysym_name(expected, expected_name, 64);
+    get_keysym_name(actual, actual_name, 64);
+
+    fprintf(stderr, "Code point 0x%lx: expected keysym: %s, actual: %s\n\n",
+            (unsigned long)ucs, expected_name, actual_name);
+    return expected == actual;
+}
+
 int
 main(void)
 {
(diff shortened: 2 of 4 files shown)


src/keysym-utf.c
```diff
@@ -897,6 +897,30 @@ xkb_keysym_to_utf32(xkb_keysym_t keysym)
     return bin_search(keysymtab, ARRAY_SIZE(keysymtab) - 1, keysym);
 }
 
+XKB_EXPORT xkb_keysym_t
+xkb_utf32_to_keysym(uint32_t ucs)
```
bluetech: It'd be easier to verify if the code corresponds exactly to `xkb_keysym_to_utf32`, comments and all, with just the requisite adjustments.

I mean particularly the "patch encoding botch" and "special keysyms" hunks.

j-kubik: I have changed the numeric literals to `XKB_KEY_*` constants - hopefully it will make the intention more clear. Since the keypad keysyms have corresponding lower keysyms that produce the same character it will never be returned. Therefore it does not need any special handling like in `xkb_keysym_to_utf32`. The same applies to handling `XKB_KEY_KP_Space` - the `/* patch encoding botch */` is not needed.

bluetech: Thanks.

> Since the keypad keysyms have corresponding lower keysyms that produce the same character it will never be returned. Therefore it does not need any special handling like in xkb_keysym_to_utf32. The same applies to handling XKB_KEY_KP_Space - the /* patch encoding botch */ is not needed.

This will be helpful as a comment in the code, just to save time for anyone comparing the two.

test/keysym.c
```diff
@@ -104,6 +104,28 @@ test_github_issue_42(void)
     setlocale(LC_CTYPE, "C");
 }
 
+static void get_keysym_name(xkb_keysym_t keysym, char *buffer, size_t size)
+{
+    int nameLength = xkb_keysym_get_name(keysym, buffer, size);
+    if ( nameLength < 0) {
```
bluetech:
```suggestion
    if (nameLength < 0) {
```

test/keysym.c
```diff
@@ -104,6 +104,28 @@ test_github_issue_42(void)
...
+        snprintf(buffer, size, "(unknown: 0x%lx)", (unsigned long)keysym);
+        buffer[size-1]=0;
```
bluetech: `snprintf` truncates with a nul byte so this shouldn't be needed.

test/keysym.c
```diff
@@ -104,6 +104,28 @@ test_github_issue_42(void)
     setlocale(LC_CTYPE, "C");
 }
 
+static void get_keysym_name(xkb_keysym_t keysym, char *buffer, size_t size)
+{
+    int nameLength = xkb_keysym_get_name(keysym, buffer, size);
```
bluetech: We use snake case so `name_length`.

test/keysym.c
```diff
@@ -104,6 +104,28 @@ test_github_issue_42(void)
...
+    if ( nameLength < 0) {
+        snprintf(buffer, size, "(unknown: 0x%lx)", (unsigned long)keysym);
+        buffer[size-1]=0;
+    }
+}
+
+static int test_utf32_to_keysym(uint32_t ucs, xkb_keysym_t expected)
+{
+    char expectedName[64];
```
bluetech: `expected_name`, `actual_name`.

test/keysym.c
```diff
@@ -182,6 +204,33 @@ main(void)
...
+    assert(test_utf32_to_keysym(0x7f, XKB_KEY_Delete));
+
+    assert(test_utf32_to_keysym(' ', XKB_KEY_space));
+    assert(test_utf32_to_keysym(',', XKB_KEY_comma));
+    assert(test_utf32_to_keysym('.', XKB_KEY_period));
+    assert(test_utf32_to_keysym('=', XKB_KEY_equal));
+    assert(test_utf32_to_keysym('9', XKB_KEY_9));
+    assert(test_utf32_to_keysym('*', XKB_KEY_asterisk));
+    assert(test_utf32_to_keysym(0xd7, XKB_KEY_multiply));
+    assert(test_utf32_to_keysym('-', XKB_KEY_minus));
```
bluetech: Can you check one that would be directly encoded? Like some emoji or such.

src/keysym-utf.c
```diff
@@ -897,6 +897,29 @@ xkb_keysym_to_utf32(xkb_keysym_t keysym)
...
+        return ucs;
+
+    /* special keysyms */
+    if ((ucs >= (XKB_KEY_BackSpace & 0x7f) && ucs <= (XKB_KEY_Clear & 0x7f)) ||
+        ucs == (XKB_KEY_Return & 0x7f) || ucs == (XKB_KEY_Escape & 0x7f))
+        return ucs | 0xff00;
+    if (ucs == (XKB_KEY_Delete & 0x7f))
+        return XKB_KEY_Delete;
+
+    for (int i = 0; i < ARRAY_SIZE(keysymtab); ++i)
```
bluetech: Would add the comment `/* search main table */` in the line above.

test/keysym.c
j-kubik: I do not want to use direct encoding as character literals behavior in such case can be platform-dependent. A few of already added tests (i.e. hebrew, cyrilic and arabic) uses explicit numeric values instead to avoid such problems.

test/keysym.c
```diff
@@ -104,6 +104,27 @@ test_github_issue_42(void)
     setlocale(LC_CTYPE, "C");
 }
 
+static void get_keysym_name(xkb_keysym_t keysym, char *buffer, size_t size)
```
bluetech: The `static void` should be in its own line, also below.

test/keysym.c
```diff
@@ -182,6 +203,34 @@ main(void)
...
+    assert(test_utf32_to_keysym('y', XKB_KEY_y));
+    assert(test_utf32_to_keysym('u', XKB_KEY_u));
+    assert(test_utf32_to_keysym('m', XKB_KEY_m));
+    assert(test_utf32_to_keysym(0x43c, XKB_KEY_Cyrillic_em));
+    assert(test_utf32_to_keysym(0x443, XKB_KEY_Cyrillic_u));
+    assert(test_utf32_to_keysym('!', XKB_KEY_exclam));
+    assert(test_utf32_to_keysym(0xF8, XKB_KEY_oslash));
+    assert(test_utf32_to_keysym(0x5D0, XKB_KEY_hebrew_aleph));
+    assert(test_utf32_to_keysym(0x634, XKB_KEY_Arabic_sheen));
+    assert(test_utf32_to_keysym(0x1F609, 0x0101F609 )); // ;) emoji
```
bluetech:
```suggestion
    assert(test_utf32_to_keysym(0x1F609, 0x0101F609)); // ;) emoji
```
😉 😉 😉

j-kubik:
I have included the suggested documentation fragment. I have also went even a step further and defined the behavior for both codepoints `>=0x10ffff` and noncharacter unicode codepoints as they are guaranteed to never have a symbol assigned to it. The private use areas and the unassigned values are a fair game IMHO, as these can be well defined in particular use context or in the future.

xkbcommon/xkbcommon.h
```diff
@@ -493,6 +493,28 @@ xkb_keysym_to_utf8(xkb_keysym_t keysym, char *buffer, size_t size);
 uint32_t
 xkb_keysym_to_utf32(xkb_keysym_t keysym);
 
+/**
+ * Get the keysym corresponding to a Unicode/UTF-32 codepoint.
+ *
+ * @returns The keysym corresponding to the specified Unicode
+ * codepoint.
```
bluetech:
```suggestion
 * codepoint, or XKB_KEY_NoSymbol if there is none.
```

PR #138, opened 2020-03-24
 Use flag instead of hardcoded value in evdev examples

NilsIrl:
This is what is done in the [guide](https://xkbcommon.org/doc/current/md_doc_quick-guide.html)

### DIFF
--- a/test/interactive-evdev.c
+++ b/test/interactive-evdev.c
@@ -453,7 +453,8 @@ main(int argc, char *argv[])
             goto err_ctx;
         }
         keymap = xkb_keymap_new_from_file(ctx, file,
-                                          XKB_KEYMAP_FORMAT_TEXT_V1, 0);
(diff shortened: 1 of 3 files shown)


NilsIrl:
> Could be squashed into one commit

Sorry, I couldn't be asked to clone the repo.

But now that i've cloned it, there's the file test/data/keymaps/quartz.xkb that's **always** modified. Nothing will remove it: https://asciinema.org/a/jPRPXoBB8ynITjabdUKMSLLlX

bluetech:
Thanks @NilsIrl. I squashed when I merged.

> But now that i've cloned it, there's the file test/data/keymaps/quartz.xkb that's always modified

That file has `\r\n` line endings. IIRC this is intentional, to verify parsing such line endings works. And your git is configured to convert `\r\n` to `\n`. There are some git configs which control this.

PR #139, opened 2020-03-25
Fix inconsistency and respect standard libxkbcommon

NilsIrl:
https://gitlab.freedesktop.org/wayland/wayland/-/blob/master/protocol/wayland.xml#L2182

We can map `size - 1` since we don't pass to `xkb_keymap_new_from_buffer`, `size - 1`.

### DIFF
--- a/test/interactive-wayland.c
+++ b/test/interactive-wayland.c
@@ -331,7 +331,7 @@ kbd_keymap(void *data, struct wl_keyboard *wl_kbd, uint32_t format,
     struct interactive_seat *seat = data;
     void *buf;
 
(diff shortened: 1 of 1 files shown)


fooishbar:
As mappings are rounded to page size anyway, I don't think this makes much difference in the end?

bluetech:
The `MAP_PRIVATE` part is https://github.com/wayland-project/wayland/commit/905c0a341ddf0a885811d19e2b79c65a3f1d210c, so I think makes sense. But should we require version 7, or does it not matter?

The `size - 1` change I think is better avoided -- let's map the size we're given, even if we only read `size - 1` of it.

bluetech:
Going to close this, but let us know if you still want to pursue this.

Issue #140, opened 2020-04-04
test/how-to-type.c nitpick


AliKet:
Running this command:
`./how-to-type -l "fr,us" 0x61  | column -ts $'\t'`

**Expected output:**
```
keysym: a (0x61)
KEYCODE           KEY NAME  LAYOUT#  LAYOUT NAME   LEVEL#  MODIFIERS
24                AD01      1        French        1       [ ]
38                AC01      2        English (US)  1       [ ]
```
**But it actually display:**
```
keysym: a (0x61)
KEYCODE           KEY NAME  LAYOUT#  LAYOUT NAME  LEVEL#  MODIFIERS
24                AD01      1        French       1       [ ]
```
This patch corrects this (at least for me):
```
diff --git a/test/how-to-type.c b/test/how-to-type.c
index df9a2be..2516bd2 100644
--- a/test/how-to-type.c
+++ b/test/how-to-type.c
@@ -171,6 +171,13 @@ main(int argc, char *argv[])
                 num_masks = xkb_keymap_key_get_mods_for_level(
                     keymap, keycode, layout, level, masks, ARRAY_SIZE(masks)
                 );
+                if (num_masks == 0)
+                {
+                    printf("%u\t%s\t%u\t%s\t%u\t[ ]\n",
+                           keycode, key_name, layout + 1, layout_name, level + 1);
+                    continue;
+                }
+
                 for (size_t i = 0; i < num_masks; i++) {
                     xkb_mod_mask_t mask = masks[i];
 
```

bluetech:
Thanks for trying this!

This is a good point which I didn't consider before. The output you give is definitely unexpected, unless we explicitly specify that only explicit matches are shown, which we probably don't want to do.

I actually (currently) think it would be better to fix this in `xkb_keymap_key_get_mods_for_level()` itself rather than in `how-to-type`. See PR #141 for a proposed fix.

PR #141, opened 2020-04-04
keymap: don't forget about fallback mappings in xkb_keymap_key_get_mods_for_level()

bluetech:
Fixes https://github.com/xkbcommon/libxkbcommon/issues/140.

If the active set of modifiers doesn't match any explicit entry of the
key type, the resulting level is 0 (i.e. Level 1). Some key types don't
explicitly map Level 1, taking advantage of this fallback.

Previously, xkb_keymap_key_get_mods_for_level didn't consider this, and
only reported masks for explicit mappings. But this causes some glaring
omissions, like matching "a" in the "us" keymap returning not results.

Since every mask which isn't explicitly mapped falls back to 0, we can't
return the all. Almost always the best choice for this is the empty
mask, so return that, when applicable.

### DIFF
--- a/src/keymap.c
+++ b/src/keymap.c
@@ -433,12 +433,39 @@ xkb_keymap_key_get_mods_for_level(struct xkb_keymap *keymap,
     const struct xkb_key_type *type = key->groups[layout].type;
 
     size_t count = 0;
-    for (unsigned i = 0; i < type->num_entries && count < masks_size; i++)
+
+    /*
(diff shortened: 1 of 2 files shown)


bluetech:
@j-kubik, would you care to take a look at the commit here? It proposes a change to `xkb_keymap_key_get_mods_for_level()`.

You actually wrote a test which verified the old behavior, which I didn't pay much mind to before.

PR #142, opened 2020-04-10
Fix building X11 tests on PE targets

jon-turney:
Link libxkbcommon_x11_internal with libxkbcommon_test_internal, rather
than libxkbcommon.

This avoids some tests linking with both libxkbcommon_test_internal and
libxkbcommon, which causes duplicate symbol problems on PE targets (e.g.
Cygwin) (as all the symbols from libxkbcommon are pulled in at link
time, which clash with libxkbcommon_test_internal)


bluetech:
Thanks!

PR #143, opened 2020-04-18
Switch to github actions

bluetech:
The Azure CI is currently broken so seemed like a good opportunity to switch. This way a microsoft account is not needed.


fooishbar:
Very slightly belated LGTM.

bluetech:
@fooishbar Thanks for taking a look!

PR #150, opened 2020-06-04
Add documentation on how to configure user-specific XKB RMLVO

whot:


doc/rules-format.md
```diff
@@ -58,12 +59,14 @@ Kccgst       ::= "keycodes" | "symbols" | "types" | "compat" | "geometry"
...
 KccgstValue  ::= <ident>
+```
 
 Notes:
 
 - Include processes the rules in the file path specified in the ident,
-  in order. %-expanasion is performed, as follows:
+  in order. %-expansion is performed, as follows:
 
+```
```
bluetech, line 69: This would maybe be better as a (sub)list? (With just the `%H` parts in ticks)

doc/rules-format.md
```diff
@@ -80,6 +84,7 @@ Notes:
 - If a Rule is matched, %-expansion is performed on the KccgstValue,
   as follows:
 
+```
```
bluetech, line 87: Here too.

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+This page describes how to add a custom layout or option so that it will be
+parsed by libxkbcommon.
+
+**The below requires libxkbcommon as keymap compiler and does not work in X**.
+
+## Data locations
+
+libxkbcommon searches the following paths for XKB configuration files:
+- `$XDG_CONFIG_HOME/xkb/`
+- `$HOME/.config/xkb/`
```
bluetech, line 12:
```suggestion
- `$XDG_CONFIG_HOME/xkb/`, or `$HOME/.config/xkb/` if the `XDG_CONFIG_HOME` environment variable is not defined
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+- `$HOME/.xkb/`
+- `$XKB_CONFIG_ROOT`
+- `<prefix>/X11/xkb/` (on most distributions this is `/usr/share/X11/xkb`)
```
bluetech, line 15:
```suggestion
- `<datadir>/X11/xkb/` (path defined by the xkeyboard-config package, on most distributions this is `/usr/share/X11/xkb`)
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+
+A keymap created with `xkb_keymap_new_from_names()` will look up those paths in
+order until the required data is found.
+
+**Note: Where libxkbcommon runs in a privileged context, only the prefix path is
```
bluetech, line 20:
```suggestion
**Note: Where libxkbcommon runs in a privileged context, only the system (datadir) path is
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+A detailed explanation of how rules files convert RMLVO to KcCGST is out of
+scope for this document. See [the rules file](md_doc_rules-format.html) page
+instead.
+
+
+## Adding a layout
+
+Adding a layout requires that the user adds **symbols** in the correct location.
```
bluetech, line 61:
```suggestion
Adding a layout requires that the user add **symbols** in the correct location.
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+
+The default rules files (usually `evdev`) have a catch-all to map a layout "foo"
+variant "bar" into the "bar" section in the file `$xkb_base_dir/symbols/foo`.
```
bluetech, line 64:
```suggestion
The default rules files (usually `evdev`) have a catch-all to map a layout, say "foo",
and a variant, say "bar", into the "bar" section in the file `symbols/foo`.
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+This is sufficient to define a new keyboard layout. The example below defines
+the keyboard layout "banana" with an optional variant "orange"
+
+```
+$ cat $XDG_CONFIG_HOME/xkb/symbols/banana
+// Like a US layout but swap the top row's so numbers are on Shift
```
bluetech, line 70:
```suggestion
// Like a US layout but swap the top row so numbers are on Shift
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+    include "eurosign(5)"
+};
+```
+
+The `default` section is loaded when no variant is given. The first example
+sections uses ``include`` to populate with a symbols list defined elsewhere
+(here: section `basic` from the file `symbols/us`, aka. the default US keyboard
+layout) and overrides parts of these
+symbols. The effect of this section is to swap the numbers and symbols in the
+top-most row (compared to the US layout) but otherwise uses the US layout.
```
bluetech, line 104:
```suggestion
top-most row (compared to the US layout) but otherwise use the US layout.
```

doc/user-configuration.md
```diff
@@ -0,0 +1,154 @@
...
+
+```
+$ cat $XDG_CONFIG_HOME/xkb/symbols/banana
+// Like a US layout but swap the top row's so numbers are on Shift
+default partial alphanumeric_keys
+xkb_symbols "basic" {
+    include "us(basic)"
+    name[Group1]= "Banana (US)";
+
+    key <AE01> {	[	exclam, 	1]	};
```
bluetech, line 76: I think it will be better to replace the tabs with spaces, their rendering is rather inconsistent.

doc/rules-format.md
whot: I had that first but it doesn't work, the % gets swallowed if the backticks and the `%H` is on the same line. This is a doxygen markdown issue, not visible in gitlab markdown. This only applies to letters, not to `%%` or `%-` in the sentence above. Backslash escape works (`\%H`) but that is confusing in the .md source then.

I'd rather go with bad rendering than ambiguity here, at least the former doesn't create bug reports :)

doc/user-configuration.md
whot: that doesn't seem right. "the user adds" or "users add". Given that it's user-specific, I'd prefer to stick to singular here.

doc/rules-format.md
whot: see above

whot:
All updated as requested minus the comments above, thanks.

PR #154, opened 2020-06-22
Remove the gitignore file

whot:
This contained three groups entries:
- autotools stuff which no longer needed
- entries to ignore built artifacts, no longer needed with meson as everyting is
  in the builddir
- special files like cscope.out, etc. These should be handled by a user in their
  global gitignore file since they're fairly specific to a user's setup.


.gitignore
```diff
@@ -1,87 +0,0 @@
...
-#
-
-.dirstamp
-cscope.out
-test-suite.log
-test-driver
-build-aux/
-/xdg-shell-unstable-v6-client-protocol.h
-/xdg-shell-unstable-v6-protocol.c
-build/
```
bluetech, line 87: I would selfishly want to see `build/` stick around, but I can also stick it in `.git/info/exclude`.

whot: why not add it to `XDG_CONFIG_HOME/git/ignore`? are there any projects where `build` matters? And if so, train yourself to use `builddir` instead (though admittedly I failed at bit that so far :)

Issue #151, opened 2020-06-14
Combine demos to one binary


bluetech:
I think it will be nice to combine all of the demo programs we have to one single `xkbcommon` demo binary. It will be able to:

- Show the version it was compiled with, include paths, build-time constants, enabled features etc.
- Interactive evdev demo (on Linux)
- Interactive X11 demo (when xkbcommon-x11 is enabled)
- Interactive Wayland demo (when wayland is enabled)
- Print RMLVO to KCCGST conversion
- Print keymap from RMLVO
- Print (full compiled) keymap from file
- "How to type" functionality
- Benchmarks (compose, key proc, rules, rulescomp)

It should use a `./xkbcommon <subcommand>` UI, with consistent option processing.

All of this already exists but is spread around a dozen different binaries.

whot:
would probably be useful to move this to `tools/` as well to keep the unit-tests separate from tools

bluetech:
@whot Do you think it's a good idea? The `xinput` tool was an inspiration for this :)

whot:
a common prefix is a good idea, makes it easier to find the tools with `./builddir/xkbcommon-<tab><tab>` when debugging. A single tool depends on the use-case you envision. If you expect users to execute those tools (quick check of `meson.build` shows we don't install them) then having a single tool is primarily interesting because of the namespace pollution otherwise.

for `xinput` I have a single binary, for libinput I picked the git-style approach of a parent tool that `execve`s into the subtools in `/usr/libexec`. Biggest advantage here is that the tools can be in e.g. python as well. In the builddir, they're still all `libinput-foo-bar`. I think that's the best approach, esp. if you don't quite know what the future will bring.

If you don't really expect those tools to ever be used outside the git build dir, then namespacing is enough, the effort to have a single tool just gets in the way of the debugger.

whot:
I had a look at this today and started thinking - what tools do we actually need and what for? Here are a few thoughts.

We have four groups of tools atm:
- `interactive-evdev`, `interactive-wayland`, `interactive-x11`. Those *could* be merged into one tool with a `--backend=evdev` option or somesuch, together with making the output fully consistent across all. 
- `registry-list`: I think the output format needs to be improved to make it end-user compatible
- `how-to-type`: standalone tool
- `print-compiled-keymap`, `rmlvo-to-kccgst`, `rmlvo-to-keymap`: these are all more or less the same tool anyway and could easily be combined into one tool. This is #159 now.

[...]

PR #155, opened 2020-06-22
Move tools to a tools/ directory

whot:


bluetech:
Thanks! LGTM, except that the `interactive-*` set probably also should be in tools/. (If you move them, make sure to `git grep test/interactive` as they are referred to by path name in a few places in the docs).

whot:
@bluetech that's the lot now, I think. includes a renaming to prefix the tools with `xkbcommon-` (part of #151).

tools/print-compiled-keymap.c
```diff
@@ -53,32 +57,64 @@ main(int argc, char *argv[])
...
+    ctx = xkb_context_new(XKB_CONTEXT_NO_FLAGS);
     if (!ctx) {
         fprintf(stderr, "Couldn't create xkb context\n");
-        goto err_out;
+        goto out;
     }
 
-    keymap = test_compile_file(ctx, keymap_path);
+    if (streq(keymap_path, "-")) {
+        FILE *tmp;
```
bluetech, line 67: Heh stdio is pretty painful. I do think we need to read in a loop though, assuming the the stdin input is at most 4096 bytes is not great.

whot: huh, I thought I had done that. Sorry, will fix.

whot:
>  I think it'd be a bit nicer to duplicate (or open code) whatever functions are needed from there instead of including.

IMO it's best to split up so that `utils.h` is a generic utilities set like the [libinput `utils-foo.(c|h)` files](https://gitlab.freedesktop.org/libinput/libinput/-/tree/master/src). Then we can use it from anywhere, and the xkbcommon-specific utilities that require knowledge of headers can go into a separate file. but tbh, after a quick look at `utils.h` I don't see a function that needs to be in a more specific header.

whot:
> rmlvo-to-kccgst.c includes xkbcomp/xkbcomp-priv.h and xkbcomp/rules.h. This perhaps indicates some missing public API, though we got by without exposing kccgst at all, so maybe we should just ignore this one, and keep the private linking.

had a look at this and yeah, it would require exposing KcCGST as public API, which is something to avoid. So I'd say let's keep it at this.

the `stdin` handling is fixed (I think, it's getting late here :) in the last set of commits, only the top commit changed.

whot:
Thanks! merging

PR #159, opened 2020-07-07
Consolidate the keymap compiliation tools

whot:
One step towards #151, this combines the separate keymap compilation tools into one - still named `xkbcommon-rmlvo-to-keymap` because I haven't yet thought of a better name.


tools/rmlvo-to-keymap.c
```diff
@@ -33,22 +33,18 @@
...
-
 static void
 usage(char **argv)
 {
-    printf("Usage: %s [--print] [--rules <rules>] [--layout <layout>] [--variant <variant>] [--options <option>]\n",
+    printf("Usage: %s [--rules <rules>] [--layout <layout>] [--variant <variant>] [--options <option>]\n",
            argv[0]);
-    printf("This tool tests the compilation from RMLVO to a keymap.\n");
-    printf("--print  print the resulting keymap\n");
+    printf("Compile the RMLVO to a keymap.\n");
```
bluetech:
```suggestion
    printf("Compile an RMLVO to a keymap and print it.\n");
```

tools/rmlvo-to-keymap.c
```diff
@@ -64,6 +65,9 @@ usage(char **argv)
            "    Enable verbose debugging output\n"
            " --print-kccgst\n"
            "    Print the KcCGST names instead of the full keymap\n"
+           " --from-kccgst\n"
```
bluetech, line 68: The input is not necessarily KcCGST, it can be any xkb file (and I occasionally use it to get the compiled form of some xkb file). So maybe `--from-xkb`?

I understand it will make it less obvious that you can pipe ``--print-kccgst`` into ``--from-kccgst``. So I suggest changing ``--print-kccgst`` to only ``--kccgst`` with this help text:

> Print a keymap which only includes the KcCGST instead of the fully processed keymap

whot: just for clarification - what do you mean by "any xkb file"? AFAIU and with a short trial-and-error it needs to be a `xkb_keymap` which can be either the KcCGST only or the full keymap. Maybe `--from-xkb-keymap` is a better option then? 
I've changed it to `--from-xkb` for now though.

tools/rmlvo-to-keymap.c
whot: fixed, thanks

tools/rmlvo-to-keymap.c
```diff
@@ -133,6 +134,7 @@ main(int argc, char **argv)
         xkb_context_set_log_verbosity(ctx, 10);
     }
 
+    xkb_context_sanitize_rule_names(ctx, &names);
```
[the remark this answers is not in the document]
whot: squashed, thanks

tools/rmlvo-to-keymap.c
```diff
@@ -34,12 +34,40 @@
 #include "xkbcomp/xkbcomp-priv.h"
 #include "xkbcommon/xkbcommon.h"
 
+#ifndef DEFAULT_XKB_VARIANT
```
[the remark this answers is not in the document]
whot: done! I used `<none>` because "unset" has the ambiguity of "not yet set, filled in later" whereas "none" means none. Sort-of, anyway :)

whot:
Thanks for the review, all comments addressed as requested. I added another  option on top with `--rmlvo` to print the full RMLVO configuration with the defaults filled in:
```
$ xkbcommon-rmlvo-to-keymap --rmlvo --model pc104
rules: "evdev"
model: "pc104"
layout: "us"
variant: ""
options: ""
```

whot:
Merging this for now so I can actually build on top of it without going crazy :)

tools/rmlvo-to-keymap.c
bluetech: By analogy to RMLVO I usually take KcCGST to mean just the 5 values (`keycodes = evdev+aliases(qwerty)` etc).

`--from-xkb-keymap` is also good though annoying to type.

whot: that's what bash/zsh completion is for ;)

Issue #153, opened 2020-06-18
Unnoticed parsing error in xkeyboard-config


whot:
Haven't fully triaged this yet but: xkeyboard-config 2.30 shipped with a leftover merge conflict marker (`======`) in the `symbols/in` file. This was not detected by the `rmlvo-to-keymap` tool that we run as part of the CI in xkeyboard-config.

Manually verified that it succeeds:
```
$ rmlvo-to-keymap --print --rules evdev --model pc105 --layout in --variant ben_probhat
```
but `xkbcomp` fails:
```
$ setxkbmap -print -layout in -variant ben_probhat | xkbcomp -xkb - -
[...]
syntax error: line 2329 of in
last scanned symbol is: question
Error:            Error interpreting include file "in"
                  Exiting
                  Abandoning symbols file "(null)"
```

This ~~could be a bug in the `rmlvo-to-keymap` tool or a~~ is a bug in the parser where it's too lenient for these cases, not sure.

whot:
Ok, the `rmlvo-to-keymap` tool only calls one function that matters here: `xkb_keymap_new_from_names()`. That one returns a valid keymap so the bug is definitely in the parser.

Whether it's intentionally more lenient or not I can't say. @bluetech, @fooishbar ?

For reference, the syntax error was: https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/commit/c8d68304c826567f27f5a913dd38f6e045b81483

bluetech:
xkbcommon has an optimization where it stops parsing once it reaches the map it needs (map == `xkb_symbols` section in this case). If I'm reading the commit you linked correctly, then only `-layout in -variant modi-kagapa` would cause it to reach the syntax error. Maybe this particular variant isn't triggered by `xkeyboard-config-test.py`?

whot:
oh, good point. Indeed, `modi-kagapa` wasn't used [1] because it's in `evdev.extras.xml` and the test tool doesn't check those. I've filed #156 for this now, thanks!


[1] see pipeline artifacts [here](https://gitlab.freedesktop.org/pmenzel/xkeyboard-config/-/jobs/3010151) but they'll be gone in 20h according to gitlab.

PR #162, opened 2020-07-10
Allow for merged KcCGST files and a system path

### LINKED ISSUE #40: Pluggable layout support
The qwerty-lafayette is a new qwerty-bqsed french layout that is not yet reqdy to be included in the xkb maps. Still, it would be desirqble for users to be able to install this layout (see issue fabi1cazenave/qwerty-lafayette#15 in French).

The diff required to install this layout system-wide is: https://gist.github.com/fabi1cazenave/2171d69d3d8d6e298225

The problem with this diff is that it is not possible to create a package from it because it requires patching the rules files. It would be desirable if libxkbcommon could read additional rule files in directories.

For example, `/usr/share/X11/xkb/rules/evdev.d/lafayette.lst` and `/usr/share/X11/xkb/rules/evdev.d/lafayette.xml` could be read as part of `/usr/share/X11/xkb/rules/evdev.lst` and `/usr/share/X11/xkb/rules/evdev.xml` respectively.

It would be even better if libxkbcommon could read files from `~/.xkb/rules/evdev.d/*.{lst,xml}`

whot:
Three additions in this branch, it sits on top of #160 otherwise, starting at commit
"Add asprintf_safe helper function"

## `/etc/xkb` as lookup path

This adds `/etc/xkb` as a lookup path for system-wide keyboard configurations. Fixes #40, see my [long comment](https://github.com/xkbcommon/libxkbcommon/issues/40#issuecomment-656473348) on why just a single path and not a `conf.d` style directory.

So our lookup path sequence becomes - in that order:
- `XDG_CONFIG_HOME/xkb`
- `$HOME/.xkb`
- `/etc/xkb`
- `/usr/share/X11/xkb`

Which gives us roughly the same behaviour as most other tools that have config files.

## `xkbcli scaffold` tool

All directories still need to adhere to the same folder structure since the folder names and file names are part of the API. To make it simpler to get started with a custom XKB layout, there's a new tool called `xkbcli scaffold`. So a user would run, e.g.
```
$ xkbcli scaffold --user --layout "us(myvariant)" --option "custom:foo"
```
And that would create the required bits in `$XDG_CONFIG_HOME/xkb`. All they need now is learn how XKB files work and fill them in :) Passing `--system` does the same in the new `/etc/xkb` folder. Otherwise it's a fairly simple tool and it only writes to files that don't yet exist so it's safe to call twice (though ineffective).

## Merge-include behaviour for KcCGST files

And since it's currently not possible to have `us(myvariant)` (see [the same comment](https://github.com/xkbcommon/libxkbcommon/issues/40#issuecomment-656473348)), change the KcCGST include behavior to run through all include paths.
So an include statement like `include "us(basic)"` will be attempted in in all `symbols/us` files in the include paths until the first one that works is found. It's thus possible to have a custom variant extending a system layout like this:

```
$ cat $XDG_CONFIG_HOME/xkb/symbols/us
partial alphanumeric_keys modifier_keys
xkb_symbols "myvariant" {
    name[Group1]= "myvariant (us)";
    include "us(basic)"
    key <CAPS> { [ Escape ] };
};
```
And this does the right thing.

cc @fooishbar and @svu because while I think this is save to do, I'm not sure if there are hidden side effects I haven't though of.


bluetech:
I'll review this when I get the next chance. In the mean time a rebase would be good.

Is this something you think 0.11.0 should include, or is it fine to release without?

whot:
Rebased on top of #173 now.

I renamed the tool to `xkbcli-scaffold-new-layout` but marked it `install: false` until we know it's useful. Rest is the same but has seen a few improvements in the code handling. Marking it as ready

> Is this something you think 0.11.0 should include, or is it fine to release without?

I think we should include it. 0.10.0 added the user-specific files but they were only usable in a narrow context. With libxkbregistry and this the handling of custom files is complete (I think) and we have a good replacement for xmodmap (and of course custom/experimental keyboard layouts). If we get that in 0.11 in one go, life is a bit easier for everyone including writing documentation.

At least on my TODO list the only thing left now is to fix the xkeyboard-config format to be less awkward, but that's not going to happen anytime soon...

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+
+
+def create_directory_structure(basedir):
+    basedir.mkdir(exist_ok=True)
+    # Note: we skip geometry
+    for d in ['compat', 'keycodes', 'rules', 'symbols', 'types']:
+        (basedir / d).mkdir(exist_ok=True)
+
+
+def create_rules_template(base_path, ruleset, layout, option):
```
bluetech, line 69: I suggest `basedir` instead of `base_path` for consistency with how it's named in the rest of the file.

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+    rules = Path(base_path) / 'rules' / ruleset
```
bluetech, line 70: Can drop the `Path()` here since it's already a `Path`

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+    if rules.exists():
+        logger.warning(f'Rules file {rules} already exists, skipping')
+        return
+
+    with open(rules, 'w') as fd:
```
bluetech, line 75: `fd` for me is file-description (a number), so would just call it `f`

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+            ''')
+        fd.write(header)
+
+        if option:
+            group, section = option
+            option_template = dedent(f'''\
+            // This section maps XKB option "{group}:{section}" to the '{section}' section in the
+            // 'symbols/{group}' file.
+            //
+            ! option = symbols
```
bluetech, line 90: This assumed that the option is a symbols option, but options for compat/types/keycodes are possible as well. I agree `symbols` is probably the best default, but perhaps we can mention this in the generated comment above.

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+              {group}:{section} = +{group}({section})
+
+            ''')
+            fd.write(option_template)
+
+        footer = dedent(f'''\
+            // Include the system '{ruleset}' file
```
bluetech, line 97: I wonder if `--user` is used we should also somehow hook up `/etc/xkb` here? Might be a little difficult as it might (probably) not exist...

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+                {default}partial alphanumeric_keys modifier_keys
+                xkb_symbols "{variant}" {{
+                    name[Group1]= "{variant} ({layout})";
+
+                    {include}
+
+                    // Example:
+                    // key <CAPS> {{ [ Escape ] }};
+                }};
+                '''))
```
bluetech, line 151: Should close `layout_fd` here (same for `options_fd` below)

whot:
> The strategy you've used for this is to reduce the granularity from the file level to the map level. This is good. The first file which
> has a matching map is used. Another alternative is to merge the maps of all files together. XKB has a bunch of merging 
> machinery for this, with 3 different merge keywords augment, override, replace. Have you considered using that somehow 
> instead  of first-wins? I haven't considered it deeply myself, but wanted to mention it.

My understanding is that those keywords apply at the map level, not at the file level. Technically, the "layout is the filename" is an implementation detail of xkeyboard-config and our parsers, there is nothing that requires this explicitly. We could switch to sqlite with the layout as index and it'd be correct too. So `augment` is really just `include-with-custom-merge-behaviour` but like the pure `include` it doesn't explicitly specify *where* the data should come from.

Merging the files together thus wouldn't change much I think - if anything it makes our behaviour parser-dependent. If you have a (virtual combined) file with two identically-named sections, is the parser going to pick the first one? or the last one? or will it throw an error in that case? The current approach is a bit more expressive, you know where things should be picked up from.

We could possibly design some sort of custom-tailored include instruction set but I think it'd be overkill for what we need.

bluetech:
> My understanding is that those keywords apply at the map level, not at the file level.

They apply at a bunch of places -- search for `MergeMode` and `OptMergeMode` in `src/xkbcomp/parser.y`.

> If you have a (virtual combined) file with two identically-named sections, is the parser going to pick the first one? or the last one? or will it throw an error in that case? The current approach is a bit more expressive, you know where things should be picked up from.

I'm not sure I understand what is parser-dependent. xkbcomp only ever looks at the XKB root, and the only other parser is us :) And this PR is changing the behavior.

Also not sure I follow about the expressiveness. The first-wins strategy is `replace` (in reverse?), so it can't be more expressive than something that offers more choices.

To be more concrete, let's say I want to modify something about the `us` layout. With the first-wins/replace strategy, the procedure is: 

- Add a variant, `us(myvariant)` to `us` file
- Make `myvariant` include `us(basic)`
- Add the desired modifications to `myvariant`
- Add `myvariant` to the rules
- Configure my system to use `myvariant`

With a map-merging augment strategy it will be

- Add a `us(basic)` map to `us` file
- Add the desired modifications to it

The trouble with this are:

1. It changes the meaning of the `us(basic)` layout itself - maybe we do want to force a different name, but I think it's fine.
2. I would require changing the parsing order such that system files are processed before the user files (so the user files can augment them). I think -- maybe not, I need to remind myself of the code here...

---

To be clear, I'm not advocating for this change, I'm pretty much offloading all of the thinking on this to you :grin: But I want to make sure we discuss all options. I'm also having a deja vu that we discussed this before already but maybe it's just a glitch in the Matrix.

whot:
> Add myvariant to the rules

just ftr, this is only needed for options, not for variants due to how the rules catchall commands work. but yeah, that's just a lucky accident.

My argument of the parser dependency was: if you merge all files into a virtual combined file, you then need to also specify how the parser will treat sections that are doubly defined and ensure that all parsers (well, "both") parse correctly. By making it a lookup-path dependency, this behaviour is implied. It's a slightly wobbly argument but it's not a fully theoretical case -  we recently had a bug where xkbcomp and libxkbcommon's parsing behaviour differed, see #153.

> It changes the meaning of the us(basic) layout itself - maybe we do want to force a different name, but I think it's fine.

I disagree here, primarily for *our* benefit. Being able to modify the `us` layout and have it still be named `us` makes bug reports less reliable, let alone the various forums where custom modifications will spread like wildfire and make it more confusing for everyone. It's also a (subjective) interpretation of what layouts are and how people want to modify them. The naming of "it's a US layout but it's the foo variant" is useful and explicit enough, "it's a modified US layout" is harder. This isn't a technical argument, it's mostly one of long-term convenience for us as maintainers. And in terms of use-case: I'm aiming for the top-down configuration only. We assume the system files are constant and custom configuration can modify them or hide them altogether, but there's no bidirectional flow - you can't (easily) create a custom variant that includes a system file that then includes your custom other variant. It's probably technically possible but it's not really a use-case we should be worrying about. Where something like that is needed, it's cheaper to just modify the whole layout and effectively copy the system one into the custom configuration.

Ok, so with all that said, let's think about the technical implementation. I simply don't know *where* we'd put those instructions. Our components are spread across up to three files, `$XDG_CONFIG_HOME`, `/etc/xkb` and `/usr/share/X11/xkb`. The lookup behaviour is merely defined in the lookup paths [1] but otherwise it's not specified and it has no effect on the actual behaviour. Right now, `augment` should work just as well as `include` (not tested though).

IOW, right now there's nothing explicit about where `us(basic)` will come from. Picking different merge behaviours would require a place where to set those and right now, we don't really have one. I guess we *could* expand to allow things like `include %S/us(basic)` and so on but I'm not sure we really want to.

Maybe I'm missing something obvious here but right now I just don't know where we'd even configure different merge behaviours across files (and whether this would really have an effect).

[1] lookup paths are controlled by the parser, not the config. That's a potential issue but, well, uhm... don't know what to do about that either.

doc/user-configuration.md
```diff
@@ -11,6 +11,8 @@ libxkbcommon searches the following paths for XKB configuration files:
 - `$XDG_CONFIG_HOME/xkb/`, or `$HOME/.config/xkb/` if the `$XDG_CONFIG_HOME`
   environment variable is not defined
 - `$HOME/.xkb/`
+- `$XKB_CONFIG_EXTRA_PATH` if set, otherswise `<sysconfdir>/xkb` (on most
```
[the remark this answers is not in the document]
whot: we sort-of need it for the tests. By overriding that variable we can test default include path handling without having to actually rely on correct configuration in `/etc/`

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+        logger.setLevel(logging.DEBUG)
+
+    if args.system:
+        basedir = Path(default_value('extrapath'))
+    else:
+        xdgdir = os.getenv('XDG_CONFIG_HOME')
+        if not xdgdir:
+            home = os.getenv('HOME')
+            if not home:
+                logger.error('Unable to resolve base directory from $XDG_CONFIG_HOME or $HOME')
```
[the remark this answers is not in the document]
whot: Default output is this one which I thought was good enough for a niche tool like this:

> ERROR:xkbcli:Unable to resolve base directory from $XDG_CONFIG_HOME or $HOME

tools/xkbcli-scaffold-new-layout.py
whot: one of the reasons I use `fd` here is that `file` is reserved and since 3.6 `f` is used for format strings. Like here the code would be:
```python
with open(rules, 'w') as f:
     header = dedent(f''' \
     ...
```
Which I always find a bit jarring. So `fd` is the compromise here, even though in C it's a number. This is not a hill I'm willing to die on though so I can rename it if you prefer.

tools/xkbcli-scaffold-new-layout.py
whot: `layout_fd` and `options_fd` may point to the same file so we can't close them until the end of the file. I've added that now, thanks.

tools/xkbcli-scaffold-new-layout.py
whot: I got curious and ran this command in `xkeyboard-config`:

```
$ git log --format=oneline --since 2015 keycodes compat types | wc -l
18
```
That's for an almost 5-year span. Quick glance shows only 11 of those are non-janitor commits and those are largely adding new keycodes in the form of, e.g.

```
+       <I372> = 372;   // #define KEY_FAVORITES           364
```
so yes, it's just a comment but at this point I'm wondering whether that would point users down a path they don't actually need to travel. The ones that care about compat and types will probably figure it out independently.

Completely unrelated: I do wonder if we should built automatic mapping for evdev keycodes into the parser so the above addition isn't needed.

tools/xkbcli-scaffold-new-layout.py
whot: yeah, and we'd need to run as root. And it's a niche case anyway, so overall I think it's not worth the bother. Or at least not until we see that users actually need this, right now it's all a bit of stab in the dark.

test/tool-option-parsing.py
```diff
@@ -302,5 +302,7 @@ def test_interactive_wayland(xkbcli_interactive_wayland):
 if __name__ == '__main__':
     with tempfile.TemporaryDirectory() as tmpdir:
         os.environ['XDG_CONFIG_HOME'] = tmpdir
+        # This needs to be separated if we do specific extra path testing
+        os.environ['XDG_EXTRA_PATH'] = tmpdir
```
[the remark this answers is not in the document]
whot: fixed as requested, thanks

tools/xkbcli-scaffold-new-layout.py
```diff
@@ -0,0 +1,323 @@
...
+    group.add_argument('--system', action='store_true', default=False,
+                       help=f'Create scaffolding in {default_value("extrapath")}')
+    group.add_argument('--user', action='store_true', default=False,
+                       help='Create scaffolding in $XDG_CONFIG_HOME/xkb')
+    parser.add_argument('--rules', type=str, default=f'{default_value("rules")}',
+                        help=f'Ruleset name (default: "{default_value("rules")}")')
+    parser.add_argument('--layout', type=str, default='us(myvariant)',
+                        help='Add scaffolding for a new layout or variant (default: "us(myvariant)")')
+    parser.add_argument('--option', type=str, default='custom:foo',
+                        help='Add scaffolding for a new option (default: "custom:foo")')
```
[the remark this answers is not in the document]
whot: fixed as requested, thanks

tools/xkbcli-scaffold-new-layout.py
whot: fixed as requested, thanks

tools/xkbcli-scaffold-new-layout.py
whot: renamed to `xmlfile` and `rulesfile` in the two instances where `fd` was used previously

tools/xkbcli-scaffold-new-layout.py
whot: I've added a short blurb to the documentation itself for this.

whot:
Force-pushed with the requested fixed, the ones I've not marked as resolved haven't been changed - please see my comments there.

bluetech:
Thanks for tackling this @whot! I think we're about ready for release. I'll aim for this weekend. And let's call it 1.0.0 :grin:

bam80:
> So our lookup path sequence becomes - in that order:
> 
>    * `XDG_CONFIG_HOME/xkb`
> 
>    * `$HOME/.xkb`

I think `$HOME/.xkb` is not necessary any more,
the random dotstyle-configs/dirs in $HOME are deprecated nowadays, being fully replaced by
XDG_CONFIG_HOME path which renders to something like `${XDG_CONFIG_HOME:-$HOME/.config}`:

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 

https://specifications.freedesktop.org/basedir-spec/latest/#variables

So I would suggest just drop `$HOME/.xkb` as legacy one to teach people good practice.

PR #164, opened 2020-07-13
Add the (v)asprintf_safe helper function

whot:
Return the allocated string or NULL which is a much better interface than the `asprintf` "return length or -1 and the string pointer is undefined". And use that in xkbcomp's include buffer handling because there's no real need for realloc here.

Edit: this is split out from #162

### DIFF
--- a/src/compose/paths.c
+++ b/src/compose/paths.c
@@ -25,6 +25,7 @@
 
 #include "utils.h"
 #include "paths.h"
+#include "utils.h"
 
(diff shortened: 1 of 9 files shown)


whot:
I wouldn't mind that name, it's certainly better than `asprintf_safe`. Could change it as follow-up if you want.

> but don't remember where.

Google can't find it, so maybe it was all a dream ;) Thanks for the review, much appreciated.

PR #165, opened 2020-07-16
Add xdg base directory support to Compose

linkmauve:
Before reading `~/.XCompose`, try to read `$XDG_CONFIG_HOME/XCompose` (falling back to `~/.config/XCompose`).

This helps unclutter the home directory of users who want that.

### DIFF
--- a/src/compose/paths.c
+++ b/src/compose/paths.c
@@ -143,10 +143,10 @@ resolve_locale(const char *locale)
     return alias ? alias : strdup(locale);
 }
 
-const char *
+char *
 get_xcomposefile_path(void)
 {
-    return secure_getenv("XCOMPOSEFILE");
+    return strdup_safe(secure_getenv("XCOMPOSEFILE"));
 }
 
 char *
--- a/src/compose/paths.c
+++ b/src/compose/paths.c
@@ -150,6 +150,23 @@ get_xcomposefile_path(void)
 }
 
 char *
+get_xdg_xcompose_file_path(void)
+{
+    const char *xdg_config_home;
+    const char *home;
+
+    xdg_config_home = secure_getenv("XDG_CONFIG_HOME");
+    if (!xdg_config_home || xdg_config_home[0] != '/') {
+        home = secure_getenv("HOME");
+        if (!home)
+            return NULL;
+        return asprintf_safe("%s/.config/XCompose", home);
+    }
+
+    return asprintf_safe("%s/XCompose", xdg_config_home);
+}
+
+char *
 get_home_xcompose_file_path(void)
 {
     const char *home;
--- a/src/compose/table.c
+++ b/src/compose/table.c
@@ -161,8 +161,7 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
                                   enum xkb_compose_compile_flags flags)
 {
     struct xkb_compose_table *table;
-    char *path = NULL;
-    const char *cpath;
+    char *path;
     FILE *file;
     bool ok;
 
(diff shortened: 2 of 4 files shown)


src/compose/table.c
```diff
@@ -183,6 +183,15 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
             goto found_path;
     }
 
+    cpath = path = get_xdg_xcompose_file_path();
```
whot, line 186: unrelated to this patch but still nice if you could fix this: the `cpath` vs `path` confusion can be fixed by a single `strdup_safe()` call in `get_xcomposefile_path()`

src/compose/paths.c
```diff
@@ -149,6 +149,23 @@ get_xcomposefile_path(void)
     return secure_getenv("XCOMPOSEFILE");
 }
 
+char *
+get_xdg_xcompose_file_path(void)
+{
+    const char *xdg_data_home;
+    const char *home;
+
+    xdg_data_home = secure_getenv("XDG_CONFIG_HOME");
```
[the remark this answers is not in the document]
linkmauve: Woops, this was a leftover from my first idea which was to consider it data, but it’s definitely config.

src/compose/table.c
linkmauve: Done. :)

src/compose/table.c
```diff
@@ -176,14 +175,16 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
-    if (cpath) {
-        file = fopen(cpath, "rb");
+    path = get_xcomposefile_path();
+    if (path) {
+        file = fopen(path, "rb");
         if (file)
             goto found_path;
     }
+    free(path);
+    path = NULL;
```
whot, line 185: nitpick: `path = NULL` is superfluous now since you're setting it in the next line anyway. Same in line 194 and 203.

linkmauve: 212 too, since we then `return NULL;` and don’t use `path` for anything in-between.

whot:
The windows CI failure is `symbols-leak-test.sh` failing with `Windows Subsystem for Linux has no installed distributions.` which suggests there's something wrong with the setup. Don't know enough about Windows to debug this, there are some hits on google for this message but they all suggest running some interactive commands. I'm tempted to think that's an issue with the github runner, maybe it'll go away next time?

Either way, unrelated to this issue.

bluetech:
> The windows CI failure is symbols-leak-test.sh failing with Windows Subsystem for Linux has no installed distributions. which suggests there's something wrong with the setup.

We've had this issue [elsewhere](https://github.com/pytest-dev/pytest/issues/7518). I'll fix it up when I get the chance.

src/compose/table.c
```diff
@@ -176,37 +175,44 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
     log_err(ctx, "couldn't find a Compose file for locale \"%s\"\n", locale);
     xkb_compose_table_unref(table);
     return NULL;
 
 found_path:
-    ok = parse_file(table, file, cpath);
+    ok = parse_file(table, file, path);
     fclose(file);
     if (!ok) {
         xkb_compose_table_unref(table);
```
bluetech, line 218:
```suggestion
        free(path);
        xkb_compose_table_unref(table);
```

xkbcommon/xkbcommon-compose.h
```diff
@@ -204,8 +204,9 @@ enum xkb_compose_format {
  * affected by the following environment variables:
  *
  * 1. `XCOMPOSEFILE` - see Compose(5).
- * 2. `HOME` - see Compose(5).
- * 3. `XLOCALEDIR` - if set, used as the base directory for the system's
+ * 2. `XDG_CONFIG_HOME` - see Compose(5).
```
bluetech, line 207:
```suggestion
 * 2. `XDG_CONFIG_HOME` - before `$HOME/.XCompose` is checked, `$XDG_CONFIG_HOME/XCompose` is checked (with a fall back to `$HOME/.config/XCompose` if `XDG_CONFIG_HOME` is not defined). This is a libxkbcommon extension to the search procedure in Compose(5) (since libxkbcommon 0.11.0). Note that other implementations, such as libX11, might not find a Compose file in this path.
```

(Please wrap this properly, too hard in this box...)

bluetech:
The CI issue is fixed now, but need to rebase on master.

xkbcommon/xkbcommon-compose.h
linkmauve: Thanks, I’ve put this in my commit now. :)

linkmauve:
And rebased!

bluetech:
Thanks @linkmauve!

PR #167, opened 2020-07-22
meson: fix ci on Windows WSL with broken bash

bluetech:
The CI started installing some wrapper instead of a real bash which is what gets found.

See: https://github.com/actions/virtual-environments/pull/1081

~~Note: the manual `run_command` can be replaced with `find_program` once/if we depend on meson>=0.52.0 which added the `version` parameter.~~ Rewritten in Python, no longer relevant.


whot:
> Disabled valgrind wrapper for now because it now also applies to the python interpreter which leaks like a sieve.

Might be worth setting an env var, e.g. `RUNNING_IN_VALGRIND` in the valgrind test run and immediately exiting with 77 if that's set. Wonder how much python leaks if all you do is `os.getenv()` and `sys.exit()`. fwiw, the proper solution would be to add the tests that can be run under valgrind to a meson test suite and running `meson test ---suite valgrind`.

PR #169, opened 2020-07-22
test: make the symbols-leak-test executable

whot:
Python leaks like crazy when run under valgrind. But if we make the script
executable **and** it has uses the env invocation (i.e. #!/usr/bin/env python3),
the leaks disappear. This is not the case for a shebang of /usr/bin/python3.

Why exactly this is the case I'm not sure but executables we plan to run
should have the exec bit set. So this is a janitor patch with the nice side
effect of fixing our valgrind runs.


whot:
First patch is the immediate fix, second patch is is the "proper" fix, at least if we're happy to have our own defined valgrind setup. Feel free to cherry-pick or squash. Unfortunately meson doesn't have the ability to exclude a test from a test setup, so we can't avoid the externally-provided `--no-suite xyz` argument.

meson.build
```diff
@@ -501,6 +502,20 @@ if get_option('enable-xkbregistry')
...
+if valgrind.found()
+    add_test_setup('valgrind',
+        exe_wrapper: [valgrind,
+                       '--leak-check=full',
+                       '--track-origins=yes',
+                       '--gen-suppressions=all',
+                       '--error-exitcode=99'],
+        timeout_multiplier : 10)
+else
+    message('valgrind not found, disabling valgrind test suite')
```
bluetech, line 515:
```suggestion
    message('valgrind not found, disabling valgrind test setup')
```

bluetech:
Merged (with the comment fixed), thanks!

whot:
huh, that explains it. thanks for finding that, I had wondered about it.

PR #170, opened 2020-07-22
meson.build: fix pkgconfig deprecation warning

whot:
../meson.build:239: DEPRECATION: Library xkbcommon was passed to the "libraries"
keyword argument of a previous call to generate() method instead of first
positional argument. Adding xkbcommon to "Requires" field, but this is a
deprecated behaviour that will change in a future version of Meson. Please
report the issue if this warning cannot be avoided in your case.

This was introduced in meson 0.49.0

Note: meson 0.49 was released in December 2018. Feel free to leave this one sitting until we're happy to require a newer meson version.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
         'warning_level=2',
         'b_lundef=true',
     ],
-    meson_version : '>= 0.41.0',
+    meson_version : '>= 0.49.0',
 )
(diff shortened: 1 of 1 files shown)


bluetech:
I tried to keep it compatible, but these warnings are indeed very annoying. December 2018 should be good enough. Also we can start using newer features (at least I remember I avoided some due to version...).

Issue #171, opened 2020-07-25
Add version of linked library in xkbcli


bluetech:
I'm thinking that in addition to `LIBXKBCOMMON_VERSION` (the version we *compile* against), we should also add an exported function to libxkbcommon

```c
struct xkb_library_version {
    int major, minor, patch;
}
struct xkb_library_version xkb_get_library_version();
```

which is the version we *link* against, which might be different.

_Originally posted by @bluetech in https://github.com/xkbcommon/libxkbcommon/pull/160#discussion_r458327568_

bluetech:
Closing per @whot's comment: https://github.com/xkbcommon/libxkbcommon/pull/172#issuecomment-663966020

PR #172, opened 2020-07-25
Add xkb_library_version()

bluetech:
Useful when wanting to query the version that is linked.

Fixes: https://github.com/xkbcommon/libxkbcommon/issues/171

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -59,6 +59,9 @@ endif
 configh_data = configuration_data()
 configh_data.set('EXIT_INVALID_USAGE', '2')
 configh_data.set_quoted('LIBXKBCOMMON_VERSION', meson.project_version())
+configh_data.set('LIBXKBCOMMON_VERSION_MAJOR', meson.project_version().split('.')[0])
+configh_data.set('LIBXKBCOMMON_VERSION_MINOR', meson.project_version().split('.')[1])
+configh_data.set('LIBXKBCOMMON_VERSION_PATCH', meson.project_version().split('.')[2])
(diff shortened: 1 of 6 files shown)


whot:
I really recommend against this, mostly because the result is either unreliable or pointless. We have basically two users: the ones who use a distribution package and the ones who build from git. In the former case, the version number is insufficient because it won't tell you the possible distribution patches [1] and it's better to get the distribution-specific version via `rpm` or whatever. In the  case of git users the version number is wrong anyway because you need the sha, not the version number master happens to have atm.

And then there's the use-case: who needs this? anything that adjusts behaviour at runtime based on the version number will run into the issues above. and anything that doesn't adjust the runtime behaviour doesn't need to care about the version number anyway.

There's a use-case for debugging tools but that's handled by `xkbcli --version` now (and it's conveniently distro-independent). The case of someone having `xkbcli` from a different version than `libxkbcommon.so` is too niche to worry about. So we might as well compile it into that tool with a simple `#define`. There's another use-case of having long-running processes where the library changes while the process is running. But that's another niche case too (and also one of the many reasons distribution updates move more and more towards "you need to restart").

So overall, it is a tempting thing to include but not as useful as it feels at first. I'd say let's stick with `xkbcli --version` as-is which will cover 99% of all use-cases we need to care about.


[1] notorious for that: the RHEL kernel is officially version x.y, in reality x.y + thousands of patches that make it anything but x.y

bluetech:
Thanks for taking the time to explain. I'm convinced that's it's not a good idea. Also, when a function like this is not present since the first release, it's effectiveness is much reduced, for a few years at least.

PR #173, opened 2020-07-27
test: ignore the real XDG_CONFIG_HOME during tests

whot:
Let's not have our tests fail if the user has an incompatible `$XDG_CONFIG_HOME/xkb` directory.

libxkbcommon has fallbacks when XDG_CONFIG_HOME isn't set so we need to override this with a real directory instead of just unsetting it.

But how can this happen? Well, for example if you're trying to run the test suite on git master when you already have a xkb directory for #162 :)

### DIFF
--- a/test/tool-option-parsing.py
+++ b/test/tool-option-parsing.py
@@ -27,6 +27,7 @@ import resource
 import sys
 import subprocess
 import logging
+import tempfile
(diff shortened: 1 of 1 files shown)


whot:
comment added, merging. thanks for the review

PR #177, opened 2020-08-30
tools/xkbcli-scaffold-new-layout: some improvements/fixes

bluetech:
Some misc changes from review and mypy.

I admit I haven't really tested them...


tools/xkbcli-scaffold-new-layout.py
```diff
@@ -100,8 +105,12 @@ def create_rules_template(base_path, ruleset, layout, option):
...
 
 
-def create_symbols_template(basedir, layout_variant, option):
-    if not layout_variant and not option:
+def create_symbols_template(
+    basedir: Path,
+    layout_variant: Optional[Tuple[str, str]],
+    option: Optional[Tuple[str, str]],
+) -> None:
+    if not layout_variant or not option:
```
bluetech, line 113: @whot if you really meant `and` here, let me know and I'll fix it (the problem with `and` in the current code is it will crash if only one is None).

whot: this was indeed intentionally `and`, that crash wasn't intended though. I've fixed this in #182 now and put your rebased commit on top.

whot (commented):
Superseded by #182 which has this commit on top.

PR #178, opened 2020-08-30
tools: convert man pages from man format to mdoc format

bluetech:
The mdoc is more semantic and consistent.

### DIFF
--- a/tools/xkbcli-how-to-type.1
+++ b/tools/xkbcli-how-to-type.1
@@ -1,44 +1,41 @@
-.TH "XKBCLI\-HOW\-TO\-TYPE" "1" "" "" "libxkbcommon manual"
+.Dd July 27, 2020
+.Dt XKBCLI\-HOW\-TO\-TYPE 1
+.Os
 .
-.SH "NAME"
-\fBxkbcli\-how\-to\-type\fR \- query how to type a given Unicode codepoint
+.Sh NAME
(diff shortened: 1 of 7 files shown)


bluetech:
@whot I like the mdoc syntax better, it's a bit more modern/semantic (but just as cryptic). WDYT?

tools/xkbcli-how-to-type.1
```diff
@@ -1,44 +1,41 @@
...
+.Os
+.
+.Sh NAME
+.Nm "xkbcli how\-to\-type"
+.Nd query how to type a given Unicode codepoint
+.
+.Sh SYNOPSIS
+.Nm
+.Op options
+.Ao Ar codepoint Ac
```
whot, line 12: not sure it's my terminal font, but whatever this uses is a weird one: `⟩` vs the normal `)`. nope, firefox says it's weird too :)

whot (approved):
LGTM, thanks. My only comment is that parenthesis weirdness for commands but that's what mdoc does, so *shrug*. I suspect the world will keep on turning, parenthesis oddness or not :)

Thanks, mdoc does indeed look cleaner than roff and seems the better choice here.

tools/xkbcli-how-to-type.1
bluetech: It does look weird, removed it now.

Issue #180, opened 2020-08-30
Avoid PATH_MAX or handle it being undefined


bluetech:
I just remembered that we used to use it in the past, but then got a patch to stop using it because GNU Hurd does not define it so it failed to build there (on Debian). See https://www.gnu.org/software/hurd/hurd/porting/guidelines.html#PATH_MAX_tt_MAX_PATH_tt_MAXPATHL.

PR #181, opened 2020-08-30
meson.build: define PATH_MAX where it's missing

whot:
PATH_MAX not POSIX and can be missing on some systems, notably Windows (which
provides MAX_PATH instead tough) and Hurd. Let's define it to a sane value where
missing, i.e.  the one it's defined as in limits.h

Fixes https://github.com/xkbcommon/libxkbcommon/issues/180

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -124,6 +124,14 @@ else
 endif
 have_getopt_long = cc.has_header_symbol('getopt.h', 'getopt_long',
                                         prefix: '#define _GNU_SOURCE')
+if not cc.has_header_symbol('limits.h', 'PATH_MAX', prefix: system_ext_define)
+    if host_machine.system() == 'windows'
+        # see https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation
+        configh_data.set('PATH_MAX', 260)
+    else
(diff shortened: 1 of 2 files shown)


bluetech:
Will this be applied to Windows then? Because on Windows the max path len is 260, unless the path has some ugly `\\?\` prefix ([ref](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation)).

whot:
> Will this be applied to Windows then?

Fixed :)

meson.build
```diff
@@ -124,6 +124,14 @@ else
 endif
 have_getopt_long = cc.has_header_symbol('getopt.h', 'getopt_long',
                                         prefix: '#define _GNU_SOURCE')
+if not cc.has_header_symbol('limits.h', 'PATH_MAX', prefix: system_ext_define)
+    if cc.get_id() == 'msvc'
```
bluetech, line 128: Since it is an OS definition, would `host_machine.system() == 'windows'` be more appropriate?

whot: fixed. I guess you can tell I'm not doing a lot of cross-platform development...

bluetech (approved):
Thanks, LGTM.

PR #183, opened 2020-09-02
Revert: tools: add xkbcli-scaffold-new-layout as helper tool

whot:
While this tool is useful for users starting with a new keyboard layout, it is a
somewhat bad fit for libxkbcommon. It's the only python tool, we don't even
install it yet (because we're not sure yet what it's supposed to do) and there's
a potential for it to expand into more corner cases.

The only tie it has to libxkbcommon is that it templates the data files that
libxkbcommon reads, but those files are effectively public API.

Let's remove this tool from there and instead move it to a separate git
repository where it can go its own way.

This reverts commit d00cf64dbc586a1052e1f0d0e7f1a48bbff293ec

@bluetech: Something has been off about this tool from the beginning, the above describes why. I leave it up to you if you want this in the xkbcommon github namespace, I think it may actually be better to have it elsewhere just so you don't have to deal with the fallout of bad configurations.


bluetech (approved):
Yeah, if we're unsure about it then let's remove it for now. I think user-configuration.md is pretty good, and most people will just do what the Arch Wiki tells them anyway :)

> I think it may actually be better to have it elsewhere just so you don't have to deal with the fallout of bad configurations.

I already had to deal with one - myself. I was playing with scaffold and left it's output unedited. Next time I came to that machine I tried to launch various applications and nothing happened. It took me a minute to figure out why :)

whot:
I already have an assortment of `$XDG_CONFIG/_HOME/xkb_`, `$XDG_CONFIG_HOME/xkb.disabled`, etc. for that same reason :)

whot:
FTR, I've created https://gitlab.freedesktop.org/whot/xkblayout with that tool now

bam80:
> > better to have it elsewhere just so you don't have to deal with the fallout of bad configurations.

> I already had to deal with one. ... It took me a minute to figure out why :)

For me it took literally a days and even lead to a bug report before I finally found it out )
https://gitlab.gnome.org/GNOME/mutter/-/issues/3347#note_2045123

PR #184, opened 2020-09-05
Bump version to 1.0.0

bluetech:


bluetech:
Putting this here temporarily while resolving some procedural issues.

PR #188, opened 2020-09-07
test/tool-option-parsing: switch from pytest to unittest

bluetech:
For me, installing pytest for libxkbcommon is a bit problematic, so I end up skipping it which is not great.

Switch to unittest which is built in to Python. It's not as nice as pytest but good enough in this case.

Note: I was too lazy to switch the plain asserts to unittest assertions...


bluetech:
Hmm so now it runs on macos, which doesn't have the necessary XKB files, it failed there. I added a second commit to improve the test isolation, let's see if it works.

bluetech:
Ah so now the issue is that macos doesn't build xkbregistry but the test still tries to execute it. We should skip tests which rely on disabled features (x11, wayland, registry). Will fix it up later.

test/tool-option-parsing.py
```diff
@@ -104,200 +99,182 @@ def run_command_missing_arg(self, args):
...
-    # To be filled in if we handle something other than --help
-    pass
-
-
-@pytest.mark.skipif(not os.getenv('WAYLAND_DISPLAY'),
-                    reason='WAYLAND_DISPLAY not set')
-def test_interactive_wayland(xkbcli_interactive_wayland):
-    # To be filled in if we handle something other than --help
-    pass
+class TestXkbcli(unittest.TestCase):
```
whot, line 111: if we're using unittest we should use the one thing it does better than pytest: hierarchy. So one generic `XkbCliBase` class with the `help` and `invalid-option` test, then a subclass of that for every tool. That's what I did in libinput first, see [the switch to pytest commit](https://gitlab.freedesktop.org/libinput/libinput/-/commit/d0eb8d58a3e3c1f1f15e612c732a43486c6479e4) which gives you a nice before/after, you just need to invert this :)

That way it'll be easier to skip a test set too if the registry isn't compiled. Sorry about that bit.

test/tool-option-parsing.py
```diff
@@ -279,10 +280,13 @@ def test_interactive_wayland(self):
 
 if __name__ == '__main__':
     with tempfile.TemporaryDirectory() as tmpdir:
+        # Use our own test xkeyboard-config copy.
+        os.environ['XKB_CONFIG_ROOT'] = top_srcdir + '/test/data'
         # libxkbcommon has fallbacks when XDG_CONFIG_HOME isn't set so we need
         # to override it with a known (empty) directory. Otherwise our test
         # behavior depends on the system the test is run on.
         os.environ['XDG_CONFIG_HOME'] = tmpdir
+        del os.environ['HOME']
```
whot, line 305: we shouldn't need this one right? If we do, then something is likely off with the test.

edit, nevermind, I think that catches the `$HOME/.xkb/` fallback for the merged kcgst files.

test/tool-option-parsing.py
bluetech: > if we're using unittest we should use the one thing it does better than pytest: hierarchy

You can use test classes in pytest as well :wink:

> That's what I did in libinput first, see the switch to pytest commit which gives you a nice before/after, you just need to invert this :)

Thanks, I'll check it out.

bluetech: I ended up not splitting into classes, but made some hack to make it skip more easily. I think it's OK.

test/tool-option-parsing.py
bluetech: Added a comment to clarify.

bluetech:
OK seems to have worked - CI is green now.

test/tool-option-parsing.py
```diff
@@ -105,200 +104,174 @@ def run_command_missing_arg(self, args):
...
+        self.xkbcli.run_command_invalid(['a'] * 64)
+
+    def test_compile_keymap_args(self):
+        for args in (
+            ['--verbose'],
+            ['--rmlvo'],
+            # ['--kccgst'],
+            ['--verbose', '--rmlvo'],
+            # ['--verbose', '--kccgst'],
+        ):
```
whot, line 173: nitpick, this looks like something flake would complain about.

whot (approved):

PR #191, opened 2020-09-10
parser: fix another format string for int64_t

hhb:


src/xkbcomp/parser.y
```diff
@@ -734,7 +734,7 @@ KeySym          :       IDENT
                 |       Integer
                         {
                             if ($1 < 0) {
-                                parser_warn(param, "unrecognized keysym \"%ld\"", $1);
+                                parser_warn(param, "unrecognized keysym \"%" PRId64 "\"", $1);
```
bluetech, line 737:
```suggestion
                                parser_warn(param, "unrecognized keysym \"%"PRId64"\"", $1);
```

bluetech:
Thanks!

Issue #193, opened 2020-10-09
document short/long layout names


bam80:
Well, even if [the issue with 2-letter layouts](https://github.com/xkbcommon/libxkbcommon/issues/192) can be considered as solved for that particular case, overall it feels to me we need document it better somewhere. It is not obvious. Maybe leave some notes in [xkb_keymap_layout_get_name()](https://xkbcommon.org/doc/current/group__components.html#ga63b85a1431caed99af85e19ff50131f6) description itself?
 @bluetech @fooishbar

PR #194, opened 2020-10-09
doc: add note about "short" layout names

bluetech:
Add a clarifying note about "short" layout names to the docs, as suggested by @bam80.

Fixes: https://github.com/xkbcommon/libxkbcommon/issues/193


bam80:
Many thanks! Much better now. I wish I had that from the beginning :)

bam80 (approved):
Excellent!

PR #200, opened 2020-11-18
x11: cache X11 atoms

bluetech:
On every keymap notify event, the keymap should be refreshed, which fetches the required X11 atoms. A big keymap might have a few hundred of atoms.

A [profile by a user](https://github.com/i3/i3/issues/3924#issuecomment-729890732) has shown this *might* be slow when some intensive amount of keymap activity is occurring. It might also be slow on a remote X server.

While I'm not really sure this is the actual bottleneck, caching the atoms is easy enough and only needs a couple kb of memory, so do that.

On the added bench-x11:

```
Before: retrieved 2500 keymaps from X in 11.233237s
After : retrieved 2500 keymaps from X in 1.592339s
```

### DIFF
--- a/bench/x11.c
+++ b/bench/x11.c
@@ -1,0 +1,108 @@
+/*
+ * Copyright © 2020 Ran Benita <ran@unusedvar.com>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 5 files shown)


WGH-:
I actually have mostly similar results on my machine:
```
Before:
retrieved 2500 keymaps from X in 10.779261s

After:
retrieved 2500 keymaps from X in 2.391614s
```

Maybe slow `xkb_x11_keymap_new_from_device` wasn't the real root cause of that problem (see referenced i3 issue) after all, and it's just something else going on that's slowing down X server during input method switch, which collaterally makes `xkb_x11_keymap_new_from_device` slow.

src/x11/util.c
```diff
@@ -163,24 +177,48 @@ adopt_atoms(struct xkb_context *ctx, xcb_connection_t *conn,
...
+                for (size_t c = 0; c < cache->len; c++) {
+                    if (cache->cache[c].from == from[i]) {
+                        to[i] = cache->cache[c].to;
+                        goto cache_hit;
+                    }
+                }
+            }
             if (from[i] != XCB_ATOM_NONE)
                 cookies[i % SIZE] = xcb_get_atom_name(conn, from[i]);
+            cache_hit: ;
```
whot, line 209: an indented label with a semicolon is a bit of a strange look tbh. How about something like this instead:

```c
   for (size_t i = start; i < stop; i++) {
       bool cache_hit = false;
       for (size_t c = 0; cache && c < cache->len; c++) {
           if (cache->cache[c].from == from[i]) {
               to[i] = cache->cache[c].to;
               cache_hit = true;
               break;
           }
       }
       if (!cache_hit && from[i] != XCB_ATOM_NONE)
           cookies[i % SIZE] = xcb_get_atom_name(conn, from[i]);
   }
```

bench/x11.c
```diff
@@ -0,0 +1,108 @@
...
+    elapsed = bench_elapsed_str(&bench);
+    fprintf(stderr, "retrieved %d keymaps from X in %ss\n",
+            BENCHMARK_ITERATIONS, elapsed);
+    free(elapsed);
+
+    xkb_context_unref(ctx);
+err_conn:
+    xcb_disconnect(conn);
+err_out:
+    exit(ret == 0 ? EXIT_SUCCESS : EXIT_FAILURE);
```
whot, line 107: `return` at the end of main() seems more conventional :)

whot (approved):

bluetech:
Thanks for the review @whot.

PR #213, opened 2021-02-17
scripts: update makeheader script for the _EVDEVK keysym defines

whot:
As of xorgproto commit 5dbb5b76597f [1], the 0x10081XXX keycode range is defined
for direct evdev kernel keycode mapping. For example, KEY_MACRO1 (0x290) is
mapped to 0x10081290. The format of the #define lines for these keys is
stable to allow for parsing:

    #define XF86XK_FooBar _EVDEVK(0x123) /* optional comment */

Update our script so we detect these new lines. Our keysym generation is a
two-step process: makeheader and then makekeys. Replacing the key with its full
value in the makeheader script means we don't have to update makekeys to handle
the _EVDEVK macro and our header file is fully resolved.

[1] https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/merge_requests/23


whot:
A note here: this makes the `a/xkbcommon/xkbcommon-keysyms.h` a bit awkward. We don't touch the xorgproto `/* Use: XK_foo` comment lines, so known keycodes are expanded but the others still refer to the `_EVDEVK` macro.

```
#define XKB_KEY_XF86WPSButton          0x10081211              /* v2.6.34 KEY_WPS_BUTTON */
/* Use: XF86XK_TouchpadToggle          _EVDEVK(0x212)             v2.6.37 KEY_TOUCHPAD_TOGGLE */
```
We could keep the `_EVDEVK` macro here if you prefer, just means we have to update the other scripts to parse that too. Not 100% what my preference is here.

scripts/makeheader
```diff
@@ -3,6 +3,9 @@ from __future__ import print_function
 import re
 import os
 
+# expected format:
+# #define XF86XK_FooBar _EVDEVK(0x123) /* some optional comment */
+evdev_pattern = re.compile(r'^#define\s+XF86XK_(?P<name>\w+)\s+_EVDEVK\((?P<value>0x[0-9A-F]+)\)')
```
bluetech, line 8: Just to be on the safe side, would allow lowercase in the hex:

```suggestion
evdev_pattern = re.compile(r'^#define\s+XF86XK_(?P<name>\w+)\s+_EVDEVK\((?P<value>0x[0-9A-Fa-f]+)\)')
```

bluetech (approved):
I think it's fine to leave the comments as-is, so I think this is good, thanks (feel free to merge)!

Would you like to also run the update in the PR, or later?

scripts/makeheader
whot: updated in latest version (both instances, there was another regex further down), thanks.

whot:
A few extra changes, the keysyms are now synched with xorgproto master. And I've added a test case to check that we're actually resolving keysyms.

bluetech:
Can we move the test to Python or C? We used to have a bash test but I removed it a while ago.

whot:
> Can we move the test to Python or C? We used to have a bash test but I removed it a while ago.

Your wish, my command, etc. etc. :)

whot:
ok, macos is happy now too. Yay!

bluetech:
Thanks @whot! I think I can do a release this weekend.

whot:
doh *facepalm*. I just realized haven't actually made a xorgproto release yet, until that's out we probably shouldn't add all the keysyms. I'll try to get this done this week and let you know if anything comes up.

Thanks for the merge though, much appreciated.

bluetech:
OK, ping me when you need a release!

whot:
all good, `xorgproto 2021.2` is out with all the keysyms, so we're good for a release here too. thanks.

PR #214, opened 2021-02-26
Fix a few keysymtab entries to match their comment in xkbcommon-keysyms.h

wismill:
Fix a few `keysymtab` entries to match their comment in `xkbcommon-keysyms.h`.

Note: there are quite a few keysyms that do not have a comment but have an entry in `keysymtab`.

Ideally `keysymtab` could be auto-generated from `xkbcommon-keysyms.h`. This requires to update the corresponding files in [xorgproto](https://gitlab.freedesktop.org/xorg/proto/xorgproto) if I understand well.


src/keysym-utf.c
```diff
@@ -810,7 +810,7 @@ static const struct codepair keysymtab[] = {
     { 0x0ef0, 0x3171 }, /*    Hangul_SunkyeongeumMieum ㅱ HANGUL LETTER KAPYEOUNMIEUM */
     { 0x0ef1, 0x3178 }, /*    Hangul_SunkyeongeumPieub ㅸ HANGUL LETTER KAPYEOUNPIEUP */
     { 0x0ef2, 0x317f }, /*              Hangul_PanSios ㅿ HANGUL LETTER PANSIOS */
-/*  0x0ef3                  Hangul_KkogjiDalrinIeung ? ??? */
+    { 0x0ef3, 0x3181 }, /*    Hangul_KkogjiDalrinIeung ㆁ HANGUL LETTER YESIEUNG */
```
bluetech, line 813: Would be great if some familiar with Ancient Hangul Consonant Characters could confirm this. The [mgk25 pdf](https://www.cl.cam.ac.uk/~mgk25/ucs/keysyms.pdf) does not state where they were taken from and does not specify a Unicode equivalent for them, and searching for "Hangul Kkogji Dalrin Ieung" does not bring up something I am able to understand...

.gitignore
```diff
@@ -0,0 +1 @@
+build/
```
bluetech, line 1: We recently removed this, for now should add it to your private .git/info/exclude file.

bluetech (commented):
Thanks @wismill. Please see my comments.

Since I am just doing a release now, I'm going to pick the TACK fixes to master, since it's a clear bug fix.

src/keysym-utf.c
bluetech: OK, I found this was recently added in xorgproto: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/commit/959d4442e46d8d3bef845e526219d23346a1328e

I assume the proper research was done there, so I'll pick this up too.

wismill: I agree. I just applied the character in the comment, I am no expert.

.gitignore
wismill: Ok

src/keysym-utf.c
```diff
@@ -614,8 +614,8 @@ static const struct codepair keysymtab[] = {
     { 0x0bd6, 0x222a }, /*                    downshoe ∪ UNION */
     { 0x0bd8, 0x2283 }, /*                   rightshoe ⊃ SUPERSET OF */
     { 0x0bda, 0x2282 }, /*                    leftshoe ⊂ SUBSET OF */
-    { 0x0bdc, 0x22a2 }, /*                    lefttack ⊢ RIGHT TACK */
-    { 0x0bfc, 0x22a3 }, /*                   righttack ⊣ LEFT TACK */
+    { 0x0bdc, 0x22a3 }, /*                    lefttack ⊣ LEFT TACK */
```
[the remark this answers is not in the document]
wismill: It is confusing ;-) I just applied the character in the comment to match the reference.

src/keysym-utf.c
```diff
@@ -536,9 +536,9 @@ static const struct codepair keysymtab[] = {
     { 0x0ab7, 0x215a }, /*                  fivesixths ⅚ VULGAR FRACTION FIVE SIXTHS */
     { 0x0ab8, 0x2105 }, /*                      careof ℅ CARE OF */
     { 0x0abb, 0x2012 }, /*                     figdash ‒ FIGURE DASH */
-    { 0x0abc, 0x27e8 }, /*            leftanglebracket ⟨ MATHEMATICAL LEFT ANGLE BRACKET */
```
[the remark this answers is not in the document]
wismill: Ok.

wismill:
I have a general question: as I understand, the translation from keysym to UCS happens in Xlib in the file [`imKStoUCS`](https://gitlab.freedesktop.org/xorg/lib/libx11/-/blob/master/src/xlibi18n/imKStoUCS.c), which is quite out of sync with [xorg proto](https://gitlab.freedesktop.org/xorg/proto/xorgproto/). It is quite confusing. 

I think `xkbcommon-keysyms.h` should be the reference for xkbcommon and exactly matches the Unicode translations used in xkbcommon.

1. Should xorgproto repo be updated to match xkbcommon Unicode translations? Then sync xlib or leave it out of sync as legacy software?
2. Or should xkbcommon python script be updated to deal with this divergences: when xkbcommon and xorgproto diverge, use the Unicode value of xkbcommon in the comment `/*U+NNNN XXXX */` and add comment _before_ the entry to document this divergence?

I think option 1 makes more sense.

.gitignore
whot: unresolving, @wismill please push the fixed commits first, then mark as resolved, thanks. Too hard to keep track of things otherwise.

wismill: Done. Since part of this MR was merge, I thought it would be closed as it happened to me in other project. Sorry for the noise ;-)

bluetech:
> I have a general question: as I understand, xorg proto is the reference for both Xlib and xkbcommon.
>
> The translation from keysym to UCS happens in Xlib in the file imKStoUCS, which is quite out of sync with xorg proto. It is quite > confusing

xorg-proto is out upstream for keysyms (`xkbcommon-keysyms.h`) but not for keysym->unicode translation. The unicode translation is rather based on [this file](https://www.cl.cam.ac.uk/~mgk25/ucs/keysym2ucs.c) with some subsequent updates. I think they both come from the same source in the end though. I agree it's confusing.

> I think xkbcommon-keysyms.h should be the reference for xkbcommon and exactly matches the Unicode translations used in xkbcommon.

Agreed, but because of the different sources it is not always possible.

The left angle confusion happened because [this xorgproto](https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/commit/a8ccf66bc924fb02c9e944f99e84d90958dd142f) commit seems to have reached a different conclusion than [we did](b63196e91e7d78e88a9012f3f44152eeef5358cf), so after a recent [sync](0bfc71e7b522ab3d15f96746f7559960c2ae6d21), it got mismatched again.

TBH, at this point the leftangle business is angling into my left ear and immediately out the right, so I'd be happy if you can sort it out for me :)

Issue #216, opened 2021-03-06
Many round-trips during connection setup


psychon:
I wrote a small-ish [python program](https://gist.github.com/psychon/b3272c4d51383015657f6f854a4caad6) that listens on a TCP port and forwards all connections to another port, but adds an artificial delay. For example, running `Xephyr -listen tcp :1 & ./tcp-proxy.py 6010 6001 0.05` provides an X11 `DISPLAY=:10` with a round-trip-time of 0.1 seconds.

<details><summary>Next, I hacked `test/x11.c` a bit to show up long things take.</summary>

```diff
diff --git a/test/x11.c b/test/x11.c
index 00f3a96..a35eb69 100644
--- a/test/x11.c
+++ b/test/x11.c
@@ -25,6 +25,19 @@
 
 #include "test.h"
 #include "xkbcommon/xkbcommon-x11.h"
+#include <sys/time.h>
+
+static struct timeval last_time;
+
+static void
+elapsed(const char *name)
+{
+    struct timeval now, diff;
+    gettimeofday(&now, NULL);
+    timersub(&now, &last_time, &diff);
+    last_time = now;
+    printf("Needed %ld.%06ld for %s\n", diff.tv_sec, diff.tv_usec, name);
+}
 
 int
 main(void)
@@ -43,11 +56,13 @@ main(void)
     * If it fails, it's not necessarily an actual problem with the code.
     * So we don't want a FAIL here.
     */
+    gettimeofday(&last_time, NULL);
     conn = xcb_connect(NULL, NULL);
     if (!conn || xcb_connection_has_error(conn)) {
         exit_code = SKIP_TEST;
         goto err_conn;
     }
+    elapsed("xcb_connect()");
 
     ret = xkb_x11_setup_xkb_extension(conn,
                                       XKB_X11_MIN_MAJOR_XKB_VERSION,
@@ -58,20 +73,25 @@ main(void)
         exit_code = SKIP_TEST;
         goto err_conn;
     }
+    elapsed("xkb_x11_setup_xkb_extension()");
 
     device_id = xkb_x11_get_core_keyboard_device_id(conn);
     assert(device_id != -1);
+    elapsed("xkb_x11_get_core_keyboard_device_id()");
 
     keymap = xkb_x11_keymap_new_from_device(ctx, conn, device_id,
                                             XKB_KEYMAP_COMPILE_NO_FLAGS);
     assert(keymap);
[...]

PR #217, opened 2021-03-07
Reduce number of X11 round trips

psychon:
This is my attempt at fixing #216.

As written there, before this PR I got this output and saw 78 round trips just for `xkb_x11_keymap_new_from_device`:

```
Needed 0.104062 for xcb_connect()
Needed 0.201753 for xkb_x11_setup_xkb_extension()
Needed 0.100868 for xkb_x11_get_core_keyboard_device_id()
Needed 7.851895 for xkb_x11_keymap_new_from_device()
Needed 0.100844 for xkb_x11_state_new_from_device()
Needed 0.003317 for xkb_keymap_get_as_string()
Needed 0.000224 for cleanup()
```

With this PR, my output is (and there are only three round trips for `new_from_device`):

```
Needed 0.104250 for xcb_connect()
Needed 0.201609 for xkb_x11_setup_xkb_extension()
Needed 0.100910 for xkb_x11_get_core_keyboard_device_id()
Needed 0.348044 for xkb_x11_keymap_new_from_device()
Needed 0.100926 for xkb_x11_state_new_from_device()
Needed 0.003080 for xkb_keymap_get_as_string()
Needed 0.000198 for cleanup()
```

Via gdb I looked at where the round trips during `test/x11` happen (just have the TCP proxy delay things by a second and then you have enough time to `ctrl+c` and `bt` in gdb to see where the test is waiting):
1. `xcb_connect` needs one round trip to set up the connection
2. Using the XKB extension requires a round trip in `xcb_get_extension_data` for the QueryExtension reply (this is internal to libxcb)
3. `xkb_x11_setup_xkb_extension` sends an XKB UseExtension request and waits for the reply
4. `xkb_x11_get_core_keyboard_device_id` obviously needs one round trip
5. the requests in `xkb_x11_keymap_new_from_device` need a round trip (the backtrace actually points at waiting for the XKB GetMap reply) [this is the first round trip in `xkb_x11_keymap_new_from_device`
6. getting the atom names in `get_names()` uses a round trip
7. getting the atom names in `x11_atom_interner_round_trip()` uses a round trip [this is the last round trip from `xkb_x11_keymap_new_from_device`]
8. `xkb_x11_state_new_from_device()` needs a round trip for an XKB GetState request

The above looks quite good. It would be possible to merge get GetAtomName stuff in step 6 and 7 in a single round trip, but at this point I stop caring. I got `xkb_x11_keymap_new_from_device` down from 78 round trips to three. Getting this down to the theoretical minimum of two round trips is left as an exercise to the reader. ;-)

## Thoughts on correctness

- I am fairly sure that the code does not actually look at any of the atoms and hence my approach to batch `GetAtomName` requests should not cause issue. Still, please double-check this.
- The new error handling here is obviously untested. To simplify the error handling, I no longer "bail out" after the first error (which would require doing the right amount of `xcb_discard_reply()` calls), but instead continue through and only "bail out" at the end. IMHO this should not be much of a downside.

I tried to follow the coding style of the surrounding code, but I did not try too hard. Feel free to tell me how to change things so that they better fit.

### DIFF
--- a/src/x11/util.c
+++ b/src/x11/util.c
@@ -169,14 +169,9 @@ struct x11_atom_cache {
     size_t len;
 };
 
-bool
-adopt_atoms(struct xkb_context *ctx, xcb_connection_t *conn,
-            const xcb_atom_t *from, xkb_atom_t *to, const size_t count)
+static struct x11_atom_cache *
+get_cache(struct xkb_context *ctx, xcb_connection_t *conn)
 {
-    enum { SIZE = 128 };
-    xcb_get_atom_name_cookie_t cookies[SIZE];
-    const size_t num_batches = ROUNDUP(count, SIZE) / SIZE;
-
     if (!ctx->x11_atom_cache) {
         ctx->x11_atom_cache = calloc(1, sizeof(struct x11_atom_cache));
     }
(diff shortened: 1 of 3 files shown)


psychon:
> Getting this down to the theoretical minimum of two round trips is left as an exercise to the reader. 

Well, I couldn't resist. I just pushed one more commit which also gets rid of this one round trip. New timings with a round trip time of one second (instead of the 0.1 seconds I used in previous tests):

```
Needed 1.005131 for xcb_connect()
Needed 2.003656 for xkb_x11_setup_xkb_extension()
Needed 1.001808 for xkb_x11_get_core_keyboard_device_id()
Needed 2.045865 for xkb_x11_keymap_new_from_device()
Needed 1.001911 for xkb_x11_state_new_from_device()
Needed 0.003242 for xkb_keymap_get_as_string()
Needed 0.000202 for cleanup()
```

bluetech:
Hi @psychon, always nice to have an expert optimize things. I'll try to review this soon.

I quickly tried it locally (xwayland) on `./bench/bench-x11`, and even without any added latency I see it runs about 2 times faster, which is great.

psychon:
> always nice to have an expert optimize things. 

Heh, thanks for that compliment. However it's just "`xcb_*_reply()` has to wait for the reply from the X11 server; it makes sense to send as many requests as possible before calling any reply function so that the next reply will already be available by the time we need it" (hm, okay, that sentence turned out longer that I expected... still, it's relatively simple.)

>  I'll try to review this soon.

No hurry. Thanks to this weird phenomenon called "a pandemic", I'll have time to react to reviews on most weekends. :-) 

> I quickly tried it locally (xwayland) on ./bench/bench-x11, and even without any added latency I see it runs about 2 times faster, which is great.

Ah, thanks. I totally missed that this exists. Seems like I accidentally optimised something you care about. :-)
Thanks for checking and I'm glad I didn't just look into a problem that no one cares about.

bluetech:
I only started to care about it recently, when a user (@WGH-) reported some slowness - see middle of this issue https://github.com/i3/i3/issues/3924. But I only added atom caching for optimizing repeated calls, didn't optimize the call itself.

src/x11/util.c
```diff
@@ -186,79 +150,155 @@ adopt_atoms(struct xkb_context *ctx, xcb_connection_t *conn,
...
+    interner->pending[idx].from = atom;
+    interner->pending[idx].out = out;
+    interner->pending[idx].cookie = xcb_get_atom_name(interner->conn, atom);
+}
 
-            reply = xcb_get_atom_name_reply(conn, cookies[i % SIZE], NULL);
-            if (!reply)
-                goto err_discard;
+void
+x11_atom_interner_adopt_atoms(struct x11_atom_interner *interner,
```
bluetech, line 216: The `x11_atom_interner_adopt_atoms` function is pretty useless now, I suggest we get rid of it (i.e. just inline the loop).

bluetech (approved):
The delayed writing is pretty clever, I wouldn't have thought to do it.

I was curious if the cache is still helpful; unfortunately, looks like it is.

I only have one minor suggestion, other than that, LGTM.

bluetech:
Thanks @psychon!

I'm kinda curious how libX11 fares with this, they do have to do the same thing, although they fetch much less (we need to reconstruct the entire `xkb_keymap` while they only fetch the client-processing parts).

Regarding a release, I'll try to do one in the next few weeks so it doesn't linger too long.

psychon:
> I'm kinda curious how libX11 fares with this,

Uhm, okay...

Short version: I do not know.

Long version:
```
$ git grep KeymapNotify src/
src/EvToWire.c:       case KeymapNotify:
src/XlibInt.c:     * KeymapNotify has no sequence number, but is always guaranteed
src/XlibInt.c:    if ((rep->type & 0x7f) == KeymapNotify)
src/XlibInt.c:        case KeymapNotify:
src/evtomask.c: KeymapStateMask,                                /* KeymapNotify */
```
None of this seems relevant (the most expensive thing I found is a `memcpy`; 
```
$ git grep MappingNotify src/
src/EvToWire.c:       case MappingNotify:
src/XlibInt.c:        case MappingNotify:
src/evtomask.c: 0,                                              /* MappingNotify */
src/xkb/XKBUse.c:            ev->type = MappingNotify;
src/xkb/XKBUse.c:            ev->type = MappingNotify;
```
Still nope. At least I found the right file: `XKBUse.c` contains a `wire_to_event` function. This contains things like this (in response to XkbMapNotify):
```c
             XkbNoteMapChanges(&xkbi->changes, mev, XKB_XLIB_MAP_MASK);
             if (xkbi->changes.changed)
                 xkbi->flags |= XkbMapPending;
```
Other interesting cases are XkbNewKeyboardNotify that also sets `XkbMapPending` and additionally `XkbXlibNewKeyboard`. All the other cases seem to just translate the event from the wire representation to an `XEvent`.

`XkbNoteMapChanges` doesn't really seem to do anything. Seems to be part of the event translation. Sigh.

Via `XkbMapPending`, we find:
```
#define _XkbCheckPendingRefresh(d,xi) { \
    if ((xi)->flags&XkbXlibNewKeyboard) \
        _XkbReloadDpy((d)); \
    else if ((xi)->flags&XkbMapPending) { \                                                                              
        if (XkbGetMapChanges((d),(xi)->desc, &(xi)->changes)==Success) { \
            LockDisplay((d)); \
            (xi)->changes.changed= 0; \
            UnlockDisplay((d)); \
        } \
    } \
}
```
That seems more relevant. Grepping for this macro says: The update is only done when the next function that needs the current keyboard "stuff" is called.

`XkbGetMapChanges` does a GetMap request (one round trip without anything else). We then end up in `_XkbReadGetMapReply` which... does things? With arrays? And error handling? In hard to read C code?

Okay, different idea:
```
$ git grep GetAtomName src/xkb/
src/xkb/XKB.c:XkbGetAtomNameFunc      _XkbGetAtomNameFunc = XGetAtomName;
src/xkb/XKB.c:XkbSetAtomFuncs(XkbInternAtomFunc getAtom, XkbGetAtomNameFunc getName)
src/xkb/XKB.c:    _XkbGetAtomNameFunc = (getName ? getName : XGetAtomName);
src/xkb/XKBlibint.h:extern  XkbGetAtomNameFunc      _XkbGetAtomNameFunc;
```
Great. There is a function pointer involved which allows to replace `XGetAtomName` with something else. But nothing uses the function pointer!?

I guess it does not?

> Regarding a release, I'll try to do one in the next few weeks so it doesn't linger too long.

You are definitely better at "release early, release often" than I am. :-)

psychon:
> I'm kinda curious how libX11 fares with this, they do have to do the same thing, although they fetch much less

Actually, no. libX11 does not need `GetAtomName` since it can just use "normal" X11 atoms. Only libxkbcommon has to deal with translating between different kinds of atoms. (But there also do not really seem to be any `InternAtom` calls in `src/xkb` - why does libxkbcommon need all these atoms anyway?)

And for the "other things" like GetControls and GetCompatMap: I don't see any libX11-internal calls to `XkbGetCompatMap` or `XkbGetControls`. I do not really know XKB, so I cannot say why libxkbcommon needs this information but libX11 does not.

bluetech:
Right, from looking at your pointers, my understanding is that libX11 only needs the `XkbGetMap` request for normal key-processing, so it only does that. It does *provide* `XkbGetIndicatorMap`/`XkbGetCompatMap`/`XkbGetNames`/`XkbGetControls`, but it's up to the library users to call them if they want them.

I didn't dig into the pending stuff, maybe it's doing a similar optimization as you did for i3, maybe it's something else...

> why does libxkbcommon need all these atoms anyway?

So, an xkbcommon `xkb_keymap` is really the server-side keymap, i.e. it contains the full keymap details. Wayland compositors for example compile the keymaps from source and use them in the "server". The initial idea was to also convert the X server itself to use xkbcommon but that never panned out. Anyway, the `xkb_keymap` is immutable and doesn't have any "partial mode", it needs to be complete. So in order to provide a proper `xkb_keymap` to X11 clients, we need to fetch all of the info, including the atoms, because they are part of the keymap, and there are APIs which access them. This is unlike the X11 XKB protocol which I guess tried to be really efficient and did the mutable, partial, delta updates thing, and ended up super complex because of it.

PR #221, opened 2021-03-28
interactive-evdev: switch from epoll(2) to poll(2)

bluetech:
Turns out FreeBSD supports evdev, so this toll can work on it; however
it does not support epoll, so switch to poll, which is portable.

Reported-by: Evgeniy Khramtsov <evgeniy@khramtsov.org>

### DIFF
--- a/tools/interactive-evdev.c
+++ b/tools/interactive-evdev.c
@@ -31,6 +31,7 @@
 #include <getopt.h>
 #include <limits.h>
 #include <locale.h>
+#include <poll.h>
 #include <signal.h>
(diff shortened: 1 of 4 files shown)


whot:
btw, there's an epoll shim for BSD, see [libinput's meson.build](https://gitlab.freedesktop.org/libinput/libinput/-/blob/master/meson.build#L221). We've had this for a while and I have yet to see complaints about it, so let's assume it works :)

bluetech:
Right, Evgeniy sent a patch by mail which does that, but I suggested just using poll instead.

PR #222, opened 2021-03-29
xkbcomp: use memcpy over strncpy to avoid analyzer warnings

whot:
The target buffer is 7 bytes long, null-termination is optional (as the comment
already suggests). Coverity is unhappy about this though so let's use memcpy
instead, it does the same thing here.

### DIFF
--- a/src/xkbcomp/action.c
+++ b/src/xkbcomp/action.c
@@ -700,15 +700,16 @@ HandlePrivate(struct xkb_context *ctx, const struct xkb_mod_set *mods,
 
             str = xkb_atom_text(ctx, val);
             len = strlen(str);
-            if (len < 1 || len > 7) {
+            if (len < 1 || len > sizeof(act->data)) {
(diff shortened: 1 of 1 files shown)


bluetech:
`strncpy` also pads with zeros, IIRC this was also the intention here. IIRC in xkbcomp this literally outputted uninitialized garbage in the keymap...

whot:
> strncpy also pads with zero

whoah. I literally never noticed that...

bluetech:
I think we should do this:

```c
/* If str is less than 7 bytes, the rest is taken as zeros. */
memset(act->data, 0, sizeof(act->data));
memcpy(act->data, str, len);
```

Also changing the hardcoded 7 a few lines above with `sizeof(act->data)` would be nice, so it's clear that the `len` is indeed valid.

whot:
Updated as requested!

bluetech:
Thanks!

PR #224, opened 2021-03-29
Scan-build fixes

whot:
There are a few more that require investigation, but these are the obvious fixes :)


bluetech:
Thanks, LGTM. I think the streq commit *might* have some performance implications, but probably not, and I'm too lazy to check :)

PR #228, opened 2021-04-06
US-intl AC11 key is broken.

sebastienPoussard:
Recent upgrade in Arch package showed that for the US Intl layout, configuration is broken.
While typing the AC11 deadkey, you will obtain  `´` instead of `'`
I´m not sure to fully understand the code but maybe changing the order of this file might resolve this issue.

### DIFF
--- a/test/data/symbols/us
+++ b/test/data/symbols/us
@@ -125,7 +125,7 @@ xkb_symbols "intl" {
 
     key <AC09> { [	   l,          L,        oslash,         Ooblique ] };
     key <AC10> { [ semicolon,      colon,     paragraph,           degree ] };
-    key <AC11> { [dead_acute, dead_diaeresis, apostrophe,        quotedbl ] };
+    key <AC11> { [dead_apostrophe, dead_quotedbl, dead_acute, dead_diaeresis ] };
 
     key <AB01> { [	   z,          Z,            ae,               AE ] };
     key <AB02> { [	   x,          X,             x,                X ] };


bluetech:
Hi @sebastienPoussard,

This test/ file is only used by tests. The real file comes from another project, xkeyboard-config. But the problem is likely not there. In the latest libxkbcommon release, I have made changes to the Compose implementation, and it's quite possible I broke something.

Can you explain with more details:
- Which locale you use (output of the command `locale` in the terminal)
- Which keyboard layouts you use (only US intl?)
- Which key sequence you type, what is the expected output, and what is the actual output?

BastienM:
Hi @bluetech,

Sebastien and I are currently trying to debug this together. And while he's using the latest (`1.2.x`), I have downgraded to `1.1` and still suffers from this issue.

```sh
$ locale
LANG=en_US.UTF-8
LC_CTYPE="C"
LC_NUMERIC="C"
LC_TIME="C"
LC_COLLATE="C"
LC_MONETARY="C"
LC_MESSAGES="C"
LC_PAPER="C"
LC_NAME="C"
LC_ADDRESS="C"
LC_TELEPHONE="C"
LC_MEASUREMENT="C"
LC_IDENTIFICATION="C"
LC_ALL=C
```
 Personally I have been using the us intl for almost two years without any problems.
 
 As for the sequence `'` + `<space>` would give me a `´`.

bluetech:
Hmm, if it is also broken in xkbcommon 1.1, then it's not due to the recent changes. I tried to reproduce it here but it seems to work as expected.

The relevant compose sequence is:

```
<dead_acute> <space>             »      : "'"   apostrophe # APOSTROPHE
```

And the relevant layout key is:

```
key <AC11> { [dead_acute, dead_diaeresis, apostrophe,        quotedbl ] };
```

and they both haven't changed in a long time.

Can you try the following:

1. Run `sudo xkbcli interactive-evdev --layout us --variant intl` in a terminal and type the AC11 key and the space key, then type Ctrl-C and pase the output.
2. Run `sudo xkbcli interactive-evdev --layout us --variant intl --enable-compose` and do the same.

BastienM:
I was understand this impression as well, but this part of OS is still quite new to me.

Here you go:
```sh
⇒ pacman -Qs libxkbcommon
local/libxkbcommon 1.1.0-2
    Keymap handling library for toolkits and window systems
local/libxkbcommon-x11 1.1.0-2
    Keyboard handling library using XKB data for X11 XCB clients

⇒ sudo xkbcli interactive-evdev --layout us --variant intl
keysyms [ dead_acute       ] unicode [  ] layout [ English (US, intl., with dead keys) (0) ] level [ 0 ] mods [ ] leds [ ] 
keysyms [ space            ] unicode [   ] layout [ English (US, intl., with dead keys) (0) ] level [ 0 ] mods [ ] leds [ ] 

⇒ sudo xkbcli interactive-evdev --layout us --variant intl --enable-compose
keysyms [ apostrophe       ] unicode [ ' ] layout [ English (US, intl., with dead keys) (0) ] level [ 0 ] mods [ ] leds [ ] 
keysyms [ Control_L        ] unicode [  ] layout [ English (US, intl., with dead keys) (0) ] level [ 0 ] mods [ ] leds [ ] 
keysyms [ c                ] unicode [  ] layout [ English (US, intl., with dead keys) (0) ] level [ 0 ] mods [ Control ] leds [ ]
```

bluetech:
Thanks. So this seems correct - `apostrophe` is the expected result. Can you try with libxkbcommon 1.2.0, just to make sure it didn't change with it? If that also works, then I can't really tell where the problem might be coming from.

BastienM:
See my edit, same behavior on both versions.
The only common point on this case is Arch, and as everything seems working as expected on this side, the only logical element left is on Arch´s side.
And we are even using two different DE (Qtile for him and Gnome for me).

Thanks for giving it a look :+1:

whot:
ftr, mclasen had an [xkeyboard-config bug](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/262) where he mentioned this problem and it's likely caused by GTK.

bluetech:
Closing then!

PR #231, opened 2021-04-10
doc: fix user-configuration sample file

joukewitteveen:
Support copy-pasting from the docs to get something functional.
Parsing errors may end up in the journal, where they are not super easy to find.


bluetech (approved):
Definitely, thanks!

PR #238, opened 2021-04-21
keysym: fix underflow in xkb_keysym_from_name

### LINKED ISSUE #237: Segfault in case insensitive xkb_keysym_from_name() when passed empty string since 1.2.0
To reproduce run this code:
```c
#include <xkbcommon/xkbcommon.h>
int main(void) {
    xkb_keysym_t k = xkb_keysym_from_name("", XKB_KEYSYM_CASE_INSENSITIVE);
    printf("%d\n", k);
}
```

ifreund:
This is hit when passing an empty string and XKB_KEYSYM_CASE_INSENSITIVE
currently, causing the array access into name_to_keysym to be out of
bounds on the next iteration of the loop.

Fixes #237

### DIFF
--- a/src/keysym.c
+++ b/src/keysym.c
@@ -69,9 +69,9 @@ xkb_keysym_get_name(xkb_keysym_t ks, char *buffer, size_t size)
         return -1;
     }
 
-    size_t lo = 0, hi = ARRAY_SIZE(keysym_to_name) - 1;
+    int32_t lo = 0, hi = ARRAY_SIZE(keysym_to_name) - 1;
     while (hi >= lo) {
-        size_t mid = (lo + hi) / 2;
+        int32_t mid = (lo + hi) / 2;
         if (ks > keysym_to_name[mid].keysym) {
             lo = mid + 1;
         } else if (ks < keysym_to_name[mid].keysym) {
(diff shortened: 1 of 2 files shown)


ifreund:
Fixed my indentation and added test cases to prevent regression.

bluetech:
Whoops, should always test these cases.

I think perhaps a nicer way to fix the underflow would be to just replace `size_t` -> `ssize_t` (for `mid`, `lo` and `hi`), then the low and high cases remain symmetric. Also in `xkb_keysym_get_name` in the same file. WDYT?

ifreund:
Switching to `ssize_t` is nicer and gets rid of the extra branch, good call. I didn't find an input that caused `xkb_keysym_get_name` to fail so I haven't added a test for that. I do believe the overflow could still occur in theory though, especially if the surrounding code is modified in the future.

bluetech:
Argh, looks like `ssize_t` is not defined on Windows (I always thought it's standard C but apparently it's only POSIX).

Let's just use `int32_t` I guess? The sizes are statically bounded so it doesn't really matter.

ifreund:
That's annoying :/ 

Switched to `int32_t` and amended the commit message.

bluetech:
Thanks! I'll try to release this some time this week.

PR #239, opened 2021-04-22
change the xkbcli list output to YAML

whot:
We have a lot of keyboard layouts and the current output format is virtually
useless at searching for a specific one to debug any issues with either the
layout list or the output from libxkbregistry.

Let's use YAML instead because that can easily be post-processed to extract the
specific layouts wanted, e.g. to get the list of all layouts:
  `xkbcli-list | yq -r ".layouts[].layout"`
to get the list of all variants of the "us" layout:
  `xkbcli-list | yq -r '.layouts[] | select(.layout == "us") | .variant`
and the number of option groups:
  `xkbcli-list | yq -r '.option_groups[] | length'`

Note that the top-level nodes have been de-capitalized, so where it was "Models"
before it is now "models" and the "Options" node is now "option_groups".

**Note:** this is a breaking change, but I'm not sure anyone cares. Plus, anyone who does care is up-to-date enough that they can switch their code around ;)

### DIFF
--- a/tools/registry-list.c
+++ b/tools/registry-list.c
@@ -136,11 +136,13 @@ main(int argc, char **argv)
         goto err;
     }
 
-    printf("Models:\n");
+    printf("models:\n");
     m = rxkb_model_first(ctx);
     assert(m); /* Empty model list is usually a bug or a bad xml file */
     while (m) {
-        printf("- %s:%s:%s\n",
+        printf("- name: %s\n"
+               "  vendor: %s\n"
+               "  description: %s\n",
                rxkb_model_get_name(m),
                rxkb_model_get_vendor(m),
                rxkb_model_get_description(m));
@@ -148,7 +150,7 @@ main(int argc, char **argv)
     }
 
     printf("\n");
-    printf("Layouts:\n");
+    printf("layouts:\n");
     l = rxkb_layout_first(ctx);
     assert(l); /* Empty layout list is usually a bug or a bad xml file */
     while (l) {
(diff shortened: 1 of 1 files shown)


tools/registry-list.c
```diff
@@ -136,76 +136,85 @@ main(int argc, char **argv)
...
-               rxkb_option_group_allows_multiple(g) ? "multiple" : "single");
+               rxkb_option_group_allows_multiple(g) ? "true" : "false");
 
         o = rxkb_option_first(g);
         assert(o); /* Empty option list is usually a bug or a bad xml file */
         while (o) {
             const char *brief = rxkb_option_get_brief(o);
 
-            printf("  - %s:%s:%s\n",
+            printf("  - name: '%s'\n"
```
bluetech, line 215: Here the values are quoted, but above they aren't. Is this intentional?

bluetech (commented):
Fine by me, just one question.

tools/registry-list.c
whot: yes, it is. Option names contain colons and without quotes that confuses YAML parsers, e.g. the unquoted yaml like this causes a complaint because it thinks it's a badly nested dictionary.

```yaml
 options:
  - name: esperanto:qwerty
    brief: ''
    description: 'At the corresponding key in a QWERTY layout'
```

OTOH we cannot use quotes for models and layouts because some of them include single quotes and we'd have to escape those - too much effort for what is a fairly tightly controlled set of entries anyway.

PR #240, opened 2021-04-26
Meson: Allow building as subproject

aperezdc:
Specify where to find the headers for `libxkbcommon_dep`, which allows other projects to correctly locate the headers when `libxkbcommon` is being built as a Meson subproject.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -253,8 +253,11 @@ install_headers(
     'xkbcommon/xkbcommon-names.h',
     subdir: 'xkbcommon',
 )
+
+# This variable may be used to build as a subproject and should not be renamed.
 libxkbcommon_dep = declare_dependency(
     link_with: libxkbcommon,
+    include_directories: include_directories('.'),
 )
 pkgconfig.generate(
     libxkbcommon,
@@ -310,8 +313,10 @@ You can disable X11 support with -Denable-x11=false.''')
         'xkbcommon/xkbcommon-x11.h',
         subdir: 'xkbcommon',
     )
+    # This variable may be used to build as a subproject and should not be renamed.
     libxkbcommon_x11_dep = declare_dependency(
         link_with: libxkbcommon_x11,
+        include_directories: include_directories('.'),
     )
     pkgconfig.generate(
         libxkbcommon_x11,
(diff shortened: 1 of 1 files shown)


aperezdc:
Currently we are carrying a modified `meson.build` file which includes this change [in libwpe](https://github.com/WebPlatformForEmbedded/libwpe/blob/master/subprojects/packagefiles/libxkbcommon-1.2.1/meson.build#L260), and it would be neat to have this upstream 😄

bluetech (commented):
Thanks, makes sense. Didn't know about the subproject feature.

Since our meson file now becomes an "API" of sort, could you add a comment above the variables whose name should remain stable, saying something like "This variable may be used from meson super-projects - try not to change the name.".

I wonder if we should move the includes into an `includes` directory, so that we don't need to specify the entire source directory here. But that can be done some other time.

xkbcommon also ships two other libraries - xkbcommon-x11 and xkbregistry. Presumably they need the same treatment. Can you add it for them as well?

aperezdc:
@bluetech Of course, I'll add the requested changes and update the PR. Thanks for considering taking this patch!

aperezdc:
All requested changes applied; I left out from this PR the idea of moving public API headers to a separate directory, probably it's better to do that separately.

bluetech (approved):

meson.build
```diff
@@ -253,8 +253,11 @@ install_headers(
     'xkbcommon/xkbcommon-names.h',
     subdir: 'xkbcommon',
 )
+
+# This variable may be used to build as a subproject and should not be renamed.
```
whot, line 257: fwiw, an approach I've used in another project (can't remember which one) is to have a specific section in the `meson.build` file to collect these "APIs", even with just simple aliases, i.e. around these lines:
```
# The below variables are considered API for subprojects
libxkbcommon_dep = dep_libxkbcommon
libxkbcommon_x11_dep =  dep_libxkbcommon_x11
```
The main benefit here is obviously to make it easier to keep these stable and have them in a single location.

whot (approved):

meson.build
bluetech: I like this idea, will do it as a follow up.

PR #241, opened 2021-04-26
Windows: Pass list of symbols to export to MSVC

aperezdc:
Arrange for passing `.def` files with the lists of symbols to export from DLLs when building on Windows with MSVC. Without this no symbols were being exported at all.

### DIFF
--- a/scripts/map-to-def
+++ b/scripts/map-to-def
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+"""A script to generate MSVC Module-Definition files from version-script
+files (which are maintained manually)."""
+
+import re
+import sys
+import pathlib
+
+
+def symbols_from_map(path):
+    return re.findall(r'^\s+(r?xkb_.*);', path.read_text('utf-8'), re.MULTILINE)
+
+
+if 2 > len(sys.argv) > 3:
+    raise SystemExit("Usage: {} file.map [file.def]".format(sys.argv[0]))
+
+
+map_file = pathlib.Path(sys.argv[1])
+map_symbols = set(symbols_from_map(map_file))
+
+if len(sys.argv) == 3:
+    def_file = open(sys.argv[2], "w", encoding="utf-8")
+else:
+    def_file = sys.stdout
+
+def_file.write("LIBRARY {}\n".format(map_file.stem))
+def_file.write("EXPORTS\n")
+for symbol in sorted(map_symbols):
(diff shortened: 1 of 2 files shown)


aperezdc:
I am not completely sure if it makes sense to do this for `libxkbcommon-x11` because... well, it's X11-specific. But I suppose it's not too far fetched anyway because one can run an X11 server on Windows, and there can be clients, too 🤔

bluetech (commented):
I wish meson would provide some abstraction for this but oh well.

We have a test `test/symbols-leak-test.py` which makes sure the files are kept up to date when new symbols are added. Can you update it to also check the new files? (I can do this myself after merging if you prefer)

bluetech:
Also:

- Let's keep the lists sorted
- Should use consistent indentation in all 3 files

aperezdc:
@bluetech I'll sort the lists and fix the indentation 👌🏼 

For reference, I did this to generate the list of functions:

```sh
sed -e '/^[[:space:]]*r\?xkb_[a-z0-9_]\+;$/{
s/;$//
p
}' -e d < xkbcommon.map >> xkbcommon.def
```

On a related note, `meson.source_root()` returns the root of the *topmost project*, so this does not work when libxkbcommon is a subproject—what we want is `meson.current_source_dir()`, but I am leaving that for a different PR 😉

aperezdc:
Now that I think of it... would it make sense to have a small Python script that generates the `.def` files from the `.map` ones?

bluetech:
If you can figure out the meson voodoo for that, sounds good.

aperezdc:
> If you can figure out the meson voodoo for that, sounds good.

Let's give it a try 😄

aperezdc:
@bluetech Done! In the convincing Meson to generate the `.def` files was not that difficult. This looks much nicer now, I think.

scripts/map-to-def
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+
```
bluetech, line 2:
```suggestion
"""A script to generate MSVC Module-Definition files from version-script
files (which are maintained manually)."""

```

bluetech (approved):
LGTM, just a couple small suggestions.

scripts/map-to-def
aperezdc: Suggestion applied, thanks!

whot (approved):

PR #242, opened 2021-04-27
atom: rewrite as a hash table

bluetech:
While the previous 1987-style[0] scheme was fun (and I reasonably
optimized it for a fair comparison), this task is more suited to a hash
table. Even a simple implementation beats the old one.

[0] Seems to have first appeared in X11R1, released September 1987.
    See server/dix/atom.c here: https://www.x.org/releases/X11R1/X.V11R1.tar.gz

### DIFF
--- a/bench/atom.c
+++ b/bench/atom.c
@@ -1,0 +1,93 @@
+/*
+ * Copyright © 2021 Ran Benita <ran@unusedvar.com>
+ *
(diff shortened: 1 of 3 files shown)


bluetech:
TODO: avoid `strndup` and `getline`, not supported by MSVC

PR #246, opened 2021-05-13
scripts/update-keysyms: fix path to the include files after de1b6943d

whot:


whot:
Merging this immediately because it's trivial and it's blocking xkeyboard-config's CI which relies on this script to test keymaps.
If required, please shout at me in the comments below :)

PR #247, opened 2021-05-22
ci: some tweaks

bluetech:
- Reduce permissions
- Update some stuff


bluetech:
Also changed the default permissions to read only.

PR #251, opened 2021-06-13
Meson: Handle absolute sysconfdir

concatime:
From doc[1], the default sysconfdir is 'etc', but when the prefix is
'/usr', it defaults to '/etc'. So we should prepend it with prefix only
when is relative. This also adds the possibility to have an odd prefix,
say '/opt/libxkbcommon', with '/etc' instead of '/opt/libxkbcommon/etc'
as sysconfdir. This approach is inspired from glib with
localstatedir[2].

[1] https://mesonbuild.com/Builtin-options.html#universal-options
[2] https://gitlab.gnome.org/GNOME/glib/-/blob/2.68.3/meson.build#L90

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -49,7 +49,11 @@ endif
 
 XKBCONFIGEXTRAPATH = get_option('xkb-config-extra-path')
 if XKBCONFIGEXTRAPATH == ''
-    XKBCONFIGEXTRAPATH = get_option('prefix')/get_option('sysconfdir')/'xkb'
+    sysconfdir = get_option('sysconfdir')
+    if not sysconfdir.startswith('/')
+        sysconfdir = get_option('prefix') / sysconfdir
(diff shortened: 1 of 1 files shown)


bluetech:
Hi @concatime, I don't understand when this would make a difference. Can you should example values for `prefix` and `sysconfdir`, and the resulting `DFLT_XKB_CONFIG_EXTRA_PATH` before and after?

concatime:
Wow, I wrongfully thought that, with prefix=`/opt` and sysconfdir=`/etc`, `'/opt' / '/etc'` would give `/opt//etc`, aka `/opt/etc`, but no, it gives `/etc` !
It’s documented [here](https://mesonbuild.com/Reference-manual.html#join_paths).
So forget my PR, sorry for the inconvenience.

PR #253, opened 2021-07-12
tools/interactive-x11: use keysym to look for Esc

emersion:
Don't assume that keycode 9 means Escape. Instead, use the keymap
to check for Esc.

Logic copied from the Wayland version.

Signed-off-by: Simon Ser <contact@emersion.fr>


bluetech (approved):
Thanks. Also checked interactive-wayland but it already gets it right.

Issue #252, opened 2021-07-02
xkbcli interactive-x11 crashes with (slightly malformed?) custom keymap


zackw:
I wrote an XKB keymap from scratch for my laptop, because I wanted total control over what each of the keys did.  I'm using [sway](https://github.com/swaywm/sway) as the primary window manager/compositor, and the keymap gives the behavior I want with native Wayland clients, but I still need to use a few X11 clients (notably emacs) and the keymap does not *quite* work as desired when fed to Xwayland.  While trying to troubleshoot this, I managed to get xkbcli interactive-x11 to crash on startup:

```
$ xkbcli interactive-x11
xkbcommon: ERROR: x11: failed to get keymap from X server: unmet condition in get_types(): wire_type->numLevels > 0
xkbcommon: ERROR: x11: failed to get keymap from X server: unmet condition in get_type_names(): type->num_levels == wire_num_levels
Segmentation fault (core dumped)
```

gdb tells me the crash is actually inside libxkbcommon-x11.so!xkb_x11_keymap_new_from_device.  I don't have debugging symbols to hand, unfortunately, so I can't be more specific.

Here is the [troublesome custom keymap](https://github.com/xkbcommon/libxkbcommon/files/6751442/troublesome-xkb.txt).  I would argue that _no matter how out of spec the keymap is_ libxkbcommon should not crash, but also I would appreciate any comments you may have on what might be wrong with this keymap.

bluetech:
A crash should definitely not happen. It's *probably* due to 1b3a1c277a033083fee669e92c8cad862716ebd1.

How do you load the keymap? I tried 

```
$ xkbcomp troublesome-xkb.txt $DISPLAY
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  135 (XKEYBOARD)
  Minor opcode of failed request:  9 (XkbSetMap)
  Value in failed request:  0x8010202
  Serial number of failed request:  27
  Current serial number in output stream:  32
```

but `xkbcli interactive-x11` doesn't crash after that.

zackw:
> How do you load the keymap?

With sway's `xkb_file` setting.  sway loads and parses the keymap, then supplies it to Xwayland using the Wayland keyboard protocol.  Xwayland is happy to load the keymap when it receives it this way, but clearly there's still something out-of-spec about it.

bluetech:
[...]

PR #258, opened 2021-08-31
x11: try to fix crash in xkb_x11_keymap_new_from_device error handling

bluetech:
In 1b3a1c277a033083fee669e92c8cad862716ebd1 we changed the error
handling in this code to not bail out immediately but only after
everything has been processed, to simplify the code. But I suspect the
code isn't prepared for this and that's what causing the crash reported
in the issue.

Bring back the short-circuit error handling which would hopefully fix
such crashes.

Fixes: https://github.com/xkbcommon/libxkbcommon/issues/252

### DIFF
--- a/src/x11/keymap.c
+++ b/src/x11/keymap.c
@@ -1160,6 +1160,10 @@ xkb_x11_keymap_new_from_device(struct xkb_context *ctx,
     struct x11_atom_interner interner;
     x11_atom_interner_init(&interner, ctx, conn);
 
+    /*
+     * Send all requests together so only one roundtrip is needed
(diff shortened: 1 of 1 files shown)


bluetech:
@psychon Would you mind taking a look to see if this looks OK to you (from what you remember)?

@zackw If you're able to test this to confirm it fixes the crash, that'd be great.

zackw:
I will not be able to test for a few days, the machine where I originally saw the crash needs hardware repairs and I'm waiting for the parts to arrive.  Hopefully soon though.

psychon (approved):

PR #260, opened 2021-09-27
Fix segmentation fault with wlroots/sway in xkb_keymap_num_layouts

TanapTheTimid:
xkb_keymap_num_layouts should be called only if keymap is not NULL, but for some reason wlroots and sway does call it with NULL after waking up from suspend on my hardware. This lets the function at least recover in edge cases by not accessing a null pointer and shouldn't affect well behaving applications.


bluetech:
xkbcommon intentionally does not check its arguments for NULL. In this case, returning 0 is not really appropriate in my opinion. I suggest capturing a back trace of the crash and reporting it to sway/wlroots, this is something they'll want to fix.

TanapTheTimid:
Alright I'll bring take issue to the sway guys. Thanks.

PR #263, opened 2021-10-08
build: add enable-tools option

Hello71:
this is mainly useful for multilib, but may be useful for other users as well

Signed-off-by: Alex Xu (Hello71) <alex_y_xu@yahoo.ca>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -125,8 +125,6 @@ elif cc.has_header_symbol('stdlib.h', '__secure_getenv', prefix: system_ext_defi
 else
     message('C library does not support secure_getenv, using getenv instead')
 endif
-have_getopt_long = cc.has_header_symbol('getopt.h', 'getopt_long',
-                                        prefix: '#define _GNU_SOURCE')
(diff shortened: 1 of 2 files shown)


whot (approved):

meson.build
```diff
@@ -406,6 +404,9 @@ endif
 man_pages = []
 
 # Tools
+have_getopt_long = cc.has_header_symbol('getopt.h', 'getopt_long',
+                                        prefix: '#define _GNU_SOURCE',
+                                        required: get_option('enable-tools'))
```
bluetech, line 409: I find this passing a `feature` to `required` business pretty confusing. I prefer not using that...

Also, we need to preserve the behavior that by default if `getopt_long` is available, the tools are built, regardless of meson's `auto-features` flag. So I'm thinking we shouldn't use `type: feature` for the option, but keep it a bool with default true.

Maybe some day we'll add proper `feature` options and deprecate the `enable` options, but until then, better to be consistent.

bluetech (commented):
Thanks, support for disabling building the tools would be nice.

I left a comment.

bluetech (approved):
Thanks!

Issue #264, opened 2021-10-10
Consider whether variants should inherit iso639 from parent 


hickford:
[libxkbregistry](https://xkbcommon.org/doc/current/group__registry.html) flattens [evdev.xml](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/master/rules/base.xml)'s parent-child relationship between layouts and variants. This is a useful convenience making it possible to handle layouts and variants the same way in a single loop.

> This library is the replacement for clients currently parsing evdev.xml directly. The library is intended to provide easy access to the set of possible MLVO configurations for a given ruleset. It is not a library to apply these configurations, merely to enumerate them. The intended users of this library are the configuration UIs that allow a user to select their keyboard layout of choice.

However because xkeyboardconfig expects ["if the variant record does not have own languageList, the parent layout's languages are used"](https://www.freedesktop.org/wiki/Software/XKeyboardConfig/Rules/), it's sometimes necessary to lookup iso639 on the parent layout, when omitted on the variant. This means the libxkbregistry user has to reconstruct the tree structure. It's no longer so easy to handle all the layouts and variants in a single loop without saving state between iterations and extra lookups.

Thus consider inheriting iso639 (and similarly iso3166) from parent layout to child variant: 

Example from base.xml:

    <layout>
      <configItem>
        <name>at</name>
        <description>German (Austria)</description>
        <languageList>
          <iso639Id>deu</iso639Id>
        </languageList>
      </configItem>
      <variantList>
        <variant>
          <configItem>
            <name>nodeadkeys</name>
            <description>German (Austria, no dead keys)</description>
          </configItem>
        </variant>

Corresponding output from `xkbcli list`, iso639 highlighted:

> - layout: at
>   variant: ''
>   brief: de
>   description: German (Austria)
>   **iso639: [ger]**
>   iso3166: []
> - layout: at
>   variant: nodeadkeys
>   brief: ''
>   description: German (Austria, no dead keys)
>   **iso639: []**   
>   iso3166: []

Here iso639 for variant 'at(nodeadkeys)' would thus be `[ger]` inherited from parent layout 'at'.

* * *

By the same reasoning, the 'brief' field should also be inherited: https://www.freedesktop.org/wiki/Software/XKeyboardConfig/Rules/

> If some variant does not provide own short description, the short description from the parent layout is expected to be used

hickford:
[...]

PR #266, opened 2021-10-24
Variants should inherit iso639, iso3166 and brief from parent layout

hickford:
Fixes https://github.com/xkbcommon/libxkbcommon/issues/264

### DIFF
--- a/src/registry.c
+++ b/src/registry.c
@@ -860,7 +860,8 @@ parse_variant(struct rxkb_context *ctx, struct rxkb_layout *l,
             v->name = strdup(l->name);
             v->variant = name;
             v->description = description;
-            v->brief = brief;
+            // if variant omits brief, inherit from parent layout.
+            v->brief = brief == NULL ? strdup_safe(l->brief) : brief;
             v->popularity = popularity;
             list_append(&ctx->layouts, &v->base.link);
 
@@ -870,11 +871,35 @@ parse_variant(struct rxkb_context *ctx, struct rxkb_layout *l,
                 if (!is_node(ci, "configItem"))
                     continue;
 
+                bool found_language_list = false;
+                bool found_country_list = false;
                 for (node = ci->children; node; node = node->next) {
-                    if (is_node(node, "languageList"))
+                    if (is_node(node, "languageList")) {
                         parse_language_list(node, v);
-                    if (is_node(node, "countryList"))
+                        found_language_list = true;
+                    }
+                    if (is_node(node, "countryList")) {
                         parse_country_list(node, v);
+                        found_country_list = true;
+                    }
+                }
+                if (!found_language_list) {
+                    // inherit from parent layout
+                    struct rxkb_iso639_code* x;
+                    list_for_each(x, &l->iso639s, base.link) {
+                        struct rxkb_iso639_code* code = rxkb_iso639_code_create(&v->base);
+                        code->code = strdup(x->code);
+                        list_append(&v->iso639s, &code->base.link);
+                    }
+                }
+                if (!found_country_list) {
+                    // inherit from parent layout
+                    struct rxkb_iso3166_code* x;
+                    list_for_each(x, &l->iso3166s, base.link) {
+                        struct rxkb_iso3166_code* code = rxkb_iso3166_code_create(&v->base);
+                        code->code = strdup(x->code);
+                        list_append(&v->iso3166s, &code->base.link);
+                    }
                 }
             }
         } else {
--- a/test/registry.c
+++ b/test/registry.c
@@ -599,7 +599,7 @@ test_load_full(void)
     rxkb_layout_unref(l);
 
     l = fetch_layout(ctx, "l1", "v2");
-    struct test_layout expected = {"l1", "v2", NULL, "vdesc2"};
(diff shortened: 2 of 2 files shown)


hickford:
Thanks. Any idea why a test fails on macos only? I didn't write any platform-specific code. Does anyone have access to a macos machine to investigate further?

hickford:
Aha. Now we can read the long lines. Could this be the problem?

> 'xkbregistry: ERROR: I/O warning : failed to load external entity "/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp2gb8gcnq/rules/evdev.xml"\nxkbregistry: ERROR: I/O warning : failed to load external entity "/tmp//rules/evdev.xml"\n')

hickford:
Here a test points to evdev.xml https://github.com/xkbcommon/libxkbcommon/blob/8531ea723fde41062a2cda7a5b75a01716e30a78/test/xkeyboard-config-test.py.in#L13

```
DEFAULT_RULES_XML = '@XKB_CONFIG_ROOT@/rules/evdev.xml'
```

hickford:
Aha! I tried three different things in https://github.com/xkbcommon/libxkbcommon/pull/266/commits/5bf2cc8734819e8157df65e6d32b30b2724b068d , and something fixed the issue. Will figure out which.

hickford:
Wahey! An extra `strdup_safe` fixed the issue. Presumably macos has some local evdev.xml with empty elements or something.

src/registry.c
```diff
@@ -870,11 +871,35 @@ parse_variant(struct rxkb_context *ctx, struct rxkb_layout *l,
...
                         parse_country_list(node, v);
+                        found_country_list = true;
+                    }
+                }
+                if (!found_language_list) {
+                    // inherit from parent layout
+                    struct rxkb_iso639_code* x;
+                    list_for_each(x, &l->iso639s, base.link) {
+                        struct rxkb_iso639_code* code = rxkb_iso639_code_create(&v->base);
+                        code->code = strdup_safe(x->code);
```
bluetech, line 891: It does not make a lot of sense to me that `x->code` can be NULL - in such a case it should not exist in the first place. E.g. I don't expect users of `rxkb_iso639_code_get_code` to handle NULL (and it's not documented as being able to return NULL).

Were you able to figure out which file caused this?

bluetech (commented):
Sorry for the delay @hickford. I left a question regarding the `strdup_safe` change.

src/registry.c
hickford: No. Not even sure where the "I/O warning : failed to load external entity" error is thrown.

https://github.com/xkbcommon/libxkbcommon/runs/4612955170?check_suite_focus=true

hickford: https://github.com/xkbcommon/libxkbcommon/blob/master/src/registry.c

> libxml2 ... My enthusiasm about this is indescribable

Interestingly from the 'install dependencies' step https://github.com/xkbcommon/libxkbcommon/runs/4612955170?check_suite_focus=true

> libxml2 is keg-only, which means it was not symlinked into /usr/local,
because macOS already provides this software and installing another version in
parallel can cause all kinds of trouble.

Maybe macosx is using an older version of libxml2 which loads external entities by default: https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html#libxml2


> starting with libxml2 version 2.9, XXE has been disabled by default

hickford: A suspicion: Is xmlCleanupParser called early?

bluetech:
@whot Do you have an opinion on the `strdup_safe`'s? Is NULL supposed to be possible for these - e.g. `rxkb_iso639_code_get_code`?

whot:
Haven't paged this all back in but I'd say no - we shouldn't have a iso639 object in that case at all

src/registry.c
whot: `"/tmp//rules/evdev.xml"` has  a double-slash suggests something that may have been a `$foo` is missing. and it starts with `/tmp` in the tool-options test, so it's likely related to the env vars in `test/tool-option-parsing.py:main`. Not 100% sure on the rest yet, this was just from a quick peek

whot: ok, had another look and that warning is a red herring, I think, and it doesn't even matter for this particular test. The stdout of the tool seems fine, stderr has that warning but that warning happens *before* the stdout so it's not the reason we abort.

I've filed #277 now to at least print the exit status to get a hint what's going on but I haven't been able to reproduce this myself in the CI. I'm half-wondering if it's just a side-effect of #275 though it really shouldn't be.

hickford: `strdup` instead of `strdup_safe` works as you suggest.

hickford:
Aha. strdup instead of strdup_safe works as you suggest, including on Mac.

bluetech (approved):
This LGTM now, thanks @hickford.

I'll leave merging to @whot.

whot (approved):

whot:
merged (with squash), thanks heaps!

whot:
I've added another test in #279 for the ISO code inheritance, easier to make it a separate PR than making you re-spin this one. Please give it a quick skim when you find the time, thanks.

PR #269, opened 2021-11-08
tools: enclose the layout name in quotes

### LINKED ISSUE #268: Norway problem in yaml output from xkbcli list
> Anyone who uses YAML long enough will eventually get burned when attempting to abbreviate Norway.

https://news.ycombinator.com/item?id=17359376

YAML output from `xkbcli list`

```yaml
- layout: no
  variant: ''
  brief: no
  description: Norwegian
  iso639: [nor, nob, nno]
  iso3166: []
```

Output of PyYAML's `yaml.safe_load`

```python
{'layout': False,
 'variant': '',
 'brief': False,
 'description': 'Norwegian',
 'iso639': ['nor', 'nob', 'nno'],
 'iso3166': []},
```

whot:
Because otherwise the 'no' layout is treated as disagreement with whatever is to
be disagreed with. Fixed in YAML 1.2 but that's not universally supported.

Fixes #268

### DIFF
--- a/tools/registry-list.c
+++ b/tools/registry-list.c
@@ -140,11 +140,12 @@ main(int argc, char **argv)
     m = rxkb_model_first(ctx);
     assert(m); /* Empty model list is usually a bug or a bad xml file */
     while (m) {
+        const char *vendor = rxkb_model_get_vendor(m);
         printf("- name: %s\n"
                "  vendor: %s\n"
                "  description: %s\n",
                rxkb_model_get_name(m),
-               rxkb_model_get_vendor(m),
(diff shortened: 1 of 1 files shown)


whot:
Actually, come to think of it we should probably enclose the variants too since we're bound to run into the same issue there too. And the ISO bits, etc... I don't remember exactly why I didn't choose to add quotes everywhere, I think it was just for readability.

whot:
ready now, this encloses most short strings that have the potential of resulting in a mis-parsing in quotes now

bluetech (approved):

PR #276, opened 2022-01-20
registry: skip over invalid ISO639 or ISO3166 entries

whot:
If the XML file is somehow off, don't load entries that are against the spec.

Note that the new tests will fail once #266 works since they don't handle merging yet.

### DIFF
--- a/test/registry.c
+++ b/test/registry.c
@@ -56,6 +56,8 @@ struct test_layout {
     const char *variant;
     const char *brief;
     const char *description;
+    const char *iso639[3];  /* language list (iso639 three letter codes), 3 is enough for our test  */
+    const char *iso3166[3]; /* country list (iso3166 two letter codes), 3 is enough for our tests */
 };
 
 struct test_option {
@@ -76,7 +78,9 @@ fprint_config_item(FILE *fp,
                    const char *name,
                    const char *vendor,
                    const char *brief,
-                   const char *description)
+                   const char *description,
+                   const char * const iso639[3],
+                   const char * const iso3166[3])
 {
     fprintf(fp, "  <configItem>\n"
                 "    <name>%s</name>\n", name);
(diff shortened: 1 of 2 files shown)


test/registry.c
```diff
@@ -56,6 +56,8 @@ struct test_layout {
     const char *variant;
     const char *brief;
     const char *description;
+    const char *iso639[3];
```
bluetech, line 59: I suggest some comment here like

> Language list (ISO-639 three-letter codes). Can be any length but we only need up to 3 in tests.

and maybe also change the name to `language_list` or `iso639s` to indicate this is a list not a single item.

(Same for `iso3166`).

test/registry.c
```diff
@@ -458,6 +485,42 @@ cmp_layouts(struct test_layout *tl, struct rxkb_layout *l)
     if (!streq_null(tl->description, rxkb_layout_get_description(l)))
         return false;
 
+    iso3166 = rxkb_layout_get_iso3166_first(l);
+    for (size_t i = 0; i < sizeof(tl->iso3166); i++) {
+        const char *iso = tl->iso3166[i];
+        if (iso == NULL && iso3166 == NULL)
+            break;
+
+        if ((iso && iso3166 == NULL) || (iso == NULL && iso3166))
```
bluetech, line 494: We have `streq_null` for this, if you want to use it.

test/registry.c
```diff
@@ -608,6 +671,87 @@ test_load_full(void)
...
+    struct rxkb_layout *l;
+    struct rxkb_iso3166_code *iso3166;
+    struct rxkb_iso639_code *iso639;
+
+    ctx = test_setup_context(system_models, NULL,
+                             system_layouts, NULL,
+                             system_groups, NULL);
+
+    l = fetch_layout(ctx, "l1", NO_VARIANT);
+    /* uvz is invalid, we expect 3 letters, verify it was ignored */
```
bluetech, line 741:
```suggestion
    /* uvw is invalid, we expect 2 letters, verify it was ignored */
```

test/registry.c
```diff
@@ -608,6 +671,87 @@ test_load_full(void)
...
+    iso3166 = rxkb_layout_get_iso3166_first(l);
+    assert(streq(rxkb_iso3166_code_get_code(iso3166), "xz"));
+    assert(rxkb_iso3166_code_next(iso3166) == NULL);
+
+    /* ab is invalid, we expect 2 letters, verify it was ignored */
```
bluetech, line 746:
```suggestion
    /* ab is invalid, we expect 3 letters, verify it was ignored */
```

bluetech (approved):
LGTM with comments

whot:
fixed and merged, thanks!

PR #279, opened 2022-01-24
test: add test for variant inheritance of ISO 639/3166 codes

whot:
See efa99624320f54e6e1ff24e4c4b1fcecdfba43d5

cc @hickford

### DIFF
--- a/test/registry.c
+++ b/test/registry.c
@@ -685,6 +685,12 @@ test_load_languages(void)
         {"l1", "v1", "vbrief1", "vdesc1",
             .iso639 = {"efg"},
             .iso3166 = {"yz"}},
+        {"l2", NO_VARIANT, "lbrief1", "ldesc1",
+            .iso639 = { "hij", "klm" },
+            .iso3166 = { "op", "qr" }},
+        {"l2", "v2", "lbrief1", "ldesc1",
+            .iso639 = { NULL }, /* inherit from parent */
+            .iso3166 = { NULL }},  /* inherit from parent */
         {NULL},
     };
     struct test_option_group system_groups[] = {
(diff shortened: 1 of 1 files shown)


test/registry.c
```diff
@@ -685,6 +685,12 @@ test_load_languages(void)
         {"l1", "v1", "vbrief1", "vdesc1",
             .iso639 = {"efg"},
             .iso3166 = {"yz"}},
+        {"l2", NO_VARIANT, "lbrief1", "ldesc1",
+            .iso639 = { "hij", "klm" },
+            .iso3166 = { "op", "qr" }},
+        {"l2", "v2", "lbrief1", "ldesc1",
+            .iso639 = { NULL }, /* inherit from parent */
```
hickford, line 692: New to C ... does this describe an empty list or a list that's absent? For the test XML will this write an empty countryList element or no countryList element?

hickford, line 692: New to C ... does this describe an empty list or a list that's null? In the test XML, will there be an empty languageList element, or no languageList element at all?

In efa99624320f54e6e1ff24e4c4b1fcecdfba43d5 I intended to inherit only if the languageList element is omitted. Otherwise it would be impossible for a variant to specify 'no languages regardless of parent'.

hickford (changes requested):

test/registry.c
whot: In C missing field initializers are set to zero [1]. so in this case, it's a struct of size 3 and only the first is defined, so the others are set to zero (or `NULL` which is just a fancy type for zero). Leaving this line out completely would also work but I put it in to highlight that it's explicitly not set.

`fprint_config_item()` only prints out `<languageList>` if `iso639[0]` is nonzero, so in this case we'd end up with no element. Which I think is what you want?

[1] if at least one other initializer is specified, i.e. `struct foo bar;` is still random memory.

hickford: Is it possible to test both absent languageList (expect iso639 to inherit) and empty languageList (expect iso639 to be empty)?

whot: not with the code as-is but adding a boolean for `use_empty_language_list` or something and the matching change in `fprint_config_item()` could work.

bluetech (approved):

Issue #270, opened 2021-12-17
Misleading parameter description in xkb_keymap_key_get_mods_for_level()


maxrd2:
https://github.com/xkbcommon/libxkbcommon/blob/8531ea723fde41062a2cda7a5b75a01716e30a78/include/xkbcommon/xkbcommon.h#L1204

I believe this should be changed to "The number of elements in the buffer pointed by masks_out".
Current description seems like size in bytes which will do buffer overflow.

Implementation is here: https://github.com/xkbcommon/libxkbcommon/blob/8531ea723fde41062a2cda7a5b75a01716e30a78/src/keymap.c#L413

```c++
XKB_EXPORT size_t
xkb_keymap_key_get_mods_for_level(struct xkb_keymap *keymap,
                                  xkb_keycode_t kc,
                                  xkb_layout_index_t layout,
                                  xkb_level_index_t level,
                                  xkb_mod_mask_t *masks_out,
                                  size_t masks_size)
{
  ...
        for (unsigned i = 0; i < type->num_entries && count < masks_size; i++)
  ...
            masks_out[count++] = 0;
  ...
    for (unsigned i = 0; i < type->num_entries && count < masks_size; i++) {
  ...
            masks_out[count++] = type->entries[i].mods.mask;
  ...
}
```

whot:
Can you file a pull request for this please? Thanks

maxrd2:
@whot closing as there is PR now... sorry for the delay :)

PR #281, opened 2022-02-24
Improve misleading comments #270

maxrd2:
Small improvement... more details are in the issue #270


bluetech (approved):
Thanks! This is indeed misleading.

PR #284, opened 2022-03-13
meson: add enable-tests option

cazfi:
Add option to disable building of the tests.

This is a patch that I've been carrying locally for some time. Originally implemented to work around some build failure on tests that I don't need anyway, but might be more generally useful.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -531,6 +531,9 @@ configure_file(input: 'test/xkeyboard-config-test.py.in',
                configuration: xkct_config)
 
 # Tests
+
+if get_option('enable-tests')
+
 test_env = environment()
 test_env.set('XKB_LOG_LEVEL', 'debug')
 test_env.set('XKB_LOG_VERBOSITY', '10')
@@ -767,6 +770,7 @@ if get_option('enable-x11')
   )
 endif
 
+endif
 
 # Documentation.
 if get_option('enable-docs')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -50,6 +50,12 @@ option(
     description: 'Enable building tools',
 )
 option(
+    'enable-tests',
+    type: 'boolean',
+    value: true,
+    description: 'Enable building of the tests',
+)
+option(
     'enable-x11',
     type: 'boolean',
     value: true,


whot:
I don't think we have any specific dependencies that are only for tests, so the tests should build wherever the main library builds.
If that's not the case it may be better fixing the build failures rather than just disabling the tests outright.

cazfi:
On Mon, 14 Mar 2022 at 00:55, Peter Hutterer ***@***.***>
wrote:

> I don't think we have any specific dependencies that are only for tests,
> so the tests should build wherever the main library builds.
> If that's not the case it may be better fixing the build failures rather
> than just disabling the tests outright.
> Message ID: ***@***.***


 In principle I agree. I ever used (and presumably keep on using) that
option just as a quick workaround. My use-case is cross-compiling to MinGW
target, and quick test with libxkbcommon without that option says:

../../src/libxkbcommon-1.4.0/test/registry.c: In function
‘test_create_rules’:
../../src/libxkbcommon-1.4.0/test/registry.c:137:12: warning: implicit
declaration of function ‘mkdtemp’; did you mean ‘mkstemp’?
[-Wimplicit-function-declaration]
  137 |     assert(mkdtemp(tmpdir) == tmpdir);
      |            ^~~~~~~
../../src/libxkbcommon-1.4.0/test/registry.c:137:12: warning: nested extern
declaration of ‘mkdtemp’ [-Wnested-externs]
../../src/libxkbcommon-1.4.0/test/registry.c:137:28: warning: comparison
between pointer and integer
  137 |     assert(mkdtemp(tmpdir) == tmpdir);
      |                            ^~
../../src/libxkbcommon-1.4.0/test/registry.c:141:10: error: too many
arguments to function ‘mkdir’
  141 |     rc = mkdir(buf, 0777);
      |          ^~~~~
In file included from /usr/share/mingw-w64/include/unistd.h:10,
                 from ../../src/libxkbcommon-1.4.0/test/registry.c:28:
/usr/share/mingw-w64/include/io.h:282:15: note: declared here
  282 |   int __cdecl mkdir (const char *)
__MINGW_ATTRIB_DEPRECATED_MSVC2005;
      |               ^~~~~

eli-schwartz:
That comes from definitions in https://github.com/Alexpux/mingw-w64/blob/d0d7f784833bbb0b2d279310ddc6afb52fe47a46/mingw-w64-headers/crt/io.h#L276-L285 which provide strange and deprecated MSVC crt functions that don't really correspond to the POSIX functions Microsoft originally designed them to incorrectly implement.

bluetech:
I committed #301 which should hopefully fix your compilation problems (thanks @eli-schwartz for the hint). Let me know if it doesn't.

PR #285, opened 2022-03-14
Rename eof() as s_eof() to avoid conflicts with mingw-w64 eof()

cazfi:
Another patch that I have had locally for some time.

I'm now guessing that I've meant 's_' -prefix to mean 'static' as the function is local to modules using it.


bluetech:
Thanks for the patch. I went with a different solution - prefix all scanner functions with `scanner_`, for consistency. It makes the code uglier but that's OK.

PR #288, opened 2022-04-27
Remove bogus euro sign entry from keysymtab

emersion:
Not sure what it's doing here, but converting "€" to a keysym
doesn't work with this entry. 0x13a4 doesn't appear in
xkbcommon-keysyms.h. 0x20ac is the keysym documented in the
header (and it's the last entry in the table).

It's been in the table since it was introduced
in e0524296d2e0 ("Add API for getting unicode representation of a
keysym").


bluetech (approved):
Thanks, this looks correct.

The 0x13a4 value probably stems from the value of Euro in [Latin-9](https://en.wikipedia.org/wiki/ISO/IEC_8859-15), but ended up not being used in favor of the Unicode value.

There are a few more cases of this, I will fix them separately.

PR #291, opened 2022-06-09
Add support for modmap None.

### LINKED ISSUE #211: "modifier_map None" is not supported
Using Xorg setxkbmap & xkbcomp tools it is possible to clear a `modifier_map` of a key using `modifier_map None`.

However, using `xkbcli compile-keymap` on a custom layout containing `modifier_map None { Caps_Lock };` I get:

```
xkbcommon: ERROR: Illegal modifier map definition; Ignoring map for non-modifier "None"
xkbcommon: ERROR: Failed to compile xkb_symbols
xkbcommon: ERROR: Failed to compile keymap
```

* setxkbmap 1.3.2
* xkbcomp 1.4.4
* xkbcli 1.0.3

wismill:
Fixes #211

### DIFF
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -109,6 +109,9 @@
 /* Don't allow more leds than we can hold in xkb_led_mask_t. */
 #define XKB_MAX_LEDS ((xkb_led_index_t) (sizeof(xkb_led_mask_t) * 8))
 
+/* Special value to handle modMap None {…} */
+#define XKB_MOD_NONE 0xffffffffU
+
 /* These should all go away. */
 enum mod_type {
     MOD_REAL = (1 << 0),
--- a/src/xkbcomp/symbols.c
+++ b/src/xkbcomp/symbols.c
@@ -162,6 +162,8 @@ ClearKeyInfo(KeyInfo *keyi)
 typedef struct {
     enum merge_mode merge;
     bool haveSymbol;
+    // NOTE: Can also be XKB_MOD_NONE, meaning
+    //       “don’t add a modifier to the modmap”.
     xkb_mod_index_t modifier;
     union {
         xkb_atom_t keyName;
@@ -1152,14 +1154,21 @@ HandleModMapDef(SymbolsInfo *info, ModMapDef *def)
     xkb_mod_index_t ndx;
     bool ok;
     struct xkb_context *ctx = info->ctx;
+    const char *modifier_name = xkb_atom_text(ctx, def->modifier);
 
-    ndx = XkbModNameToIndex(&info->mods, def->modifier, MOD_REAL);
-    if (ndx == XKB_MOD_INVALID) {
-        log_err(info->ctx,
-                "Illegal modifier map definition; "
-                "Ignoring map for non-modifier \"%s\"\n",
-                xkb_atom_text(ctx, def->modifier));
-        return false;
+    if (istreq(modifier_name, "none")) {
+        // Handle special "None" entry
+        ndx = XKB_MOD_NONE;
+    } else {
+        // Handle normal entry
+        ndx = XkbModNameToIndex(&info->mods, def->modifier, MOD_REAL);
+        if (ndx == XKB_MOD_INVALID) {
+            log_err(info->ctx,
+                    "Illegal modifier map definition; "
+                    "Ignoring map for non-modifier \"%s\"\n",
+                    xkb_atom_text(ctx, def->modifier));
+            return false;
+        }
     }
 
     ok = true;
@@ -1523,7 +1532,12 @@ CopyModMapDefToKeymap(struct xkb_keymap *keymap, SymbolsInfo *info,
         }
     }
 
-    key->modmap |= (1u << entry->modifier);
+    // Handle modMap None
+    if (entry->modifier != XKB_MOD_NONE) {
+        // Convert modifier index to modifier mask
+        key->modmap |= (1u << entry->modifier);
+    }
+
     return true;
 }
 
--- a/test/data/keymaps/modmap-none.xkb
+++ b/test/data/keymaps/modmap-none.xkb
@@ -1,0 +1,148 @@
+xkb_keymap {
+xkb_keycodes "test" {
+    minimum = 8;
+    maximum = 255;
+	<LVL3>  = 92;
+    <LFSH>  = 50;
+	<RTSH>  = 62;
+	<LALT>  = 64;
+	<RALT>  = 108;
+	<LWIN>  = 133;
+	<RWIN>  = 134;
+	<LCTL>  = 37;
+	<RCTL>  = 105;
+	<CAPS>  = 66;
+
+	<AD01> = 24;
+	<AD02> = 25;
+	<AD03> = 26;
+	<AD04> = 27;
+	<AD05> = 28;
+	<AD06> = 29;
+	<AD07> = 30;
+	<AD08> = 31;
+	<AD09> = 32;
+};
+
+xkb_types "complete" {
+    type "ONE_LEVEL" {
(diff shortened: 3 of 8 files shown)


wismill:
@bluetech could you guide me to write the corresponding tests?

wismill:
@bluetech @whot kind reminder for a review.

whot:
ETIME on the actual code but please add the information from #211 to the commit message so it's not 3 clicks away from anyone trying to figure out why this was added.

src/darray.h
```diff
@@ -143,6 +143,13 @@ typedef darray (unsigned long)  darray_ulong;
     (arr).item[--(arr).size] = 0; \
 } while (0)
 
+#define darray_remove(arr, i) do { \
+	size_t index_ = (i); \
+	if (index_ < arr.size-1)    \
```
whot, line 148: style nitpick:
```suggestion
	if (index_ < arr.size-1) \
```

also, `arr.size` and `size_t` are both unsigned so removing any index from a zero-sized array will proceed without a warning (or rather: it's not a noop as expected). Might be better to use `ssize_t` here to check for `<0` as well.

I strongly recommend splitting the `darray_remove` addition into a separate patch and adding test cases for it so we don't have to worry about this. There's nothing existing that I can find, but a `test/darray.c` won't hurt :)

src/darray.h
```diff
@@ -143,6 +143,13 @@ typedef darray (unsigned long)  darray_ulong;
...
+		memmove(&(arr).item[index_], &(arr).item[index_+1], ((arr).size-1-index_)*sizeof(*(arr).item)); \
```
whot, line 149: style nitpick
```suggestion
		memmove(&(arr).item[index_], &(arr).item[index_+1], ((arr).size-1 - index_)*sizeof(*(arr).item)); \
```

src/xkbcomp/symbols.c
```diff
@@ -462,6 +463,12 @@ AddModMapEntry(SymbolsInfo *info, ModMapEntry *new)
         use = (clobber ? new->modifier : old->modifier);
         ignore = (clobber ? old->modifier : new->modifier);
 
+        // Remove entry
```
whot, line 466: This file only uses `/* */`-style comments

whot (changes requested):
A few style fixes and the darray bit needs fixing, but other than that this LGTM (though I'm not well-versed in the internals of the actual modifier mappings)

src/darray.h
wismill: Done

src/xkbcomp/symbols.c
wismill: Done

src/darray.h
wismill: Done

whot (approved):
thanks, this LGTM

wismill:
> thanks, this LGTM

@whot I think we should add a test for the new feature. Can you give me  some guidance?

wismill:
Kind reminder

wismill:
@whot @bluetech What is blocking this PR? Kind reminder.

bluetech (commented):
I've read the xkbcomp code, and after much sighing over the wonders of 90s style C code, and after finally figuring out that the entire thing relies on the fact that `1U << entry->modifier` ends up being zero when `entry->modifier` is `XkbNoModifier` (i.e. 0xFF), I think I understand it.

So here are my review comments:

1. No reason to expose `XKB_NO_MODIFIER` in the public header.

3. In fact, will be better not to use `XKB_NO_MODIFIER` at all, since it's in-band data that is prone to error. Instead, should just add a flag to `ModMapEntry` saying "this is no-modifier/clearing entry". Luckily this stuff is entirely encapsulated in xkbcommon so it's not a problem.

5. The mechanism by which `None` works in xkbcomp is that the `None` entry overrides conflicting entries using the usual merge mechanism, and ends up being a no-op in `CopyModMapDefToKeymap`. This PR instead makes `None` physically remove conflicting entries (and not adding a "no-modifier" entry). I think the xkbcomp approach is preferable, because:
   1. It preserves compat with xkbcomp.
   2. It preserve the usual merging semantics which is what is expected.
   3. It's simpler (no need for `darray_remove` which is better avoided in general).

WDYT?

wismill:
@bluetech Thanks for the code archaeology! Now I see how to implement it.

I followed your suggestion and added an extensive comment in the commit to explain the behaviour in X11’s xkbcomp and why we do not use their trick. Please review that I got the correct understanding!

I reverted the previous commits, rebased and added some tests.

src/xkbcomp/symbols.c
```diff
@@ -1152,14 +1152,21 @@ HandleModMapDef(SymbolsInfo *info, ModMapDef *def)
...
-        log_err(info->ctx,
-                "Illegal modifier map definition; "
-                "Ignoring map for non-modifier \"%s\"\n",
-                xkb_atom_text(ctx, def->modifier));
-        return false;
+    xkb_atom_t none = xkb_atom_intern(ctx, "None", 4);
+
+    if (def->modifier == none) {
+        // Handle special "None" entry
+        ndx = XKB_MOD_NONE;
```
bluetech, line 1159: Let's add a comment on the ModMapDef `modifier` field in the struct definition, `// Can also be XKB_MOD_NONE, meaning "don't add a modifier to the modmap".`

bluetech (commented):
Thanks for the update, LGTM with just a couple of small comments.

src/xkbcomp/symbols.c
wismill: I think you meant ModMapEntry. I added a comment to both.

test/data/keymaps/modmap-none.xkb
```diff
@@ -0,0 +1,93 @@
...
+    key <LFSH> { [Shift_L] };
+    modmap Shift { <LFSH> };
+    modmap None  { <LFSH> };
+
+    // Reset modmap with one symbol
+    key <RTSH> { [Shift_R] };
+    modmap Shift { Shift_R };
+    modmap None  { Shift_R };
+
+    // Cycle
```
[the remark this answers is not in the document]
wismill: I added some. Do you have other in mind?

src/xkbcomp/symbols.c
```diff
@@ -1152,14 +1152,21 @@ HandleModMapDef(SymbolsInfo *info, ModMapDef *def)
...
-    ndx = XkbModNameToIndex(&info->mods, def->modifier, MOD_REAL);
-    if (ndx == XKB_MOD_INVALID) {
-        log_err(info->ctx,
-                "Illegal modifier map definition; "
-                "Ignoring map for non-modifier \"%s\"\n",
-                xkb_atom_text(ctx, def->modifier));
-        return false;
+    xkb_atom_t none = xkb_atom_intern(ctx, "None", 4);
+
+    if (def->modifier == none) {
```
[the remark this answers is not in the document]
wismill: Fixed

src/keymap.h
```diff
@@ -109,6 +109,9 @@
 /* Don't allow more leds than we can hold in xkb_led_mask_t. */
 #define XKB_MAX_LEDS ((xkb_led_index_t) (sizeof(xkb_led_mask_t) * 8))
 
+/* Special value to handle modMap None {…} */
+#define XKB_MOD_NONE 0xffffffff
```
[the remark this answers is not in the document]
wismill: Fixed

test/data/keymaps/modmap-none.xkb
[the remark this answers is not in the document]
bluetech: No, the test is great.

bluetech (approved):

PR #294, opened 2022-07-15
ci: store the logs on failure and split test and valgrind-test

whot:
Let's keep the test logs so we can figure *why* a test failed. And since valgrind tends to spam the logs, let's run a normal test run first to catch any actual test case failures that aren't valgrind-related.


bluetech (approved):
Makes sense, thanks.

PR #295, opened 2022-07-15
registry: translate briefs and descriptions

### LINKED ISSUE #293: rxkb_layout_get_description does not have i18n
It seems that the strings read from `xkeyboard-config` are not being translated. I think they should be handled similarly to `libxklavier`. Please see the usage of `dgettext` and `XKB_DOMAIN` in https://github.com/freedesktop/libxklavier/blob/65c132a65e90a42e898f07243ef544109ada53c9/libxklavier/xklavier_config.c

whot:
Filing this as Draft for now because I reckon this is an API break. It may be be better to add `rxkb_*_get_description_i18n()` or something. As it is the returned string changes, so if you're e.g. libgnome-desktop you then pump that again into `gettext()` and get the already-translated string translated. This could lead to breakage.

Alternatively, if we don't want a new API we could make this a flag at context creation.

Fixes #293

cc @bluetech, @tari01

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -26,7 +26,11 @@ jobs:
           sudo apt update
           sudo apt install -y \
             doxygen libxcb-xkb-dev valgrind ninja-build \
-            libwayland-dev wayland-protocols bison graphviz
+            libwayland-dev wayland-protocols bison graphviz \
+            xkb-data
+      - name: Install languages for i18n test
+        run: |
+          sudo apt install -y language-pack-de language-pack-es language-pack-fr
       - name: Setup
         run: |
           meson setup build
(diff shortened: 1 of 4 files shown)


sunweaver:
Thanks for working on this. @tari01, can you test this? Thanks.

whot:
I'm going to close this MR, I don't think it's worth the effort. In the end, the actual code dealing with i18n here is a single call: `dgettext("xkeyboard-config", value)`. But doing this in libxkbcommon is likely to trigger bugs since we potentially double-translate now. So we need a new API, all for what is just a single call anyway.

We should probably document this though.

I also suspect that a good chunk of clients needing translations already have code in place since previously they probably parsed the evdev.xml file directly. Like GNOME (and why I never noticed the lack of translations).

PR #298, opened 2022-09-10
Add an identifier for when the evdev code is 2

n3rdopolis:
usually meaning key repeat on some keyboards


bluetech:
I'm assuming you mean the `input_event`'s `value` field, not the `code` field.

The way key-repeats are should be handled in xkbcommon is e.g. DOWN, DOWN, DOWN, UP. There is no need for e.g. DOWN, HELD, HELD, UP in the xkbcommon state machine.

n3rdopolis:
OK, Yeah, I meant value. I figured it might have been useful for filtering out if a modifier key is being held, and it's one of those keyboards that generate multiple modifier presses, but if this is invalid, this can be closed

bluetech:
OK, so I'll close this. There is no reason to add a value to the enum if it's not used by `xkb_state_update_key`.

PR #312, opened 2022-12-15
context: add XKB_CONTEXT_NO_SECURE_GETENV flag

### LINKED ISSUE #308: Make a XKB_CONTEXT_YES_DEFAULT_INCLUDES flag?
Hi there.

I'm working on sway, which uses CAP_SYS_NICE to gain realtime priority. This has the unfortunate side-effect of borking `xkb_context_include_path_append_default` because secure_getenv refuses to read $HOME, $XDG_CONFIG_HOME, etc. and this is a hassle to avoid.

I would be interested in maybe a XKB_CONTEXT_YES_DEFAULT_INCLUDES flag (opposite of XKB_CONTEXT_NO_DEFAULT_INCLUDES) that would add these default configs using getenv on request. Sway doesn't support suid operation so I believe there would be no security risk to create our xkb context with such a flag.

Related: swaywm/sway#7311

RPigott:
This flag is useful for clients that may have relatively benign capabilities set, like CAP_SYS_NICE, that also want to use the xkb configuration from the environment and user configs in XDG_CONFIG_HOME.

Fixes #308

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -585,8 +585,7 @@ enum xkb_context_flags {
     XKB_CONTEXT_NO_ENVIRONMENT_NAMES = (1 << 1),
     /**
      * Disable the use of secure_getenv for this context, so that privileged
-     * processes can use XKB configuration environment variables. Client uses at
-     * their own risk.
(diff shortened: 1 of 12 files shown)


bluetech:
Thanks. I made some edits and verified that it works as intended, so it's good to go.

Issue #310, opened 2022-12-10
suggestion: allow xkb_state_update_mask() to accept negative groups to unlock group wrapping functionality


bam80:
https://github.com/xkbcommon/libxkbcommon/blob/f60bdb1680650a131e8e21ffa4a8a16775a35c9f/src/state.c#L792-L798
`xkb_layout_index_t` is unsigned int which prevents us from passing a negative group to utilize internal group wrapping:
https://github.com/xkbcommon/libxkbcommon/blob/f60bdb1680650a131e8e21ffa4a8a16775a35c9f/src/state.c#L192-L195
This forces the client to re-implement that wrapping on it's side, which is unnecessary.

Could we just declare the `*_group` arguments as `int32_t`?

bluetech:
I'm curious where negative values come from exactly?

bam80:
The calling code could pass just `current_group - 1` to switch to previous layout.

bam80:
Right now, we have to do the wrapping ourselves here in KWin, which duplicates the code above:
https://lxr.kde.org/source/plasma/kwin/src/xkb.cpp#0606

bluetech:
I see, thanks.

As you can see from the `XkbWrapGroupIntoRange` code you linked to, XKB has a concept of "Out of range group action", which determines what happens when the group index goes of range. It can be configured per-key using the following values in the XKB symbols file:

- groupswrap - wrap
- groupsclamp - saturate, i.e. stop at zero if underflow and at last group if overflow
- groupsredirect=<groupnumber> - set to <groupnumber>

Like many things in XKB, I'm pretty sure this functionality is entirely unused, but it's supported by libxkbcommon and you can never know what configs people have.

So question is, when switching to prev/next in KWin UI, do you want the "Out of range group action" to be consulted, or always wrap?

bluetech:
Right, I was confusing the "global" active layout (which always wraps as we don't support the global "XKB Controls" like X11 does), and the per-key derived layout. So that's fine.

The next problem is that `xkb_state_update_mask` is not really meant to be used in the way KWin uses it in `XKB::switchToLayout`. As the `xkb_state_update_mask` docstring says:

```
 * This entry point is intended for window systems and the like, where a
 * master process holds an xkb_state, then serializes it over a wire
 * protocol, and clients then use the serialization to feed in to their own
 * xkb_state.
 *
 * All parameters must always be passed, or the resulting state may be
 * incoherent.
 *
[...]

PR #314, opened 2022-12-17
state: add API for updating latched and locked mods & layout in server state

bluetech:
Up to now, the "server state" xkb_state API only offered one entry point to update the server state - `xkb_state_update_key`, which reflects the direct keyboard keys state. But some updates come out-of-band from keyboard input events stream, for example, a GUI layout switcher.

The X11 XKB protocol has a request which allows for such updates, [`XkbLatchLockState`], but xkbcommon does not have similar functionality. So server applications ended up using `xkb_state_update_state` for this, but that's a function intended for client applications, not servers.

Add support for updating the latched & locked state of the mods and layout. Note that the depressed states cannot be updated in this way -- XKB does not expect them to be updated out of band.

[`XkbLatchLockState`]: https://www.x.org/releases/X11R7.7/doc/kbproto/xkbproto.html#Querying_and_Changing_Keyboard_State

Fixes: https://github.com/xkbcommon/libxkbcommon/issues/310

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -81,9 +81,10 @@
 #ifndef _XKBCOMMON_H_
 #define _XKBCOMMON_H_
 
+#include <stdarg.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
-#include <stdarg.h>
 
 #include <xkbcommon/xkbcommon-names.h>
 #include <xkbcommon/xkbcommon-keysyms.h>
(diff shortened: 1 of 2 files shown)


bluetech:
Design note: libX11 takes the `XkbLatchLockState` request and splits it to four functions - `XkbLockGroup`, `XkbLockModifiers`, `XkbLatchGroup`, `XkbLatchModifiers`. This is cleaner than the monster function, but I didn't do it here, because it makes it impossible to do an atomic update while getting the best `enum xkb_state_component` return mask.

bluetech:
The latched part will need some work, might drop it if I don't find the time to implement it.

bam80:
> So server applications ended up using xkb_state_update_state for this, but that's a function intended for client applications, not servers.

Could you elaborate what are negative consequences of this, if it already works? Won't we just add a syntactic sugar expanding the API this way?

wismill:
I added a fix for the latch group action in #483.

wismill:
Closing in favor of #521

PR #315, opened 2023-02-18
build: override dependency for use as subproject

emersion:
This allows xkbcommon to be used as a subproject.


bluetech (commented):
Looks OK to me as far as I understand [it](https://mesonbuild.com/Reference-manual_builtin_meson.html#mesonoverride_dependency).

xkbcommon also includes two other libraries, xkbcommon-x11 and xkbregistry. Should this be done for them as well?

emersion:
Yeah, done!

bluetech (approved):
Thanks!

PR #318, opened 2023-04-08
Fallback to Compose file in locale directory if no mapping exists in compose.dir

JSkold:
I've created and distribute a custom locale that a mix between my native language's locale and en_US. There unfortunately seems to be no way for me to extend the `compose.dir` file to include a mapping for my new locale.  

This MR add a new fallback in case a `compose.dir` entry is missing for the current locale and will cause libxkbcommon to use the compose file at:
`/usr/share/X11/locale/en_XX.UTF-8/Compose`
In this case en_XX is missing from `compose.dir`.

### DIFF
--- a/src/compose/paths.c
+++ b/src/compose/paths.c
@@ -200,8 +200,9 @@ get_locale_compose_file_path(struct xkb_context *ctx, const char *locale)
         locale = "en_US.UTF-8";
 
     resolved = resolve_name(ctx, "compose.dir", RIGHT_TO_LEFT, locale);
-    if (!resolved)
-        return NULL;
+    if (!resolved) {
+        return asprintf_safe("%s/%s/Compose", get_xlocaledir_path(ctx), locale);
+    }
 
     if (resolved[0] == '/') {
         path = resolved;


bluetech:
Do you also create a custom Compose file for your locale?

bluetech:
While having a custom locale is probably rare, having a custom locale *and* a custom Compose is surely very rare. Most people will want to fallback to the `en_US.UTF-8` Compose.

Current status quo is that if you're using a custom locale, you have no solution.

With this PR, if you're using a custom locale, Compose is still broken, but you can fix with an additional step of adding an `/usr/share/X11/locale/en_XX.UTF-8/Compose` file with the magic `include` incantation. This seems not much better to me than adding this include line to your `~/.XCompose` or setting `XCOMPOSEFILE` (both of which take precedence over the locale).

I think the preferable state would be that, if you're using a custom locale, you fall back to `en_US.UTF-8`, with (optionally) some provision to provide a custom file. However, this will need to be considered carefully, as doing it naively means `xkb_compose_table_from_locale` will no longer fail on bad/misspelled locales, which is not great.

JSkold:
In my case fallback to `en_US.UTF-8` would work, but I don't like to assume that is what everyone wants. My hope is that any locale should just be able to be installed like any other package.

The benefit of checking for a fallback is that the administrator does not need to make additional configuration and the locale would work just like any other. A package would ship with two files:
`/usr/share/i18n/locales/en_XX`  
`/usr/share/X11/locale/en_XX.UTF-8/Compose`  
and it would "Just Work". In the other case the administrator would have to either configure every user to have a specific `~/.XCompose` or put `XCOMPOSEFILE` in something like `/etc/environment`. It works but it's more unnecessary steps for the administrator to take.

PR #319, opened 2023-04-11
keysyms: add new keysyms XF86EmojiPicker, XF86Dictate

bluetech:
Ref: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/commit/914d8f5e0f469cd0416364dd008e9eea752bf703
Ref: https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/commit/a839f0c7fc5596d10e786394d3b0953eb8a1731b


whot:
Adding the two new syms looks good, I'm assuming the rest is autogenerated and thus keeps working :)

bluetech:
Yep it's generated by scripts/update-keysyms

Issue #305, opened 2022-10-31
crosscompile xkbcommon on linux for windows (1/2)


awarnke:
Hello,

crosscompiling xkbcommon on linux (OpenSuse 15.4) for windows works, when applying some changes:

1) #ifdef _MSC_VER does not work when using the gnu cross compiler. I can fix this by setting this variable:

        meson configure -Dbuild.c_args=-D_MSC_VER=1

I guess this is just a hack, the right way would be to use another compiler variable. I'm no expert here - maybe _WIN32 is the right one, as used in other libraries:

src/fontconfig-2.13.96/src/fcxml.c:#ifdef _WIN32

src/libjpeg-turbo-2.1.3/tjexample.c:#ifdef _WIN32

src/expat-2.4.4/lib/xmlparse.c:#ifdef _WIN32

2) There is another issue with the tests, I guess I best create another ticket for this.

Just a note: in meson.build, 

elif cc.get_argument_syntax() == 'msvc'

does also not work, the cross compiler might have a different signature.

PS: Some infos on setting up cross-compile of libxkbcommon for win are documented here:
https://github.com/awarnke/crystal-facet-uml/tree/master/build/win

Kind Regards
Andreas

bluetech:
> I'm no expert here - maybe _WIN32 is the right one, as used in other libraries:

No expert either, but this sounds OK to me, we can make this change.

> elif cc.get_argument_syntax() == 'msvc' does also not work, the cross compiler might have a different signature.

Hmm this one does seem correct to me, here we do actually care about the specific compiler type.

What problem did this cause?

bluetech:
The _MSC_VER -> _WIN32 change is done in #320, will be included in the next release.

PR #320, opened 2023-04-11
Possible fix for non-MSVC windows compilers

bluetech:
`_MSC_VER` is specific to MSVC, but there can be other compilers targeting windows. Hopefully they do define `_WIN32`, so let's use that.

Refs: https://github.com/xkbcommon/libxkbcommon/issues/305

### DIFF
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -30,7 +30,7 @@
 #include "bench.h"
 #include "../src/utils.h"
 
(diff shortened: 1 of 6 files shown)


awarnke:
The changes look good to me (even if of course _WIN32 sounds strange to me in times of 64 bit architectures).
Thank you for the fix!

PR #322, opened 2023-04-30
Enabled fine-grained warnings for parser

wismill:
I am trying to make it easier to spot errors in `xkeyboard-config`.

I think fine-grained could help us with this:
- Allow to enable/disable single warnings.
- WIP: Allow to convert warnings to errors. It does not seem easy to implement this feature.
- Add new parser warning for numeric keysyms (except 0-9).

@bluetech @whot

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,3 +12,7 @@ max_line_length = 80
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.map]
+indent_style = tab
+indent_size = 4
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -650,6 +650,116 @@ xkb_context_set_user_data(struct xkb_context *context, void *user_data);
 void *
 xkb_context_get_user_data(struct xkb_context *context);
 
+/**
+ * Control the enabled warnings.
+ *
+ * @since 1.6.0
+**/
+enum xkb_warning_flags {
+    /** No warning */
+    XKB_WARNING_NONE = 0,
+    /** Warn on unrecognized keysyms */
+    XKB_WARNING_UNRECOGNIZED_KEYSYM = (1 << 1),
+    /** Warn on numeric keysym (other than 0-9) */
+    XKB_WARNING_NUMERIC_KEYSYM = (1 << 2),
+    /** Warn if there are conflicting key types while merging groups. */
+    XKB_WARNING_CONFLICTING_KEY_TYPE = (1 << 3),
+    /** Warn if there are conflicting actions while merging keys. */
+    XKB_WARNING_CONFLICTING_KEY_ACTION = (1 << 4),
+    /** Warn if there are conflicting keysyms while merging keys. */
+    XKB_WARNING_CONFLICTING_KEY_KEYSYM = (1 << 5),
+    /** Warn if there are conflicting fields while merging keys. */
+    XKB_WARNING_CONFLICTING_KEY_SYMBOL_MAP_FIELDS = (1 << 6),
+    /** Warn if there are conflicting modmap definitions. */
+    XKB_WARNING_CONFLICTING_MODMAP = (1 << 7),
+    /** Warn if a group name was defined for group other than the first one. */
+    XKB_WARNING_GROUP_NAME_FOR_NON_BASE_GROUP = (1 << 8),
+    /** Warn if a key defines multiple groups at once. */
+    XKB_WARNING_MULTIPLE_GROUP_AT_ONCE = (1 << 9),
+    /** Warn if no key type can be inferred. */
+    XKB_WARNING_CANNOT_INFER_KEY_TYPE = (1 << 10),
+    /** Warn if using an undefined key type. */
+    XKB_WARNING_UNDEFINED_KEY_TYPE = (1 << 11),
+    /** Warn if using a symbol not defined in the keymap. */
+    XKB_WARNING_UNRESOLVED_KEYMAP_SYMBOL = (1 << 12),
+    // TODO:
+    XKB_WARNING_EXTRA_SYMBOLS_IGNORED = (1 << 13),
+    // TODO:
+    XKB_WARNING_UNDEFINED_KEYCODE = (1 << 14),
+    // TODO: deprecated keysym
+    // TODO: unicode keysym when named and recommended keysym exists
+    /** Default warnings */
+    XKB_WARNING_DEFAULT = ( \
+        XKB_WARNING_UNRECOGNIZED_KEYSYM | \
+        XKB_WARNING_CONFLICTING_KEY_TYPE | \
+        XKB_WARNING_CONFLICTING_KEY_ACTION | \
+        XKB_WARNING_CONFLICTING_KEY_KEYSYM | \
+        XKB_WARNING_CONFLICTING_KEY_SYMBOL_MAP_FIELDS | \
+        XKB_WARNING_CONFLICTING_MODMAP | \
+        XKB_WARNING_GROUP_NAME_FOR_NON_BASE_GROUP | \
+        XKB_WARNING_MULTIPLE_GROUP_AT_ONCE | \
+        XKB_WARNING_CANNOT_INFER_KEY_TYPE | \
+        XKB_WARNING_UNDEFINED_KEY_TYPE | \
+        XKB_WARNING_UNRESOLVED_KEYMAP_SYMBOL | \
+        XKB_WARNING_EXTRA_SYMBOLS_IGNORED | \
+        XKB_WARNING_UNDEFINED_KEYCODE \
+    ),
+    /** All the previous warnings */
+    XKB_WARNING_ALL = ( \
+        XKB_WARNING_DEFAULT | \
+        XKB_WARNING_NUMERIC_KEYSYM \
+    ),
+};
+
+/** Default errors */
+#define XKB_ERROR_DEFAULT 0
+
+/**
+ * Store warning flags in the context.
+ *
+ * @memberof xkb_context
+ *
+ * @since 1.6.0
+**/
+void
+xkb_context_set_warning_flags(struct xkb_context *context,
+                              enum xkb_warning_flags warning_flags);
+
+/**
+* Retrieves the warning flags from the context.
+ *
+ * @returns The current warning flags.
+ *
+ * @memberof xkb_context
+ *
+ * @since 1.6.0
+*/
+enum xkb_warning_flags
+xkb_context_get_warning_flags(struct xkb_context *context);
+
+/**
+ * Store error flags in the context.
+ *
+ * @memberof xkb_context
+ *
+ * @since 1.6.0
+**/
+void
+xkb_context_set_error_flags(struct xkb_context *context,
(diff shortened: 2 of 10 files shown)


xkbcommon.map
```diff
@@ -9,6 +9,8 @@ global:
 	xkb_context_unref;
 	xkb_context_set_user_data;
 	xkb_context_get_user_data;
+    xkb_context_set_warning_flags;
+    xkb_context_get_warning_flags;
```
whot, line 13: this looks like an indentation issue

tools/compile-keymap.c
```diff
@@ -114,29 +138,59 @@ parse_options(int argc, char **argv, struct xkb_rule_names *names)
...
+        {"Wno-multiple-group-at-once",            no_argument, &warnings, -XKB_WARNING_MULTIPLE_GROUP_AT_ONCE},
+        {"Wcannot-infer-key-type",                no_argument, &warnings, XKB_WARNING_CANNOT_INFER_KEY_TYPE},
+        {"Wno-cannot-infer-key-type",             no_argument, &warnings, -XKB_WARNING_CANNOT_INFER_KEY_TYPE},
+        {"Wundefined-key-type",                   no_argument, &warnings, XKB_WARNING_UNDEFINED_KEY_TYPE},
+        {"Wno-undefined-key-type",                no_argument, &warnings, -XKB_WARNING_UNDEFINED_KEY_TYPE},
+        {"Wunresolved-keymap-symbol",             no_argument, &warnings, XKB_WARNING_UNRESOLVED_KEYMAP_SYMBOL},
+        {"Wno-unresolved-keymap-symbol",          no_argument, &warnings, -XKB_WARNING_UNRESOLVED_KEYMAP_SYMBOL},
+        {"Wall",                                  no_argument, &warnings, XKB_WARNING_ALL},
+        {"Wnone",                                 no_argument, &warnings, -XKB_WARNING_ALL},
+        {"Werror",                                no_argument, &warnings, XKB_WARNING_IS_ERROR},
```
whot, line 186: `Werror` is a large hammer, I personally prefer what e.g. `flake8` does where there's `W` and `E` for each warning. I can imagine something like `-Wconflicting-modmap -Eunrecognized-keysym`, it'd be nice if we could control it this way.

tools/compile-keymap.c
```diff
@@ -114,29 +138,59 @@ parse_options(int argc, char **argv, struct xkb_rule_names *names)
...
+        {"from-xkb",         no_argument,            NULL, OPT_FROM_XKB},
+        {"include",          required_argument,      NULL, OPT_INCLUDE},
+        {"include-defaults", no_argument,            NULL, OPT_INCLUDE_DEFAULTS},
+        {"rules",            required_argument,      NULL, OPT_RULES},
+        {"model",            required_argument,      NULL, OPT_MODEL},
+        {"layout",           required_argument,      NULL, OPT_LAYOUT},
+        {"variant",          required_argument,      NULL, OPT_VARIANT},
+        {"options",          required_argument,      NULL, OPT_OPTION},
+        // Warnings
+        {"Wunrecognized-keysym",                  no_argument, &warnings, XKB_WARNING_UNRECOGNIZED_KEYSYM},
```
whot, line 160: this is a long option, so it'd be `--Wunrecognized-keysym` - unlike the gcc option. That seems a bit odd, given that the format is supposed to resemble the compiler options.

Instead, you could do it like `c = getopt_long(argc, argv, "hW:", opts, &option_index);` and the `optarg` to `'W'` is the flag. You'll have to strcmp instead of having it in this option then but I don't think that's a huge issue. Plus it'd be easy to add the `E` handling (see below) then with the same code if you factor it out into a helper function.

whot:
Thought bubble: can those warnings be standardised so the they can be *parsed* easily? I don't want to say JSON or YAML because I'm not sure whether that's easily possible but maybe we can enforce some format that then allows for filtering them externally. Most of the time we don't have control over the context (e.g. when looking at logs), so being able to do some form of `grep "XKB" | yq` would be nice.

tools/compile-keymap.c
wismill: Yes, still toying with it. I think I just have to add an `error_flags` field similar to the new `warning_flags`.

This would only affect the logging part atm though. I am not sure if it is easy to change the exit status to reflect those errors. The exit status could be a mask of the errors flags.

tools/compile-keymap.c
wismill: Fixed

tools/compile-keymap.c
wismill: I implemented the `error_flags` idea.

include/xkbcommon/xkbcommon.h
```diff
@@ -650,6 +650,116 @@ xkb_context_set_user_data(struct xkb_context *context, void *user_data);
...
+    ),
+    /** All the previous warnings */
+    XKB_WARNING_ALL = ( \
+        XKB_WARNING_DEFAULT | \
+        XKB_WARNING_NUMERIC_KEYSYM \
+    ),
+};
+
+/** Default errors */
+#define XKB_ERROR_DEFAULT 0
```
whot, line 715:
```suggestion
#define XKB_ERROR_DEFAULT XKB_WARNING_NONE
```
for correctness, it's not like C cares much about types :)

include/xkbcommon/xkbcommon.h
```diff
@@ -650,6 +650,116 @@ xkb_context_set_user_data(struct xkb_context *context, void *user_data);
...
+
+/**
+ * Store warning flags in the context.
```
whot, line 718: needs docs that any flags here will unset the equivalent error flags, so in the strictest sense `-Efoo -Wfoo` downgrades `foo` to a warning.  Which isn't quite what I'd expect tbh.

include/xkbcommon/xkbcommon.h
```diff
@@ -650,6 +650,116 @@ xkb_context_set_user_data(struct xkb_context *context, void *user_data);
...
+ *
+ * @memberof xkb_context
+ *
+ * @since 1.6.0
+*/
+enum xkb_warning_flags
+xkb_context_get_warning_flags(struct xkb_context *context);
+
+/**
+ * Store error flags in the context.
```
whot, line 741: needs docs that any flags here will unset the equivalent warning flags.

src/context.c
```diff
@@ -367,3 +369,35 @@ xkb_context_set_user_data(struct xkb_context *ctx, void *user_data)
...
+        return ctx->warning_flags;
+    return -1;
+}
+
+XKB_EXPORT void
+xkb_context_set_warning_flags(struct xkb_context *ctx,
+                              enum xkb_warning_flags warning_flags)
+{
+    ctx->warning_flags = warning_flags;
+    ctx->error_flags &= ~warning_flags;
```
whot, line 386: tbh, I don't think this bit is a good idea - adding a warning flag shouldn't remove an error flag. It's fine for warning upgraded to error but downgrading is not good. Errors default to 0 anyway so it's easier to handle that case.

.editorconfig
```diff
@@ -12,3 +12,7 @@ max_line_length = 80
 [*.yml]
 indent_style = space
 indent_size = 2
+
+[*.map]
+indent_style = tab
+indent_size = 4
```
whot, line 18: I'd prefer that as separate commit please, we could merge that immediately.

src/xkbcomp/symbols.c
```diff
@@ -61,6 +61,30 @@
 #include "include.h"
 #include "keysym.h"
 
+
+// TODO: how to convert warnings to program errors?
```
whot, line 65: you can't, not in the library. It's up to the caller to detect a log call with `XKB_LOG_LEVEL_ERROR` and convert that to a program error.

Though the problem with that is that you can't differ between a generic error that occurred (e.g. "cant' find file") and a parser error that you specifically enabled.

whot (changes requested):
Code looks good, a few nitpicks but otherwise fine.

whot:
> Thought bubble: can those warnings be standardised so the they can be _parsed_ easily? I don't want to say JSON or YAML because I'm not sure whether that's easily possible but maybe we can enforce some format that then allows for filtering them externally. Most of the time we don't have control over the context (e.g. when looking at logs), so being able to do some form of `grep "XKB" | yq` would be nice.

After reviewing the current patch, I have a possible solution to this:
```diff
diff --git a/src/xkbcomp/symbols.c b/src/xkbcomp/symbols.c
index f990529..8521ed1 100644
--- a/src/xkbcomp/symbols.c
+++ b/src/xkbcomp/symbols.c
@@ -242,6 +242,7 @@ MergeGroups(SymbolsInfo *info, GroupInfo *into, GroupInfo *from, bool clobber,
 
             if (report)
                 log_warn(info->ctx,
+                         "conflicting-key-type:"
                          "Multiple definitions for group %d type of key %s; "
                          "Using %s, ignoring %s\n",
                          group + 1, KeyNameText(info->ctx, key_name),
```
Now imagine this being an array/`#define` like your `WarningLookupEntry` instead of a string and bam, you get logs that can be grepped for errors. And the decision on what to treat as error is up to the caller.

wismill:
> Now imagine this being an array/`#define` like your `WarningLookupEntry` instead of a string and bam, you get logs that can be grepped for errors. And the decision on what to treat as error is up to the caller.

Let's see if I understand you:

Instead of having flag to select warning and their optional promotions to error, you propose to use e.g. 

```c
// index of a lookup table, no longer exposed
#define XKB_WARNING_CONFLICTING_KEYTYPE 0
// Some lookup table with first entry:
"conflicting-key-type"
```
Then change the log functions to use a standard format, e.g.: 

```
xkbcommon: <LOG LEVEL> [<warning id>] <warning message>
```

```c
log_warn(ctx, XKB_WARNING_CONFLICTING_KEYTYPE, ...);
```

The log level is fixed by the log function caller and not parametrable from CLI.

Then it is up to the user wishing to detect some warnings to grep the logs.

If so, most of this PR is useless. Also, how can the user keep up to date with the warning ids? Maybe we need a command that output all of them, then if not present the user knowns something has changed.

whot:
Yep, good summary, that's the proposal.

> If so, most of this PR is useless

-ish. they're mostly complementary and address different use-cases. In your case and afaict, you want to be able  to tweak the warnings to easily spot the errors in the `compile-keymap` subcommand - which is useful for the xkeyboard-config CI tests for example to reduce the noise. This same use-case could be solved by standardising the warnings and grepping the logs afterwards.

The second case is to find those warnings in existing systems, e.g. when mutter (or whatever) runs libxkbcommon and the logs end up somewhere, maybe the journal. Grepping would work for that case, but the warning flags don't - not unless whatever calls libxkbcommon exposes those flags to be configurable.

The third case is just general better hygiene, e.g. ` No symbols defined for <foo>` should be default disabled but enabled for debugging by helper tools. Grepping can't solve that issue, but your PR can.

> Also, how can the user keep up to date with the warning ids? Maybe we need a command that output all of them, then if not present the user knowns something has changed.

Those prefixes would become part of the public API, possibly as `#define` strings, but... not 100% thought that bit through yet :)

wismill:
I am going to close this MR and start a new one on a better base.

PR #324, opened 2023-05-02
registry: two old fixes leftover in my branch

whot:
Found these two 2y old fixes on a branch during spring (well, autumn) clean. Figured I might as well push them :)

### DIFF
--- a/src/registry.c
+++ b/src/registry.c
@@ -594,7 +594,7 @@ XKB_EXPORT bool
 rxkb_context_include_path_append_default(struct rxkb_context *ctx)
 {
     const char *home, *xdg, *root, *extra;
-    char *user_path;
+    char user_path[PATH_MAX];
     bool ret = false;
 
     if (ctx->context_state != CONTEXT_NEW) {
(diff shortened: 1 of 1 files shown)


bluetech (approved):
LGTM.

I remember we got rid of PATH_MAX in the past because it doesn't exist in Gnu Hurd and is kind of a lie anyway? (commit ec9a02a239d42b24a897e985c7cd3c3833d68f33). But I see we got them back since then, and no complaints, so probably fine.

whot:
> I remember we got rid of PATH_MAX in the past because it doesn't exist in Gnu Hurd and is kind of a lie anyway? 

heh, tbh I think getting rid of `PATH_MAX` in userspace is quite a windmill to fight... But we do have a `#define` in `meson.build` for it where it's missing - see 4d0d509129e5f80c4a9efe4523fc4b0aea58622b. I don't think I'v ever seen this as anything but 4096 though (or, apparently 260 on windows).

bluetech:
Hah I remember now :)

PR #325, opened 2023-05-02
scanner: allow for a zero terminated string as keymap

### LINKED ISSUE #307: xkb_keymap_new_from_buffer does not allow terminating null bytes
If a  null-terminated buffer is given to `xkb_keymap_new_from_buffer()`, our keymap fails to parse:
```
xkbcommon: ERROR: (input string):1928:3: unrecognized token
xkbcommon: ERROR: (input string):1928:3: syntax error
xkbcommon: ERROR: Failed to parse input xkb string
```
Fairly easy to verify by modifying our existing keymap compile test to include the NULL-byte.

```diff
diff --git a/test/buffercomp.c b/test/buffercomp.c
index 12b67fe..88a369d 100644
--- a/test/buffercomp.c
+++ b/test/buffercomp.c
@@ -66,9 +66,15 @@ main(int argc, char *argv[])
         assert(0);
     }
 
+    free(dump);
+    xkb_keymap_unref(keymap);
+
+    /* Load a prebuild keymap but include the trailing NULL byte */
+    keymap = test_compile_buffer(ctx, original, strlen(original) + 1);
+    assert(keymap);
+    xkb_keymap_unref(keymap);
+
     free(original);
-    free(dump);
-    xkb_keymap_unref(keymap);
 
     /* Make sure we can't (falsely claim to) compile an empty string. */
     keymap = test_compile_buffer(ctx, "", 0);
```
This seems needlessly pedantic :) 

In our case the keymap comes from a different process (the compositor) so we don't control if the null-byte is included or not, it'd be nice to just have this handled correctly.

whot:
As the documentation for `xkb_keymap_new_from_string()` states, the "input string does not have to be zero-terminated". The actual implementation however failed with "unrecognized token/syntax error" when it encountered a null byte.

Fix this by allowing a null byte at the last position of the buffer. Anything else is likely a client error anyway.

Fixes #307

cc @ofourdan

### DIFF
--- a/src/keymap.c
+++ b/src/keymap.c
@@ -194,6 +194,10 @@ xkb_keymap_new_from_buffer(struct xkb_context *ctx,
     if (!keymap)
         return NULL;
 
+    /* Allow a zero-terminated string as a buffer */
+    if (length > 0 && buffer[length - 1] == '\0')
+        length--;
+
     if (!ops->keymap_new_from_string(keymap, buffer, length)) {
         xkb_keymap_unref(keymap);
         return NULL;
--- a/test/buffercomp.c
+++ b/test/buffercomp.c
@@ -46,29 +46,33 @@ main(int argc, char *argv[])
     original = test_read_file(DATA_PATH);
     assert(original);
 
-    keymap = test_compile_buffer(ctx, original, strlen(original));
-    assert(keymap);
+    /* Load a prebuild keymap, once without, once with the trailing \0 */
+    for (int i = 0; i <= 1; i++) {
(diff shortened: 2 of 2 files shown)


bluetech (commented):
The commit message refers to `xkb_keymap_new_from_string` but should say `xkb_keymap_new_from_buffer`.

IIRC, these scanner checks are quite performance sensitive. While this one probably isn't, it still might be better to just add special handling of this in `xkb_keymap_new_from_buffer`? i.e. `if (len > 0 && buf[len-1] == '\0') len--;`

bluetech (approved):

Issue #327, opened 2023-05-03
executables have "@rpath" linkages to libraries


kencu:
Installing libxkbcommon results in the installed executables being linked to libxkbcommon.0.dylib using an '@rpath' linkage to the library:

```
% otool -L /opt/local/{bin/xkbcli,lib/libxkbcommon.dylib}
/opt/local/bin/xkbcli:
	@rpath/libxkbcommon.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
/opt/local/lib/libxkbcommon.dylib:
	/opt/local/lib/libxkbcommon.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.0.0)
% xkbcli 
dyld[24238]: Library not loaded: '@rpath/libxkbcommon.0.dylib'
  Referenced from: '/opt/local/bin/xkbcli'
  Reason: tried: '/usr/local/lib/libxkbcommon.0.dylib' (no such file), '/usr/lib/libxkbcommon.0.dylib' (no such file)
zsh: abort      xkbcli
```

One way this can be fixed is by telling meson that the executables link against libxkbcommon.0.dylib like this:

```
 % cat *.diff
--- meson.build.orig	2023-05-02 18:42:33
+++ meson.build	2023-05-02 18:46:02
@@ -414,6 +414,7 @@
     tools_dep = declare_dependency(
         include_directories: [include_directories('tools', 'include')],
         link_with: libxkbcommon_tools_internal,
+        dependencies: dep_libxkbcommon,
     )
 
     executable('xkbcli', 'tools/xkbcli.c',
```

and then:

```
% otool -L /opt/local/{bin/xkbcli,lib/libxkbcommon.dylib} 
/opt/local/bin/xkbcli:
	/opt/local/lib/libxkbcommon.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
/opt/local/lib/libxkbcommon.dylib:
	/opt/local/lib/libxkbcommon.0.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)


 % xkbcli
[...]

PR #328, opened 2023-05-03
meson.build: register libxkbcommon.dylib link

kencu:
meson needs to know that the executable tools link against libxkbcommon.dylib so that the
@rpath references used during the build/test phases can be rewritten to full path names on install

Fixes: #327


whot:
merged, thanks

PR #330, opened 2023-05-04
.editorconfig: add .md, .css and .map configs

wismill:
Separate simple patch, as asked in #322

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,14 @@ indent_style = space
 indent_size = 4
 max_line_length = 80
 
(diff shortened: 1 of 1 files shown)


bluetech (commented):
Thanks!

I noticed `*.map` says to use tabs, but we're not consistent on that -- `xkbregistry.map` doesn't use tabs. Since tabs suck, I think it will be better to switch to 4-space.

Also for CSS 8-space seems excessive, let's go for 4 there as well.

I'll do these changes and take this commit directly to master.

PR #332, opened 2023-05-04
Structured log messages with  unique identifiers

wismill:
Add unique identifier to log messages in order to
make it easier to spot specific warning & errors.

This is especially relevant for `xkeyboard-config`.

Follow-up of #322 

@whot


wismill:
Some thoughts:
- The log message IDs should not be exposed in the public API for now. This is experimental and incomplete. I will change that.
- I think it would be good to design log messages with the perspective of an error code index. Some examples: [Rust](https://doc.rust-lang.org/error_codes/), [Haskell](https://errors.haskell.org/).
- Maybe a numeric code is better: it would provide more concise messages and avoid repetition. Allow it will keep messages aligned in the terminal.
- We should provide some automatic way for tools pattern matching the logs that their patterns are up to date.

wismill:
@whot @bluetech I implemented a first version of an error index:
- Warnings and errors are defined in the YAML file `doc/error-index.yaml`. The entries are never removed in order to ensure backward compatibility.
- The script `script/update-error-index.py` generates:
  - `logging.h` and `logging.c` for using error identifier in logging
  - `error-index.md` to have a documentation page that lists the errors/warnings.

This is still very rough, but it illustrates the idea.

For the tools that parse these logs, I would like to add also some mechanism to ensure they keep up to date with the list of active errors. Maybe just print a list and let the tool parse it?

whot:
Apologies, I won't have a lot of time this week, so only one comment: the updater script I think you should look into using `jinja2`. That way the generated files will be a lot more readable and you don't need loads of `yield` statements - you can write out the templates like this:
```
%{ for entry in entries %}
### {{entry.name}} - {{entry.name.capitalize()}}
<dl>
   ... {{entry.foo}} ..
</dl>
%{ endfor %}
```
The YAML file is alrady a dict, so you can hand that to jinja without extra processing.

jinja has lots of documentation, the most recent bits I wrote with is [this XML parser](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/proto/ei-scanner) (look at `generate_source()` for the jinja bits) that processes [this C template](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/ei-proto.c.tmpl) (there are other `.tmpl` files in the same tree).

Since we only ever need to update this when the warnings change, requiring a `pip install jinja2` is not going to hurt.

wismill:
@whot Done

scripts/update-message-index.py
```diff
@@ -0,0 +1,175 @@
...
+import jinja2
+
+
+class Version(NamedTuple):
+    major: int
+    minor: int
+    patch: int = 0
+
+    def __str__(self):
+        return ".".join(map(str, self))
```
whot, line 18: how about
```python

    @classmethod
    def from_string(raw_version: str) -> "Version":
         version = raw_version.split(".")
         assert 2 <= len(version) <= 3 and all(n.isdecimal() for n in version), raw_version
         return Version(*map(int, version))
```

instead of a free floating function that looks all lonely? :)


Come to think of it, this would also be a good idea for the `parse_entry` and `parse_example` functions.

scripts/update-message-index.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class Entry(NamedTuple):
+    code: int
+    id: str
+    type: str
+    description: str
+    details: str
+    added: Version
+    removed: Version | None
```
whot, line 33: Can you double check which version is required for this type annotation please? I've always used `Optional[Version]` because *I think* this approach didn't work back then.

scripts/update-message-index.py
```diff
@@ -0,0 +1,175 @@
...
+        examples=examples,
+    )
+
+
+def skip_removed(entries: Sequence[Entry]):
+    return filter(lambda e: e.removed is None, entries)
+
+
+def make_message_code(entry: Entry) -> str:
+    return f"XKB-{entry.code:0>3}"
```
whot, line 87:
```python
class Entry(NamedTuple):
...
    @property
    def message_code(self) -> str:
        return f"XKB-{entry.code:0>3}"
```
and the same for message_index and name

scripts/update-message-index.py
```diff
@@ -0,0 +1,175 @@
...
+    path = root / "src" / "messages.c"
+    with path.open("wt", encoding="utf-8") as fd:
+        fd.writelines(template.generate(entries=skip_removed(index)))
+
+
+def generate_doc(index: Sequence[Entry], env: jinja2.Environment, root: Path):
+    path = root / "doc" / "message-index.md"
+    template = env.get_template("doc/message-index.md.jinja")
+    with path.open("wt", encoding="utf-8") as fd:
+        fd.writelines(template.generate(entries=index))
```
whot, line 124: nitpick: you could pass the filename into a more generic `generate()` function and either append `.jinja` or use `Path.stem()`. So basically:

```python
generate(message_index, jinja_env, args.output, root / "src" / "messages.h", skip_removed=True)
generate(message_index, jinja_env, args.output, root / "src" / "messages.c", skip_removed=True)
generate(message_index, jinja_env, args.output, root / "doc" / "message-index.md")
```

src/messages.c
```diff
@@ -0,0 +1,54 @@
...
+
+const char *
+XkbMessageDescription(enum xkb_message_index message_index)
+{
+    if (message_index < 0 || message_index >= XKB_LOG_MESSAGE_MAX_INDEX) {
+        return NULL;
+    } else {
+        return xkb_messages[message_index].description;
+    }
+}
```
whot, line 54: the `typedefs` and `XkbCamelCase` style naming is a leftover from the xkbcomp sources imported, iirc all of modern libxkbcommon uses `struct foo` and `xkb_foo`. Let's keep that here please.

whot (changes requested):
Not an in-depth review yet but I like what I'm seeing!

src/messages.c
wismill: Sure, let’s be consistent

scripts/update-message-index.py
wismill: Good idea, although I ended to use them as normal function because they are use as Jinja filters.

scripts/update-message-index.py
wismill: Python [3.10](https://docs.python.org/3.10/whatsnew/3.10.html)

> Python 3.10 was released on October 4, 2021.

I guess that is old enough? What are the requirements for Python version in xkbcommon?

Anyway it is just a developer tool and it is not used in tests.

scripts/update-message-index.py
wismill: Good idea.

wismill:
@whot Rebased, added review fixes and improved doc.

doc/message-registry.md
```diff
@@ -0,0 +1,253 @@
...
+This page lists the warnings and errors generated by xkbcommon.
+There are currently 15 entries.
+
+@todo The documentation of the log messages is a work in progress.
+
+## Index
+
+| Code      | Identifier                   | Description | Type |
+| --------- | ---------------------------- | ----------- | ---- |
+| [XKB-107] | `unrecognized-keysym` | Warn on unrecognized keysyms. | Warning |
```
bluetech, line 17: What is the logic behind the numbering? I mean why is it not 001, 002 etc.?

scripts/update-message-registry.py
```diff
@@ -0,0 +1,202 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import argparse
+from pathlib import Path
+from typing import NamedTuple, Sequence
```
bluetech, line 7: I recommend using [dataclasses](https://docs.python.org/3/library/dataclasses.html), they have nicer semantics than named tuples.

scripts/update-message-registry.py
```diff
@@ -0,0 +1,202 @@
...
+    details: str
+    """A long description of the meaning of the message"""
+    added: Version
+    """Version of xkbcommon the message has been added"""
+    removed: Version | None
+    """Version of xkbcommon the message has been removed"""
+    examples: tuple[Example, ...]
+    """
+    Optional examples of situtation the message occurs.
+    If the message is an error, provide a also hints how to fix it.
```
bluetech, line 77:
```suggestion
    Optional examples of situations in which the message occurs.
    If the message is an error, also provide hints on how to fix it.
```

src/messages.h.jinja
```diff
@@ -0,0 +1,42 @@
...
+ */
+enum xkb_message_index {
+    {% for entry in entries %}
+    /** {{ entry.description }} */
+    {{ entry|message_index }},
+    {% endfor %}
+    XKB_LOG_MESSAGE_MAX_INDEX
+};
+
+typedef uint32_t xkb_message_code_t;
```
bluetech, line 23: Why do we need to distinguish between message code and message index?

src/messages.h.jinja
```diff
@@ -0,0 +1,42 @@
+// NOTE: This file has been generated automatically by “update-message-registry.py”.
+//       Do not edit manually!
+
+#ifndef _XKBCOMMON_MESSAGE_H_
```
bluetech, line 4: We use this format for public headers; for internal headers should use `MESSAGES_H`

scripts/update-message-index.py
bluetech: Since the file uses `from __future__ import annotations`, the Python version doesn't matter for this.

bluetech (commented):
This is nice, I like it :+1:

Do you have plans to eventually expose this in the public API? That is, exposing the structured data in the library log callback.

doc/message-registry.md
wismill: No and this is on purpose. I left a note in `doc/message-registry.yaml` with this link: [Code Assignment Recommendations](https://github.com/haskellfoundation/error-message-index/blob/main/tool-developers.md#code-assignment-recommendations).

I like the reasoning, but of course this is one point we have not discussed yet.

src/messages.h.jinja
wismill: Fixed.

scripts/update-message-registry.py
wismill: Fixed

scripts/update-message-registry.py
wismill: Ok, I left `Version` out, but I figured how to implement order & use it as an iterable. Will do on next update.

scripts/update-message-registry.py
```diff
@@ -0,0 +1,205 @@
...
+template_loader = jinja2.FileSystemLoader(args.root, encoding="utf-8")
+jinja_env = jinja2.Environment(
+    loader=template_loader,
+    keep_trailing_newline=True,
+    trim_blocks=True,
+    lstrip_blocks=True,
+)
+jinja_env.filters["message_index"] = Entry.message_index
+jinja_env.filters["message_code"] = Entry.message_code
+jinja_env.filters["message_name"] = Entry.message_name
```
whot, line 195: jinja shouldn't need separate filters for these, you can use `{{entry.message_index}}` and pretend it's just an attribute. That requires `@property` in the class but otherwise it'll just work. Or it should, anyway :)

scripts/update-message-registry.py
wismill: Fixed

scripts/update-message-registry.py
wismill: I see, fixed. I use to know these things…

wismill:
@bluetech @whot I squashed the commits with a helpful message. I think the framework is almost finished.

There are still 3 points to discuss:
1. Guidelines: we need guidelines for consistent message. In particular, we need to define how to choose error _codes_. I really like the reasoning in [Code Assignment Recommendations].
2. Message code vs index: code is a stable and user-facing identifier, while index is unstable and only an implementation detail for fast lookup.
3. Public API: the only API/tool feature I would like to add is the possibility for a tool to check if the message codes it uses (e.g. to check output) are up-to-date. I think we could add this in `xkbcli`, e.g.:
  - `xkbcli messages list`: print the currently used messages “XKB-NNN: short description”.
  - `xkbcli messages check XKB-107`: exits with 0 if active, NN if removed and MM if unknown. Additional output to redirect the user to the doc in case it exist with non-0 code.

[Code Assignment Recommendations]: https://github.com/haskellfoundation/error-message-index/blob/main/tool-developers.md#code-assignment-recommendations

doc/message-registry.md
```diff
@@ -0,0 +1,253 @@
...
+You may find the list of supported keysyms in
+`include/xkbcommon/xkbcommon-keysyms.h`.
+
+
+#### Examples
+
+<details>
+  <summary>Unrecognized keysym “`coma`”</summary>
+
+__Error message:__
```
whot, line 54: I'm feeling a bit silly now but - is there a difference between `**` and `__`? seems like both make it bold and afaik `**` is vastly more common.

doc/message-registry.yaml
```diff
@@ -0,0 +1,136 @@
...
+# • Codes may be retired or introduced in new releases.
+# See the following example of guidelines:
+# https://github.com/haskellfoundation/error-message-index/blob/main/tool-developers.md#code-assignment-recommendations
+
+# TODO: add precise guidelines
+# TODO: fix missing detailed description, examples, resolution
+
+- id: "unrecognized-keysym"
+  code: 107
+  added: UNKNOWN
```
whot, line 12: 1.0.0 will do for any of the really old ones. or "always" :)

doc/message-registry.yaml
```diff
@@ -0,0 +1,136 @@
...
+  type: warning
+  description: "Warn on unrecognized keysyms."
```
whot, line 14: feel free to ignore this but afaict for short sentences like that English doesn't usually use a full stop (but German for example does)

scripts/update-message-registry.py
```diff
@@ -0,0 +1,213 @@
...
+    """A unique short human-readable string identifier"""
+    type: str
+    """Log level of the message"""
+    description: str
+    """A short description of the meaning of the message"""
+    details: str
+    """A long description of the meaning of the message"""
+    added: Version
+    """Version of xkbcommon the message has been added"""
+    removed: Version | None
```
whot, line 79: I think this requires Python 3.9 which *may* be a problem (and the walrus requires 3.8 iirc).  I'm having a deja-vu here but github is not fond of easily showing me outdated conversations, so apologies if we already discussed this here somewhere.

scripts/update-message-registry.py
```diff
@@ -0,0 +1,213 @@
...
+        assert id is not None, entry
+        type_ = entry.get("type")
+        assert type_ in cls.VALID_TYPES, entry
+        description = entry.get("description")
+        assert description is not None, entry
+        details = entry.get("details", "")
+        raw_added = entry.get("added", "")
+        assert raw_added, entry
+        added = Version.parse(raw_added)
+        assert added, entry
```
whot, line 101: I think the above code would probably be a bit easier to read with some grouping of the `get` and the `asserts`
```python
code = entry.get("code")
id = entry.get("id")
....
assert code is not None and ...
assert entry is not None
....
```

same for `Example.parse` but there it's only two fieldsd

src/xkbcomp/symbols.c
```diff
@@ -1523,7 +1545,8 @@ CopyModMapDefToKeymap(struct xkb_keymap *keymap, SymbolsInfo *info,
     else {
         key = FindKeyForSymbol(keymap, entry->u.keySym);
         if (!key) {
-            log_vrb(info->ctx, 5,
+            log_vrb_with_id(info->ctx, 5,
```
whot, line 1548: I think there's an argument to have `log_vrb()` and friends always take an id and have an `XKB_NO_ERROR_ID = 0` that can be special-cased in the log function implementations. I think there are only a few cases where we don't log with id, so better to optimise for the majority?

whot:
> Guidelines: we need guidelines for consistent message. In particular, we need to define how to choose error codes. I really like the reasoning in [Code Assignment Recommendations](https://github.com/haskellfoundation/error-message-index/blob/main/tool-developers.md#code-assignment-recommendations).

Agree, plus it means we can just point to that for explanation ;)

> Message code vs index: code is a stable and user-facing identifier, while index is unstable and only an implementation detail for fast lookup.

Admittedly I only did a relatively quick review but the code vs index approach seems fine IMO.

> Public API: the only API/tool feature I would like to add is the possibility for a tool to check if the message codes it uses (e.g. to check output) are up-to-date. I think we could add this in xkbcli, e.g.:

tbh, if we start using error messages I'd argue that they should either go into the wiki (which we'd have to enable here first :) or into the xkbcommon documentation. Beyond happy typing and some effort to maintain things, I don't quite see a benefit of either of the `xkbcli messages` tools.

scripts/update-message-registry.py
wismill: 1. this is a script aimed for _developers only_, so it should not be distributed and there is no issue with CI.
2. Python 3.9 has been released in 2020. It is not exactly recent and it brings lots of benefits.

I am not against lowering Python bounds, but I do not see why we should held back on old Python versions for _developer scripts_. Recent versions bring lots of nice enhancements and it is quite easy to have multiple versions of Python installed in parallel.

I agree Python version must be carefully chosen for things such as build scripts, running on CI or on user side, but here this is _not_ the case.

We should agree and indicate the minimum requirements to work on this repo.

src/xkbcomp/symbols.c
wismill: I agree, I just did not want to start such refactor for now, because we need to agree on previous points and it is going to create a huge diff. What do you prefer:

- Create a further MR to do this refactor.
- Make this refactor in a separate commit: it facilitate reviewing by commit, then it can be squashed.

scripts/update-message-registry.py
wismill: I find it better when code is grouped semantically and when it fails early, to avoid unnecessary work. I added spaces to improve the reading. Is it good enough so?

doc/message-registry.yaml
wismill: Fixed

doc/message-registry.yaml
wismill: It is displayed as "1.0.0" in the generated doc, but I prefer to keep “UNKNOWN” in the register. I think it is better to be explicit. Maybe “TODO” is a good alternative.

doc/message-registry.md
wismill: Same. I find the style `__` easier to read and I used it in other MR.

wismill:
>>    Public API: the only API/tool feature I would like to add is the possibility for a tool to check if the message codes it uses (e.g. to check output) are up-to-date. I think we could add this in xkbcli, e.g.:

> tbh, if we start using error messages I'd argue that they should either go into the wiki (which we'd have to enable here first :) or into the xkbcommon documentation. 

This MR does add a documentation page for error messages.

> Beyond happy typing and some effort to maintain things, I don't quite see a benefit of either of the xkbcli messages tools.

The benefit is that out-of-sync will not be _silently ignored_ in CI. A tool parsing log output should have some way to avoid out-of-sync, else it is worthless because it may give the illusion tests are passing.

I intend just to make the most basic tool to check this, so there should be no maintenance once written. I think `xkbcli messages-check XKB-107 XKB-489` or maybe better `xkbcli messages-check 107 489` would be quite handy.

src/xkbcomp/symbols.c
whot: honestly, whatever is the combination of easiest to do vs. easiest to review. I'm happy with this being a follow-up - in the end the effort typing and testing is going to be worse than the effort looking at the patch after so it's up to you

whot:
> @whot This MR does add a documentation page for error messages.

oh, right, whoops. I did stare at that but then skipped ahead because it was autogenerated :)

> I intend just to make the most basic tool to check this, so there should be no maintenance once written. I think xkbcli messages-check XKB-107 XKB-489 or maybe better xkbcli messages-check 107 489 would be quite handy.

In that case I'd argue make them non-installed helper tools, possibly in the `test/` directory. Because the tools have their use in the CI but I'd rather not have them look like official tools (or worse: installed) because then it's another tool that we need to worry about usages for.

doc/message-registry.yaml
wismill: Done: I used ALWAYS and added a comment to give context.

src/messages.h.jinja
wismill: @bluetech does this approach seems right to you? I would like to have a good design for the message framework before starting to modify every log entry.

bluetech: I think it would be better to eliminate the "message index" concept, and use just the codes, in order to simplify things. Because the codes are not consecutive, it means that whenever they're looked up, will need to do the binary search on the array. But since it is absolutely not performance critical, and the array is quite small (even if it had 1000 entries it would still be small for that matter), I don't think that's a problem.

However, if you prefer to keep the indexes, that's OK by me.

src/messages.h.jinja
```diff
@@ -0,0 +1,45 @@
...
+    {% endfor %}
+    XKB_LOG_MESSAGE_MAX_INDEX
+};
+
+typedef uint32_t xkb_message_code_t;
+
+#define XKB_INVALID_MESSAGE_INDEX 0xFFFFFFFFu
+
+typedef struct {
+    xkb_message_code_t code;
```
bluetech, line 28:
```suggestion
    const xkb_message_code_t code;
```

src/messages.h.jinja
```diff
@@ -0,0 +1,45 @@
...
+    const char *identifier;
+    const char *description;
+} xkb_message_entry;
+
+xkb_message_code_t
+xkb_message_code(enum xkb_message_index message_index);
```
bluetech, line 34: If it were me, I'd have just one function, `const *struct xkb_message_entry xkb_message_get(xkb_message_code_t code)`, and use it everywhere by accessing the fields, e.g. `xkb_message_get(...).identifier` instead of `xkb_message_identifier(...)`.

But that's a matter of taste, take it or leave it.

bluetech (approved):
I left a couple of comments for your consideration, you can take them or leave them. I've read the comments so far and the current state LGTM, so ready to merge once @whot also approves.

src/messages.h.jinja
wismill: Removed the index concept. In fact we do not need to do the lookup.

src/messages.h.jinja
wismill: Fixed

meson.build
```diff
@@ -532,6 +532,16 @@ You can disable the Wayland xkbcli programs with -Denable-wayland=false.''')
...
+
+    if get_option('enable-xkb-check-messages')
+        executable('xkb-check-messages',
+                   'tools/check-messages.c',
+                   'src/messages.h',
+                   'src/messages.c',
+                   dependencies: [tools_dep],
+                   include_directories: [include_directories('src', 'include')],
+                   install: false)
+    endif
```
whot, line 544: remove the option and build it unconditionally - this doesn't introduce any new dependencies (unless I'm misreading something here) and since it doesn't get installed there's no point hiding it behind an option. Apologies if my rant against new tools was unclear regarding that :)

tools/check-messages.c
```diff
@@ -0,0 +1,83 @@
...
+                    argv[k]);
+            rc |= 0b01;
+            continue;
+        }
+        entry = xkb_message_get(code);
+        if (entry == NULL) {
+            fprintf(stderr,
+                    XKB_CHECK_MSG_ERROR_PREFIX "Unsupported message code: %s\n",
+                    argv[k]);
+            rc |= 0b10;
```
whot, line 77: you and your new-fangled features :)

tbh, in this case I don't think it's better than just a `rc |= 2;`, binary notation for me doesn't really become useful until we hit 4 binary digits or more. Not going to fight this too hard though (though I expect there'll be *some* system somewhere that will complain)

src/messages.c
```diff
@@ -0,0 +1,47 @@
...
+    /* binary search in xkb_messages */
+    while (last >= first) {
+        size_t mid = (first + last) / 2;
+        if (xkb_messages[mid].code < code)
+            first = mid + 1;
+        else if (xkb_messages[mid].code > code)
+            last = mid - 1;
+        else /* found it */
+            return &xkb_messages[mid];
+    }
```
whot, line 44: my absolute guess is that with the handful of messages we have, simply iterating through the list until found or `code > current` would only be fractions of a milliseconds but make the code simpler and more robust to the few edge cases. 

As ridiculous as this sounds I'd like to see a test *somewhere* to make sure this function works with the various edge cases. But it looks correct and it's not in a critical path, so this can be done as follow-up.

whot (approved):
LGTM! I think we should remove the meson option, it doesn't provide anything of value right now but other than that this can be merged.

src/messages.c.jinja
```diff
@@ -0,0 +1,47 @@
...
+//       Do not edit manually!
+
+#include <stdlib.h>
+#include "messages.h"
+
+static const struct xkb_message_entry xkb_messages[] = {
+    {% for entry in entries %}
+    { {#--#}{{ entry.message_code_constant}}{#, "{{entry.id}}", "{{entry.description}}"#}},
+    {% endfor %}
+    {0{#, NULL, NULL #}}
```
bluetech, line 11: Why is this needed BTW?

src/messages.c.jinja
```diff
@@ -0,0 +1,47 @@
...
+};
+
+const struct xkb_message_entry*
+xkb_message_get(xkb_message_code_t code)
+{
+{# Binary search seems overkill for now
+    if (code < XKB_LOG_MESSAGE_MIN_CODE || code > XKB_LOG_MESSAGE_MAX_CODE)
```
bluetech, line 18: Probably better to just remove this commented out code (can leave the "Binary search seems overkill for now" comment though).

src/messages.c.jinja
```diff
@@ -0,0 +1,47 @@
...
+        else /* found it */
+            return &xkb_messages[mid];
+    }
+#}
+    if (code < XKB_LOG_MESSAGE_MIN_CODE || code > XKB_LOG_MESSAGE_MAX_CODE)
+        return NULL;
+
+    size_t idx_max = sizeof(xkb_messages) / sizeof(*xkb_messages) - 1;
+
+    for (size_t idx = 0; idx < idx_max; idx++) {
```
bluetech, line 40:
```suggestion
    for (size_t idx = 0; idx < ARRAY_SIZE(xkb_messages) - 1; idx++) {
```

(drop the `-1` if you remove the `{0}` from `xkb_messages`)

src/scanner-utils.h
```diff
@@ -57,15 +57,28 @@ struct scanner {
     void *priv;
 };
 
+#define scanner_log_with_id(scanner, level, log_msg_id, fmt, ...) \
```
bluetech, line 60: Maybe replace `_with_id` -> `_with_code` now? Since we no longer use IDs

bluetech (approved):

src/messages.c.jinja
wismill: The table is needed for the lookup. But in fact we do not use it in the main lib. I refactored it in the last version: I moved it to `tools/` and use the code only in the tool and the corresponding test.

src/messages.c.jinja
wismill: Done

src/messages.c.jinja
wismill: Done

wismill:
I refactored the code: as for now, the lib does not need to perform message code lookup. So the corresponding table is moved to `tools/` and will be used only by `xkb-check-messages` and the corresponding test.

I am waiting for #357 to be merge in order to rebase and adapt the new numeric keysym warning.

tools/messages.c
```diff
@@ -0,0 +1,78 @@
...
+{
+    *messages = xkb_messages;
+    return ARRAY_SIZE(xkb_messages);
+}
+
+const struct xkb_message_entry*
+xkb_message_get(xkb_message_code_t code)
+{
+if (code < XKB_LOG_MESSAGE_MIN_CODE || code > XKB_LOG_MESSAGE_MAX_CODE)
+        return NULL;
```
whot, line 69: indentation is off here

src/messages-codes.h
```diff
@@ -0,0 +1,52 @@
...
+    XKB_WARNING_CONFLICTING_MODMAP = 800,
+    /** Warn if there are conflicting actions while merging keys */
+    XKB_WARNING_CONFLICTING_KEY_ACTION = 883,
+    /** Warn if there are conflicting key types while merging groups */
+    XKB_WARNING_CONFLICTING_KEY_TYPE = 893,
+    /** Warn if there are conflicting fields while merging keys */
+    XKB_WARNING_CONFLICTING_KEY_FIELDS = 935,
+    /** Warn if using a symbol not defined in the keymap */
+    XKB_WARNING_UNRESOLVED_KEYMAP_SYMBOL = 965,
+    XKB_LOG_MESSAGE_MAX_CODE = 965
```
whot, line 47: can you prefix min/max with `_` please? Makes them stick out more. Can be follow-up, it's not public API anyway and I don't want to make you do more busywork...

whot (approved):

tools/messages.c
wismill: Fixed

src/messages-codes.h
wismill: Done

src/messages.c
wismill: Done

tools/check-messages.c
wismill: Done

meson.build

doc/message-registry.md
wismill: Documented in registry & commit

wismill:
Rebased.

wismill:
I would like to continue this work in a follow-up PR, that ideally would be part of 1.6 release.

Issue #311, opened 2022-12-13
can't add more than 4 layouts to the keymap


bam80:
https://github.com/xkbcommon/libxkbcommon/blob/2530f6444bfad2bf67ee926e57df8987afeebf4a/src/keymap.h#L91-L104
> This limit is artificially enforced

It's not documented anywhere, which is an issue per se I already faced, and also it might be useful to add more layouts for example as in our KDE KWin's unit tests (I'll provide the link).
@bluetech

whot:
Realistically, the  main limitation here is XKB. The XKB protocol (the X11 bit) has the limit of 4 groups hardcoded, so if we want to stay compatible to that (and we do, for the forseeable future) extending this is not an option.

A quick peek of the protocol indicates only `XkbGetMap` has that *actually* as a sized limit (`ktIndex` is limited to 4 groups), afaict everything else could make use of at least 256 or (as in that comment) 32 where used as mask. But `XkbNumKbdGroups` has been `#defined` to 4 for decades now and I'm sure there is plenty of code out there that hardcodes it. the protocol also defines `KB_GROUPMASK` as 1-4 groups, so extending this will also cause issues.

bam80:
Could we make in opt-in maybe? Because right now we already faced with the real issue above due to the limitation, which needs to be solved anyway.

wismill:
Duplicate of #37, let’s continue the discussion there as it has more info.

PR #334, opened 2023-05-07
Add support for :all in rules

wismill:
This PR adds support for the [`:all`][xkbcomp issue] modifier in rules & include statements.

For example, the `xkeyboard-config` issue “[rules: add custom per-group mapping for level3(ralt_alt)](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/253)” could be solved easier with this feature:

```
! layout        option	        = symbols
*               lv3:ralt_alt    = +level3(ralt_alt):all
```

`xkbcomp` would require a patch too, though (see the corresponding [xkbcomp issue]).

This feature works by detecting the number of layouts in the RMLVO configuration. Example: it will replace the include statement `xx:all` by `xx:1+xx:2+xx:3` if there are 3 layouts and the merge mode is “override”.

__Caveat:__ it only works when a keymap is created from RMLVO configuration; i.e. it will not work when creating  a keymap e.g. from a string. But I guess this case is not so important, because as I understand querying the keymap to the windows system will return a complete & resolved  keymap without include statements.

Note: this could the first step to remove the limit of layout number (see #311, @bam80). 

[xkbcomp issue]: https://gitlab.freedesktop.org/xorg/app/xkbcomp/-/issues/23

### DIFF
--- a/bench/rules.c
+++ b/bench/rules.c
@@ -53,7 +53,8 @@ main(int argc, char *argv[])
     for (i = 0; i < BENCHMARK_ITERATIONS; i++) {
         struct xkb_component_names kccgst;
 
-        assert(xkb_components_from_rules(ctx, &rmlvo, &kccgst));
+        // [FIXME] num_explicit_groups
+        assert(xkb_components_from_rules(ctx, &rmlvo, &kccgst, NULL));
         free(kccgst.keycodes);
         free(kccgst.types);
         free(kccgst.compat);
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -68,6 +68,7 @@ xkb_keymap_new(struct xkb_context *ctx,
 
     keymap->format = format;
     keymap->flags = flags;
+    keymap->num_explicit_groups = 0;
 
     update_builtin_keymap_fields(keymap);
 
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -389,6 +389,8 @@ struct xkb_keymap {
 
     /* Number of groups in the key with the most groups. */
     xkb_layout_index_t num_groups;
+    /* Number of groups that were explicitly declared at section level. */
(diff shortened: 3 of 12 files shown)


bam80:
Thanks for notifying me Wismill!
Looking forward to get this merged.

bam80:
> For example, the xkeyboard-config issue “[rules: add custom per-group mapping for level3(ralt_alt)](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/253)” could be solved easier with this feature:

> ```
> ! layout        option	        = symbols
> *               lv3:ralt_alt    = +level3(ralt_alt):all
> ```

Could we actually make it easier as part of this PR?

wismill:
You mean patching the rule file in `xkeyboard-config` ? Different project; it will require X11 `xkbcomp` to be patched before.

bam80:
OK, do you consider to make a patch for X11 xkbcomp also?

wismill:
> OK, do you consider to make a patch for X11 xkbcomp also?

@bam80 Let’s focus on this PR. Having a first draft enables us to discuss its design with the experts @whot and @bluetech. I also have to fix the tests first!

bluetech:
@wismill I'd like to review this PR (it'd be really great to fix this!), it might take me a bit of time to get to it though, wanted to give you a heads up.

wismill:
No rush, currently the tests are failing. I am content waiting for your review: I would like to get the design right, as it would be wonderful to port it to X11’s `xkbcomp` then start moving things in `xkeyboard-config`. Though it would probably take years to transition the rules, if  distros need to first pickup compatible `xkbcomp`.

wismill:
Closing. I am working on another approach, where thw processing happen only in the rules. This should facilitate a possible adaptation to libxkbfile.

PR #337, opened 2023-05-10
Add warning when RMLVO with no layout but variant set

### LINKED ISSUE #336: Just variant in RMLVO does nothing
It's surprising that if you only set the variant in the RMLVO, the variant is ignored.

I am using Go bindings but this:

```
names := xkb.RuleNames{Variant: "dvorak"}
ctx := xkb.ContextNew(xkb.ContextNoFlags)
keymap = ctx.KeymapNewFromNames(&names, xkb.KeymapCompileNoFlags)
```
shows:

    xkbcommon: DEBUG: Compiling from RMLVO: rules 'evdev', model 'pc105', layout 'us', variant '(null)', options ''
    xkbcommon: DEBUG: Compiling from KcCGST: keycodes 'evdev+aliases(qwerty)', types 'complete', compat 'complete', symbols 'pc+us+inet(evdev)'

Whereas this:

```
names := xkb.RuleNames{Layout: "us", Variant: "dvorak"}
ctx := xkb.ContextNew(xkb.ContextNoFlags)
keymap = ctx.KeymapNewFromNames(&names, xkb.KeymapCompileNoFlags)
```
shows:

    xkbcommon: DEBUG: Compiling from RMLVO: rules 'evdev', model 'pc105', layout 'us', variant 'dvorak', options ''
    xkbcommon: DEBUG: Compiling from KcCGST: keycodes 'evdev+aliases(qwerty)', types 'complete', compat 'complete', symbols 'pc+us(dvorak)+inet(evdev)'

If this is the intended behavior, how can I make keymap with user's default layout but a specified variant?

wismill:
There is no feedback that they are both replaced with default values.

Fixes #336

### DIFF
--- a/src/context-priv.c
+++ b/src/context-priv.c
@@ -179,6 +179,16 @@ xkb_context_sanitize_rule_names(struct xkb_context *ctx,
      * the caller and one from the environment. */
     if (isempty(rmlvo->layout)) {
         rmlvo->layout = xkb_context_get_default_layout(ctx);
+        if (!isempty(rmlvo->variant)) {
+            const char *variant = xkb_context_get_default_variant(ctx);
+            log_warn(ctx,
+                "Layout not provided, but variant set to \"%s\": "
+                "ignoring variant and using defaults for both: "
+                "layout=\"%s\", variant=\"%s\".\n",
+                rmlvo->variant,
+                rmlvo->layout,
+                variant ? variant : "");
+        }
         rmlvo->variant = xkb_context_get_default_variant(ctx);
     }
     /* Options can be empty, so respect that if passed in. */


wismill:
Should we also say what are these defaults? It seems a good idea.

src/context-priv.c
```diff
@@ -178,6 +178,11 @@ xkb_context_sanitize_rule_names(struct xkb_context *ctx,
     /* Layout and variant are tied together, so don't try to use one from
      * the caller and one from the environment. */
     if (isempty(rmlvo->layout)) {
+        if (!isempty(rmlvo->variant))
+            log_warn(ctx,
+                "Layout not provided, but variant set to \"%s\": "
+                "ignore variant and use defaults for both.\n",
```
bluetech, line 184:
```suggestion
                "ignoring variant and using defaults for both.\n",
```

bluetech (commented):
Thanks, looks good to me.

> Should we also say what are these defaults? It seems a good idea.

I think that would be nice.

src/context-priv.c
wismill: Fixed.

wismill:
> > Should we also say what are these defaults? It seems a good idea.
> 
> I think that would be nice.

Now I am thinking we could do the same for the other settings, but at INFO level because it is not erroneous. What do you think?

bluetech:
> Now I am thinking we could do the same for the other settings, but at INFO level because it is not erroneous. What do you think?

Maybe a single info log showing the full RMLVO selected? (If there isn't already one that is)

PR #338, opened 2023-05-10
Do not interpret nor emit invalid Unicode encoding forms

### LINKED ISSUE #60: Surrogates not handled
Surrogates (0xd800..0xdfff) are not handled by utf32_to_utf8 and by xkb_keysym_to_utf32. They are invalid code points on utf32 and utf8 but used for code pairs in utf16. is_valid_utf8 works as it should and return numbers inside this range as invalid. 

Please, see my fork about it. I recoded the utf32_to_utf8 and is_valid_utf8 to be more straightforward (and easier to maintain). As an added benefit we get an utf8_to_utf32 function.

Regards,
André

wismill:
Surrogates are invalid in both [UTF-32](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#G28875) and [UTF-8](https://www.unicode.org/versions/Unicode15.0.0/ch03.pdf#G31703).

Fixes #60

### DIFF
--- a/src/keysym-utf.c
+++ b/src/keysym-utf.c
@@ -41,6 +41,8 @@
 #include "utils.h"
 #include "utf8.h"
 
+#define NO_KEYSYM_UNICODE_CONVERSION 0
+
 /* We don't use the uint32_t types here, to save some space. */
 struct codepair {
     uint16_t keysym;
@@ -847,7 +849,7 @@ bin_search(const struct codepair *table, size_t length, xkb_keysym_t keysym)
     }
 
     /* no matching Unicode value found in table */
-    return 0;
+    return NO_KEYSYM_UNICODE_CONVERSION;
 }
 
 XKB_EXPORT uint32_t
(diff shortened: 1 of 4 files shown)


src/keysym-utf.c
```diff
@@ -871,6 +873,12 @@ xkb_keysym_to_utf32(xkb_keysym_t keysym)
         return keysym & 0x7f;
 
     /* also check for directly encoded Unicode codepoints */
+    /* Exclude surrogates: they are invalid in UTF-32.
```
bluetech, line 876:
```suggestion
    /* also check for directly encoded Unicode codepoints */

    /* Exclude surrogates: they are invalid in UTF-32.
```

src/utf8.c
```diff
@@ -47,6 +52,10 @@ utf32_to_utf8(uint32_t unichar, char *buffer)
         length = 2;
         head = 0xc0;
     }
+    /* Handle surrogates */
+    else if (0xd800 <= unichar && unichar <= 0xdfff) {
+        goto ill_formed_code_unit_subsequence;
```
bluetech, line 57: goto from an if branch to an else branch is too wacky for me :) I suggest one of the following:

1. Repeat the `buffer[0] = '\0'; return 0;` code.
2. Move `ill_formed_code_unit_subsequence` label to the end of the function, after the `return length + 1;`, and `goto` there from both branches.

bluetech (commented):
Thanks, I think we should reject the surrogates as the standards specify.

I left a couple of comments, can merge after they are fixed.

src/utf8.c
wismill: Fixed using 2nd option

PR #339, opened 2023-05-10
Re-enable x11comp test

### LINKED ISSUE #30: test/x11comp started to fail
Presumably after a recent update to either xkbcomp or xserver. In retrospect, this test is not very future proof... But it is useful, so I should check out the diff.

wismill:
This test works on my machine, let’s see if it works on our CI.

Fixes #30

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -31,6 +31,16 @@ jobs:
           sudo apt install -y \
             doxygen libxcb-xkb-dev valgrind ninja-build \
             libwayland-dev wayland-protocols bison graphviz
+      - name: Install xkeyboard-config
+        run: |
+          # HACK: Force xkeyboard-config version to master
+          pushd ~
+          git clone --depth=1 https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config.git
+          cd "xkeyboard-config"
+          BUILDDIR=build
+          meson setup $BUILDDIR -Dprefix=/usr
+          meson install -C $BUILDDIR
+          popd
       - name: Setup
         run: |
           # -gdwarf-4 - see https://github.com/llvm/llvm-project/issues/56550.
--- a/test/x11comp.c
+++ b/test/x11comp.c
@@ -56,7 +56,7 @@ main(void)
     pid_t xkbcomp_pid;
 
     char *xhost = NULL;
-    int xdpy_current;
+    int xdpy_current = 0;
     int xdpy_candidate;
 
     /*
@@ -71,8 +71,12 @@ main(void)
      * 6. Kill the server & clean up.
      */
 
-    ret = xcb_parse_display(NULL, &xhost, &xdpy_current, NULL);
-    assert(ret != 0);
+    /*
+     * Detect current display, in order to avoid reusing it.
+     * It is OK if it fails (e.g. in headless mode).
+     */
+    xcb_parse_display(NULL, &xhost, &xdpy_current, NULL);
+
     /*
      * IANA assigns TCP port numbers from 6000 through 6063 to X11
      * clients.  In addition, the current XCB implementaion shows
@@ -85,7 +89,9 @@ main(void)
         if (xdpy_candidate == xdpy_current) {
             continue;
         }
-        snprintf(display, sizeof(display), "%s:%d", xhost, xdpy_candidate);
+        snprintf(display, sizeof(display),
+                 "%s:%d", (xhost != NULL) ? xhost : "",
+                 xdpy_candidate);
         ret = posix_spawnp(&xvfb_pid, "Xvfb", NULL, NULL, xvfb_argv, envp);
         if (ret == 0) {
             break;
--- a/meson.build
+++ b/meson.build
@@ -689,9 +689,11 @@ if get_option('enable-x11')
         executable('test-x11', 'test/x11.c', dependencies: x11_test_dep),
         env: test_env,
     )
-    # test/x11comp is meant to be run, but it is (temporarily?) disabled.
-    # See: https://github.com/xkbcommon/libxkbcommon/issues/30
-    executable('test-x11comp', 'test/x11comp.c', dependencies: x11_test_dep)
+    test(
+        'x11comp',
+        executable('test-x11comp', 'test/x11comp.c', dependencies: x11_test_dep),
+        env: test_env,
+    )
 endif
 if get_option('enable-xkbregistry')
     test(


wismill:
I am doing some trials locally, but I encounter a weird bug that seems to be https://gitlab.freedesktop.org/xorg/xserver/-/issues/1249.

It may be the same bug that triggers with this test when using `xkbcomp`. I do not think there is much we can do for now…

bluetech:
Seems like the bug you linked to was fixed by @whot some months ago, I guess the xserver in CI is too old to include it? I just updated the CI linux job to run on the latest ubuntu (22.04); it's probably too old as well, but worth a try anyway to rebase on master and see what happens.

wismill:
IIRC, I already tried 22.04 locally and the issue was still there.

bluetech:
Then I'll guess we'll have to wait a few more year :)

However, if there are changes you want to make to the test, while still skipping it for now, let me know, we can merge that.

wismill:
It's not ready to merge. Maybe the only change worth to notice is the use of `xvfb-run`.

whot:
> I guess the xserver in CI is too old to include it? 

Looks like it *should* be included, at least in 22.04:
```
xserver $ git tag --contains 2c6989f81e62bac05a583c39fa09a06172d1ece5
xorg-server-21.1.2
xorg-server-21.1.3
xorg-server-21.1.4
...
```
Note this is the cherry-pick commit of that patch so it only lists the `21.1.x` releases.  The [ubuntu package](https://launchpad.net/ubuntu/jammy/+package/xserver-xorg-dev) includes 21.1.3 atm. So this *should* work if it's the same bug.

But checking the workflow yaml file: we don't appear to be installing X on purpose, might be good to have this listed explicitly. maybe it's pulled in by graphviz?

Anyway, testlog shows it does indeed run but otherwise the output is not very useful. Best way to reproduce locally would be with
```
$ podman  run --rm -it ubuntu:22.04
```
and then install the dependencies and run in that container. At least that way it can be debugged interactively instead of having short meson logs.

wismill:
@whot @wismill I think I finally make it work!

Some comments:
- I initially tried to use `xvfb-run`, but it turns out to be problematic with valgrind: some memory issues with `dash`.
- Linux CI: For some reasons it does not work with older versions of `xkb-data` package, i.e. `xkeyboard-config`. I do not know which one precisely, but it seems 2.38 (Ubuntu 23.04) and 2.39 work. I use a _hack_ to override the system version. Not pretty, but it works. Do you have a better idea?
- macOS CI: for some reason unrelated to my modifications this CI started to fail. Any idea?

.github/workflows/linux.yml
```diff
@@ -26,11 +26,23 @@ jobs:
...
+            libwayland-dev wayland-protocols bison graphviz curl
+          python -m pip install --upgrade PyYAML
+      - name: Install xkeyboard-config
+        run: |
+          # HACK: Force xkeyboard-config version
+          pushd ~
+          XKC_VER=2.39
+          curl -O "https://www.x.org/releases/individual/data/xkeyboard-config/xkeyboard-config-$XKC_VER.tar.xz"
+          tar -xf "xkeyboard-config-$XKC_VER.tar.xz"
+          cd "xkeyboard-config-$XKC_VER"
```
whot, line 41: mostly a personal preference, but why don't we build from a `git clone --depth=1 --branch=2.39`? tarballs seem like *sooo* last century :)

test/x11comp.c
```diff
@@ -71,8 +71,9 @@ main(void)
      * 6. Kill the server & clean up.
      */
 
+    // NOTE: it is OK if it fails.
```
whot, line 74:
```suggestion
    /* NOTE: it is OK if it fails. */
```

.github/workflows/linux.yml
```diff
@@ -26,11 +26,23 @@ jobs:
...
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade meson PyYAML
           sudo apt update
           sudo apt install -y \
-            doxygen libxcb-xkb-dev valgrind ninja-build \
-            libwayland-dev wayland-protocols bison graphviz
+            doxygen libxcb-xkb-dev valgrind ninja-build meson \
+            libwayland-dev wayland-protocols bison graphviz curl
+          python -m pip install --upgrade PyYAML
```
whot, line 33: this rework means we're no longer testing against the latest released meson but against whatever our ubuntu version ships. That's generally not ideal - any specific reason why you reshuffled this? I don't see the old failing logs so it's unclear

whot:
> I initially tried to use xvfb-run, but it turns out to be problematic with valgrind: some memory issues with dash.

can you install bash and run in that somehow? You can also add a valgrind suppressions file like [the one libinput has](https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/test/valgrind.suppressions) - that one has a few bash-specifics in it because we really don't care about those.

>  I use a hack to override the system version. Not pretty, but it works. Do you have a better idea?

tbh, building against whatever the latest version of xkeyboard-config is seems like a good idea anyway, the two projects are quite intertwined so we want things to blow up early.

test/x11comp.c
wismill: Fixed. I also made it more explicit.

.github/workflows/linux.yml
wismill: Use master, as per your last comment.

wismill:
> can you install bash and run in that somehow? You can also add a valgrind suppressions file like [the one libinput has](https://gitlab.freedesktop.org/libinput/libinput/-/blob/main/test/valgrind.suppressions) - that one has a few bash-specifics in it because we really don't care about those.

@whot I already tried bash and it has similar issues. I think it is better to to avoid `xvfb-run` entirely, as these suppressions files may also require maintenance.

The macOS CI is failing. Is `sh: dot: command not found` the only reason?

.github/workflows/linux.yml
wismill: Fixed.

Testing locally with podman, it uses `root`, which cannot use pip to update system dependencies. How I reproduce the exact same config as the CI locally?

wismill:
Let’s investigate the macOS issue in #346. Then I will rebase.

.github/workflows/linux.yml
whot: hmm, good question, I don't actually know how to pull the same containers (VMs?) github uses. if you just need an extra user you can create that inside podman and then `su` to it.

whot (approved):
better run + skip than not run at all! :)

whot:
> as valgrind suppressions files may also require maintenance.

Luckily very little but yes, they do. I think in libinput it averaged to one or two per year but none since 2020 so it's not the worst of all things (and libinput builds against *a lot* of distros/versions).

PR #342, opened 2023-05-13
Add support for stable doc URLs

### LINKED ISSUE #341: Ensure documentation pages have stable URLs
The recent changes I made changed some URLs. This is not a good idea (see: “[Cool URIs don't change](https://www.w3.org/Provider/Style/URI.html)”).

On the other hand, embracing Doxygen features is probably a good idea for better documentation with cross references, etc.

On solution I see is to add redirections: see e.g. [this example](https://stackoverflow.com/a/51175138).

wismill:
Doc URLs may change with time because they depend on Doxygen machinery. This is unfortunate because it is good practice to keep valid URLs (see: https://www.w3.org/Provider/Style/URI.html).

This commit adds a registry of URLs that has three functions:
- Check no previous URL is invalid.
- Add aliases for moved pages.
- Generate redirection pages for aliases.

Fixes #341


bluetech (approved):
Looks great to me, thanks!

bluetech:
@wismill can you shoot me an email (my address should be visible in https://github.com/bluetech or in git logs) - I'd like to invite to you to be a member of @xkbcommon/maintainers so you can merge this yourself :)

wismill:
@bluetech thank you for the invitation, I sent you an email. Please check your spam, as my domain wismill.eu is not always treated gently by servers 😒

wismill:
The CI correctly caught the missing doc pages that I intentionally left unfixed. Now let’s check if it works with the updated registry.

PR #346, opened 2023-06-23
Fix mac os CI

wismill:


wismill:
The issue seems to be with Doxygen 1.9.7, which changed the generated file names. It does not trigger in Linux because we use an older version.

wismill:
There is a warning in the macOS CI:

```
Warning: Refusing to link macOS provided/shadowed software: bison
If you need to have bison first in your PATH, run:
  echo 'export PATH="/usr/local/opt/bison/bin:$PATH"' >> /Users/runner/.bash_profile

For compilers to find bison you may need to set:
  export LDFLAGS="-L/usr/local/opt/bison/lib"
```

I do not know if we should fix it.

whot (approved):
Can you add a reference to the doxygen bug in the commit message please? Saves us from having to dig up this PR in the future to check if it's still an issue and whatnot. Otherwise this LGTM, thanks!

PR #348, opened 2023-06-26
Test: Use a xvfb wrapper  for x11 test

wismill:
The X11 test is currently silently skipped in CI. Use a `xvfb` wrapper to ensure to run it. See #339 for the rationale.

The code is adapted from #339 ~and it is intended to be rebased on it, then refactor the x11comp test as well with the wrapper~, which is closed in favour of this new PR.

Question: should we use an option to disable the wrapper, in order to test with a local X11 server as previously?

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -31,6 +31,19 @@ jobs:
           sudo apt install -y \
             doxygen libxcb-xkb-dev valgrind ninja-build \
             libwayland-dev wayland-protocols bison graphviz
+      - name: Install xkeyboard-config
+        run: |
+          # Install master version of xkeyboard-config, in order to ensure
+          # its latest version works well with xkbcommon.
+          # HACK: We use meson to install, while it would be cleaner
+          #       to create a proper package to install or use some PPA.
+          pushd ~
+          git clone --depth=1 https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config.git
+          cd "xkeyboard-config"
+          BUILDDIR=build
+          meson setup $BUILDDIR -Dprefix=/usr
+          meson install -C $BUILDDIR
+          popd
       - name: Setup
         run: |
           # -gdwarf-4 - see https://github.com/llvm/llvm-project/issues/56550.
--- a/meson.build
+++ b/meson.build
@@ -573,9 +573,11 @@ test_dep = declare_dependency(
     link_with: libxkbcommon_test_internal,
 )
 if get_option('enable-x11')
-    libxkbcommon_x11_internal = static_library(
+    libxkbcommon_x11_test_internal = static_library(
         'xkbcommon-x11-internal',
         libxkbcommon_x11_sources,
+        'test/xvfb-wrapper.c',
+        'test/xvfb-wrapper.h',
         include_directories: include_directories('src', 'include'),
         link_with: libxkbcommon_test_internal,
         dependencies: [
@@ -584,7 +586,7 @@ if get_option('enable-x11')
         ],
     )
     x11_test_dep = declare_dependency(
-        link_with: libxkbcommon_x11_internal,
+        link_with: libxkbcommon_x11_test_internal,
         dependencies: [
             test_dep,
             xcb_dep,
--- a/meson.build
+++ b/meson.build
@@ -691,9 +691,11 @@ if get_option('enable-x11')
         executable('test-x11', 'test/x11.c', dependencies: x11_test_dep),
         env: test_env,
     )
-    # test/x11comp is meant to be run, but it is (temporarily?) disabled.
-    # See: https://github.com/xkbcommon/libxkbcommon/issues/30
-    executable('test-x11comp', 'test/x11comp.c', dependencies: x11_test_dep)
+    test(
+        'x11comp',
+        executable('test-x11comp', 'test/x11comp.c', dependencies: x11_test_dep),
+        env: test_env,
+    )
 endif
 if get_option('enable-xkbregistry')
     test(
--- a/test/x11.c
+++ b/test/x11.c
@@ -24,10 +24,10 @@
 #include "config.h"
 
 #include "test.h"
+#include "xvfb-wrapper.h"
 #include "xkbcommon/xkbcommon-x11.h"
 
-int
-main(void)
+X11_TEST(test_basic)
 {
     struct xkb_context *ctx = test_get_context(0);
     xcb_connection_t *conn;
@@ -43,7 +43,7 @@ main(void)
     * If it fails, it's not necessarily an actual problem with the code.
     * So we don't want a FAIL here.
     */
-    conn = xcb_connect(NULL, NULL);
+    conn = xcb_connect(display, NULL);
     if (!conn || xcb_connection_has_error(conn)) {
         exit_code = SKIP_TEST;
         goto err_conn;
@@ -83,4 +83,8 @@ err_conn:
     xkb_context_unref(ctx);
 
     return exit_code;
+}
+
+int main(void) {
+    return x11_tests_run();
 }
--- a/test/xvfb-wrapper.c
+++ b/test/xvfb-wrapper.c
@@ -35,17 +35,28 @@
 #include "xvfb-wrapper.h"
 #include "xkbcommon/xkbcommon-x11.h"
 
+static bool xvfb_is_ready;
+
+static void
+sigusr1_handler(int signal)
+{
+    xvfb_is_ready = true;
+}
+
 int
 xvfb_wrapper(int (*test_func)(char* display))
 {
     int ret = 0;
     FILE * display_fd;
     char display_fd_string[32];
+    sigset_t mask;
+    struct sigaction sa;
     char *xvfb_argv[] = {
         (char *) "Xvfb", (char *) "-displayfd", display_fd_string, NULL
     };
     char *envp[] = { NULL };
     pid_t xvfb_pid = 0;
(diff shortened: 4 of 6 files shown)


test/xvfb-wrapper.c
```diff
@@ -0,0 +1,105 @@
...
+
+    /*
+     * Detect current display, in order to avoid reusing it.
+     * It is OK if it fails (e.g. in headless mode).
+     */
+    xcb_parse_display(NULL, &xhost, &xdpy_current, NULL);
+
+    /*
+     * IANA assigns TCP port numbers from 6000 through 6063 to X11
+     * clients.  In addition, the current XCB implementaion shows
```
whot, line 70:
```suggestion
     * clients.  In addition, the current XCB implementation shows
```

meson.build
```diff
@@ -576,6 +576,8 @@ if get_option('enable-x11')
     libxkbcommon_x11_internal = static_library(
```
whot, line 576: fwiw, I'd appreciate it if you could rename this to `libxkbcommon_x11_test_internal` or something, I had to double-check `meson.build` to ensure this wasn't used anywhere else. Having it in the name makes it more obvious IMO

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,105 @@
...
+        }
+        snprintf(display, sizeof(display),
+                 "%s:%d", (xhost != NULL) ? xhost : "",
+                 xdpy_candidate);
+        ret = posix_spawnp(&xvfb_pid, "Xvfb", NULL, NULL, xvfb_argv, envp);
+        if (ret == 0) {
+            break;
+        }
+    }
+    free(xhost);
```
whot, line 88: this is probably nitpicking but: in what environment would a non-NULL `host` matter for the test? I think you could shortcut this to "if host is not NULL, go straight to `posix_spawn()`" because by definition Xvfb would run locally.
Though tbh I'm not sure if that would make the code any easier to read :)


Anyway, what *would* be useful is to use the `-displayfd` argument and have the server do the work of finding a free display. See `man Xserver` (it's not in `man Xvfb` because it's a generic option).

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,105 @@
...
+
+    if (ret != 0) {
+        ret = SKIP_TEST;
+        goto err_xvfd;
+    }
+
+    /* Wait for Xvfb fully waking up to accept a connection from a client. */
+    sleep(1);
```
whot, line 96: fun fact: the X server sends `SIGUSR1` when it's ready to accept connections (see `man Xserver`) and/or [`xserver/os/connection.c:InitParentProcess`](https://gitlab.freedesktop.org/xorg/xserver/-/blob/master/os/connection.c#L184)

test/x11.c
```diff
@@ -84,3 +85,7 @@ main(void)
 
     return exit_code;
 }
+
+int main(void) {
+    xvfb_wrapper(&test);
+}
```
whot, line 91: side-note: you could wrap this more generically by plonking the actual tests into a special ELF section and then have the *wrapper* loop through the sections to execute each test individually. This way the tests stay pristine with no knowledge on how they're invoked.

And you could easily duplicate the wrapper with a non-Xvfb version without having to touch the tests - just recompile with a different `wrapper.c`.

libei does this with the [util-munit.h]([https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/util-munit.h]) and [util-munit.c](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/util-munit.c) so that the actual invocation of the unit test (in the current binary) is a simple call to `run_all`, see [unit-tests.c](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/test/unit-tests.c)
Note that libei uses munit as test suite but those bits are quite trivially removed.

whot (commented):
Personally, the only thing that needs to be fixed is the typo, the rest is good enough for a test case. Some suggestions on improvements but at some point there's a valid effort vs gain argument...

test/xvfb-wrapper.c
wismill: I use `-displayfd`. It works but I get 3 times `_XSERVTransSocketUNIXCreateListener: ...SocketCreateListener() failed` on my machine (openSUSE Tumblweed, KDE Wayland), while nothing in using podman.

test/x11.c
wismill: I gave it a try. I am not sure this is much better.

meson.build

.github/workflows/linux.yml
```diff
@@ -31,6 +31,16 @@ jobs:
           sudo apt install -y \
             doxygen libxcb-xkb-dev valgrind ninja-build \
             libwayland-dev wayland-protocols bison graphviz
+      - name: Install xkeyboard-config
+        run: |
+          # HACK: Force xkeyboard-config version to master
```
bluetech, line 36: It is not clear if the hack is the fact that xkeyboard-config is forced to master, or how it's done. I understand it's the latter, but let's clarify.

bluetech (approved):
LGTM! It'd be nice to have the x11 not rely on a running x server.

bluetech:
> Question: should we use an option to disable the wrapper, in order to test with a local X11 server as previously?

I think it's not necessary.

.github/workflows/linux.yml
wismill: Fixed

test/xvfb-wrapper.c
whot: I think I have it working, we can do that as follow-up. See the top commit on [this branch here](https://github.com/whot/libxkbcommon/tree/wip/test/use-xvfb-for-x11-test), feel free to integrate/squash in as you see fit.

test/x11.c
whot: it's not going to look like much until you have to add other tests to it - that's when it'll pay off because there's no manually maintained test registry anymore.

whot (approved):
my only "concern" right now is the different coding style in the wrapper function compared to the rest of xkbcommon (which also introduces a potential uninitialized variable bug). Other than that, the usual nitpicks but this LGTM, thanks.

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,116 @@
...
+        rewind(display_fd);
+        fread(&display[1], 1, length, display_fd);
+        display[length] = '\0';
+    } else {
+        ret = SKIP_TEST;
+        goto err_xvfd;
+    }
+
+    /* Run the function requiring a running X server */
+    ret = f(display);
```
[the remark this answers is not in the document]
wismill: Fixed

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,116 @@
...
+    sleep(1);
+
+    /* Retrieve the display number: Xvfd writes the display number as a newline-
+     * terminated string; copy this number to form a proper display string. */
+    char display[6] = ":";
+    size_t length = ftell(display_fd);
+    if (length) {
+        rewind(display_fd);
+        fread(&display[1], 1, length, display_fd);
+        display[length] = '\0';
```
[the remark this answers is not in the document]
wismill: Done

test/xvfb-wrapper.c
wismill: Ok, I added a comment.

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,116 @@
...
+#include <assert.h>
+#include <signal.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+
+#include "test.h"
+#include "xvfb-wrapper.h"
+#include "xkbcommon/xkbcommon-x11.h"
+
+int xvfb_wrapper(int (*f)(char* display)) {
```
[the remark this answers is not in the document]
wismill: Done

test/xvfb-wrapper.c
wismill: I included your work

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,116 @@
...
+    /* File descriptor to retrieve the display number */
+    FILE * display_fd = tmpfile();
+    if (display_fd == NULL){
+        fprintf(stderr, "Unable to create temporary file.\n");
+        goto err_display_fd;
```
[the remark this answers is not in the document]
wismill: Done. Honestly, declaring variables close to their use site seems so much natural. I understand we want consistency in code style, but maybe it is time to allow modern coding style?

test/xvfb-wrapper.c
```diff
@@ -0,0 +1,116 @@
...
+
+    /* Run the function requiring a running X server */
+    ret = f(display);
+
+err_xvfd:
+    if (xvfb_pid > 0)
+        kill(xvfb_pid, SIGTERM);
+    fclose(display_fd);
+err_display_fd:
+    return ret;
```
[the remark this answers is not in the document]
wismill: Not sure of this: I get a warning if I set `display_fd` to `-1`. We surely do not want to depend on undefined behaviour.

test/xvfb-wrapper.c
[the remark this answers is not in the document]
bluetech: Fine by me!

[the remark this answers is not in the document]
wismill: Let’s discuss this in #360

wismill:
Rebased & improved commits. Review done and tests passed. Let’s merge!

PR #350, opened 2023-06-26
Test (compose): fix assumption on environment variables

### LINKED ISSUE #349: Test compose fails if there exists a user Compose file
`test/compose.c` fails on my machine because I have a `~/.XCompose` file. Indeed, the test fails using `xkb_compose_table_new_from_locale`, because this function looks for a compose file configured by the user before looking for `$XLOCALEDIR` or system files.

We should unset the corresponding environment variables or propose a flag to ignore them.

wismill:
The test `test_from_locale` fails when there is a user-defined compose file, e.g. ~/.XCompose.

Indeed, the function `xkb_compose_table_new_from_locale` use various environment variables to determine the location of the compose file.

Ensure these environment variables are initially explicitly unset, in order to have robust tests.

Fixes #349

### DIFF
--- a/test/compose.c
+++ b/test/compose.c
@@ -588,6 +588,22 @@ main(int argc, char *argv[])
     ctx = test_get_context(CONTEXT_NO_FLAG);
     assert(ctx);
 
+    /*
+     * Ensure no environment variables but “top_srcdir” is set. This ensures
+     * that user Compose file paths are unset before the tests and set
+     * explicitely when necessary.
+     */
+#ifdef __linux__
+    const char *srcdir = getenv("top_srcdir");
+    clearenv();
+    setenv("top_srcdir", srcdir, 1);
+#else
+    unsetenv("XCOMPOSEFILE");
+    unsetenv("XDG_CONFIG_HOME");
+    unsetenv("HOME");
+    unsetenv("XLOCALEDIR");
+#endif
+
     test_seqs(ctx);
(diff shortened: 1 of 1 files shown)


test/compose.c
```diff
@@ -588,6 +588,16 @@ main(int argc, char *argv[])
...
 
+    /*
+     * Ensure the environment variables that configure Compose file paths are
+     * unset before the tests.
+     * They are set explicitely in the corresponding tests when necessary.
+     */
+    unsetenv("XCOMPOSEFILE");
+    unsetenv("XDG_CONFIG_HOME");
+    unsetenv("HOME");
+    unsetenv("XLOCALEDIR");
```
whot, line 599: Stared at this and then thought it's a whack-a-mole. So I tested this hunk locally and behold, it works!

```c
    const char *srcdir = getenv("top_srcdir");
    clearenv();
    setenv("top_srcdir", srcdir, 1);
```
apparently the only env var we need here is `top_srcdir`.

whot (approved):
Surprisingly `clearenv()` works so I'd prefer that but yours works just as well

test/compose.c
wismill: Done

wismill: `clearenv` is not available on all platforms 😒

Fixed

whot (approved):

PR #352, opened 2023-06-29
Improve interactive tools

wismill:
I would like to revive #36 but I realize that I use a bunch of `printf` for debugging modifiers.

This MR add new options to `interactive-evdev` for debugging (see commits). Some options use the private API, so I added a non-installable version of the tool with this API.

I did not modify the other tools ATM, because `interactive-evdev` seem the most flexible one, being able to load whatever configuration without depending nor impacting the environment.

There is one commit that improve the display of control characters from key strokes translation.

### DIFF
--- a/tools/interactive-evdev.c
+++ b/tools/interactive-evdev.c
@@ -58,7 +58,12 @@ static int evdev_offset = 8;
 static bool report_state_changes;
 static bool with_compose;
 static enum xkb_consumed_mode consumed_mode = XKB_CONSUMED_MODE_XKB;
+
+#ifdef ENABLE_PRIVATE_APIS
+#define DEFAULT_PRINT_FIELDS (PRINT_ALL_FIELDS & ~PRINT_MODMAPS)
+#else
 #define DEFAULT_PRINT_FIELDS PRINT_ALL_FIELDS
+#endif
 print_state_fields_mask_t print_fields = DEFAULT_PRINT_FIELDS;
 
 #define DEFAULT_INCLUDE_PATH_PLACEHOLDER "__defaults__"
@@ -378,6 +383,9 @@ usage(FILE *fp, char *progname)
         fprintf(fp, "      or: %s --keymap <path to keymap file>\n",
                 progname);
         fprintf(fp, "For both:\n"
+#ifdef ENABLE_PRIVATE_APIS
+                        "          --print-modmaps (print real & virtual key modmaps)\n"
+#endif
                         "          --short (do not print layout nor Unicode keysym translation)\n"
                         "          --report-state-changes (report changes to the state)\n"
                         "          --enable-compose (enable Compose)\n"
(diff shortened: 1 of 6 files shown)


tools/interactive-evdev.c
```diff
@@ -499,6 +509,8 @@ main(int argc, char *argv[])
         {"consumed-mode",        required_argument,      0, OPT_CONSUMED_MODE},
         {"enable-compose",       no_argument,            0, OPT_COMPOSE},
         {"print-keycode",        no_argument,            0, OPT_PRINT_KEYCODE},
+        {"hide-layout",          no_argument,            0, OPT_HIDE_LAYOUT},
```
bluetech, line 459: I feel like adding these options is not ideal.

Is the problem that it becomes too long that wraps the terminal or something else?

tools/interactive-evdev.c
```diff
@@ -61,6 +61,8 @@ static int evdev_offset = 8;
 static bool report_state_changes;
 static bool with_compose;
 static bool print_keycode = false;
+print_state_fields_mask_t print_layout = PRINT_LAYOUT;
```
bluetech, line 61: These globals should be `static`

tools/tools-common.c
```diff
@@ -99,7 +100,15 @@ tools_print_keycode_state(struct xkb_state *state,
         xkb_compose_state_get_utf8(compose_state, s, sizeof(s));
     else
         xkb_state_key_get_utf8(state, keycode, s, sizeof(s));
-    printf("unicode [ %s ] ", s);
+    /* HACK: escape single control characters from C0 set using the
+     * Unicode codepoint convention. Ideally we would like to escape
+     * any non-printable character in the string.
+     */
+    if (strlen(s) == 1 && (*s <= 0x1F)) {
+        printf("unicode [ U+%04hX ] ", *s);
```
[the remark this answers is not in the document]
wismill: This is a standard notation. I agree every developer prefers the `\n` escape sequence, but `U+NNNN` is a _standard_ Unicode notation and gives good result in search engine. Anyway you will also have the keysym `Return` displayed: I think this is already a good hint.

tools/interactive-evdev.c
wismill: Yes, this is the problem.

tools/interactive-evdev.c
wismill: Good catch

tools/interactive-evdev.c
```diff
@@ -424,6 +441,7 @@ main(int argc, char *argv[])
         {"keymap",               required_argument,      0, OPT_KEYMAP},
         {"consumed-mode",        required_argument,      0, OPT_CONSUMED_MODE},
         {"enable-compose",       no_argument,            0, OPT_COMPOSE},
+        {"print-keycode",        no_argument,            0, OPT_PRINT_KEYCODE},
```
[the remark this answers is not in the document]
wismill: My thought exactly. But I wanted to keep previous defaults. Done

tools/interactive-evdev.c
bluetech: The "hide" approach is not very scalable. I think we should either go all-in and allow configuring the exact set of fields generically, or go the other way and only add a `--short` option which makes the output shorted in an unspecified manner. Personally I'd go with `--short`.

wismill: Goo idea. Done.

wismill:
Rebased. This PR requires squashing some commits. Will get to it later.

wismill:
Rebased & squashed.

PR #353, opened 2023-06-30
Support querying whether virtual modifiers are active (alternative with overlapping mods API)

wismill:
I would like to revive #36. I found some issues (see #36), so I used another implementation. It only lifts the restriction to _real_ modifiers of some functions.

- [ ] Doc to warn of possible over matching.
- [ ] I am not satisfied with the  `xkb_keymap_mod_get_overlapping_mods` API. It exposes the internal mapping of modifiers, but I use the term _canonical_ to refer to _real_ modifiers: having modifiers defined as a set of other modifiers does make sense, even if we drop the limitations of X11 real modifiers.


bam80:
Thanks for looking into this @wismill. I'm not ready to do any review though, sorry.

src/state.c
```diff
@@ -1138,7 +1138,14 @@ xkb_state_mod_index_is_active(struct xkb_state *state,
     if (idx >= xkb_keymap_num_mods(state->keymap))
         return -1;
 
-    return !!(xkb_state_serialize_mods(state, type) & (1u << idx));
+    if (is_real_mod(state->keymap, idx)) {
+        return !!(xkb_state_serialize_mods(state, type) & (1u << idx));
+    } else {
+        /* WARNING: this may over matches */
```
whot, line 1144: "overmatch" (same in commit message please)

src/state.c
```diff
@@ -1138,7 +1138,14 @@ xkb_state_mod_index_is_active(struct xkb_state *state,
...
+        return !!((xkb_state_serialize_mods(state, type) &
+                   state->keymap->mods.mods[idx].mapping) ==
+                  state->keymap->mods.mods[idx].mapping);
```
whot, line 1147: at the risk of sounding stupid: can we do this with a temp variable please - with the linebreaks and parens it gets a bit hard to parse:
```c
   something_t mapping = state->keymap->mods[idx].mapping;
   return !!((xkb_state_serial_mods(state, type) & mapping) == mapping);
```

src/state.c
```diff
@@ -1237,7 +1246,9 @@ xkb_state_mod_names_are_active(struct xkb_state *state,
             ret = -1;
             break;
         }
-        wanted |= (1u << idx);
+        wanted |= is_real_mod(state->keymap, idx)
+            ? (1u << idx)
+            : state->keymap->mods.mods[idx].mapping;
```
whot, line 1251: this screams for a little helper function where the name explains what it does

include/xkbcommon/xkbcommon-names.h
```diff
@@ -34,9 +34,13 @@
 #define XKB_MOD_NAME_SHIFT      "Shift"
 #define XKB_MOD_NAME_CAPS       "Lock"
 #define XKB_MOD_NAME_CTRL       "Control"
-#define XKB_MOD_NAME_ALT        "Mod1"
-#define XKB_MOD_NAME_NUM        "Mod2"
-#define XKB_MOD_NAME_LOGO       "Mod4"
+#define XKB_MOD_NAME_ALT        "Alt"
+#define XKB_MOD_NAME_META       "Meta"
+#define XKB_MOD_NAME_LOGO       "Super"
+#define XKB_MOD_NAME_GUI        "Super"
```
whot, line 40: Where does the GUI name come from? I don't think I've ever seen that one before

include/xkbcommon/xkbcommon-names.h
```diff
@@ -34,9 +34,13 @@
...
+#define XKB_MOD_NAME_SUPER      "Super"
+#define XKB_MOD_NAME_HYPER      "Hyper"
+#define XKB_MOD_NAME_NUM        "NumLock"
```
whot, line 43: hmm, I wonder if that and the `ALT`/`LOGO` changes are an API/ABI break...

include/xkbcommon/xkbcommon.h
```diff
@@ -1091,6 +1091,37 @@ xkb_keymap_mod_get_name(struct xkb_keymap *keymap, xkb_mod_index_t idx);
...
+ * returns 0.
+ *
+ * @sa xkb_mod_mask_t
+ * @memberof xkb_keymap
+ * @since 1.6.0
+ */
+xkb_mod_mask_t
+xkb_keymap_mod_get_overlapping_mods(struct xkb_keymap *keymap,
+                                    xkb_mod_index_t idx,
+                                    enum xkb_mod_overlapping_type);
```
whot, line 1123:
```suggestion
                                    enum xkb_mod_overlapping_type type);
```

whot (commented):
a few nitpicks but this LGTM though I fully admit I'm not aware enough of the API to really make decisions on this.

I would like to see the evdev define commit as a separate PR though because that is one that we can merge at any time and it's almost independent of this one here.

src/state.c
wismill: Done

src/state.c
wismill: Done

include/xkbcommon/xkbcommon-names.h
wismill: I thought I’ve seen it in Wikipedia or W3C’s [“UI Events KeyboardEvent key Values”](https://www.w3.org/TR/uievents-key/#keys-modifier), but no 😅. There is also “OS”, but it is not much better than “Logo”.

I like “GUI” because this modifier is usually used for GUI shortcuts. But this may not always be the case, so let’s stick to previous and explicit modifier names.

include/xkbcommon/xkbcommon-names.h
wismill: They may be breaking changes in the case that the modifier mapping is different.

For `Alt`, I found only 2 vendor layouts in xkeyboard-config which does not map to `Mod1`:
- `symbols/fujitsu_vndr/us`
- `symbols/hp_vndr/us`

For `NumLock`, I found only 2 vendor layouts in xkeyboard-config which does not map to `Mod2`:
- `symbols/fujitsu_vndr/us`
- `symbols/digital_vndr/pc`

I did not find other mapping for `Super` than `Mod4`.

The other possible issue is when these modifiers are not even mapped! But with the base layout `pc` they should be mapped by default.

include/xkbcommon/xkbcommon.h
wismill: Fixed

wismill:
See #512 for a smaller version of this PR.

PR #357, opened 2023-07-03
Fix numeric keysym parsing

wismill:
While working on #332, I realized that numeric keysyms parsing has some issues. Let’s deal with it in this separate PR.

This PR fixes some edges cases and improve tests.

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -304,6 +304,15 @@ typedef uint32_t xkb_led_mask_t;
 #define XKB_KEYCODE_MAX     (0xffffffff - 1)
 
 /**
+ * Maximum keysym value
+ *
+ * @since 1.6.0
+ * @sa xkb_keysym_t
+ * @ingroup keysyms
+ */
+#define XKB_KEYSYM_MAX      0x1fffffff
+
+/**
  * Test whether a value is a valid extended keycode.
  * @sa xkb_keycode_t
  **/
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -169,28 +169,37 @@ typedef uint32_t xkb_keycode_t;
  *
  * A key, represented by a keycode, may generate different symbols according
  * to keyboard state.  For example, on a QWERTY keyboard, pressing the key
- * labled \<A\> generates the symbol 'a'.  If the Shift key is held, it
- * generates the symbol 'A'.  If a different layout is used, say Greek,
- * it generates the symbol 'α'.  And so on.
+ * labled \<A\> generates the symbol ‘a’.  If the Shift key is held, it
+ * generates the symbol ‘A’.  If a different layout is used, say Greek,
+ * it generates the symbol ‘α’.  And so on.
  *
- * Each such symbol is represented by a keysym.  Note that keysyms are
- * somewhat more general, in that they can also represent some "function",
- * such as "Left" or "Right" for the arrow keys.  For more information,
- * see:
- * https://www.x.org/releases/current/doc/xproto/x11protocol.html#keysym_encoding
+ * Each such symbol is represented by a *keysym* (short for “key symbol”).
+ * Note that keysyms are somewhat more general, in that they can also represent
+ * some “function”, such as “Left” or “Right” for the arrow keys.  For more
+ * information, see: Appendix A [“KEYSYM Encoding”][encoding] of the X Window
+ * System Protocol.
  *
  * Specifically named keysyms can be found in the
  * xkbcommon/xkbcommon-keysyms.h header file.  Their name does not include
- * the XKB_KEY_ prefix.
+ * the `XKB_KEY_` prefix.
  *
- * Besides those, any Unicode/ISO 10646 character in the range U0100 to
- * U10FFFF can be represented by a keysym value in the range 0x01000100 to
- * 0x0110FFFF.  The name of Unicode keysyms is "U<codepoint>", e.g. "UA1B2".
+ * Besides those, any Unicode/ISO&nbsp;10646 character in the range U+0100 to
+ * U+10FFFF can be represented by a keysym value in the range 0x01000100 to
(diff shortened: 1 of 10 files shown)


include/xkbcommon/xkbcommon-keysyms.h
```diff
@@ -3250,6 +3250,11 @@ performance, or use of this material.
 #define XKB_KEY_longminus            0x100000f6
 #define XKB_KEY_block                0x100000fc
 
+/***********************************************************
+ * Keysyms are 32-bit integers with the 3 most significant bits always set
+ * to zero. See: Appendix A “KEYSYM Encoding” of the X Window System Protocol.
+ */
+#define XKB_MAX_KEYSYM               0x1fffffff
```
bluetech, line 3257: This define should go in `xkbcommon.h`, like the other MAX defines. For consistency, it should be called `XKB_KEYSYM_MAX`.

src/keysym.c
```diff
@@ -64,7 +64,7 @@ get_name(const struct name_keysym *entry)
 XKB_EXPORT int
 xkb_keysym_get_name(xkb_keysym_t ks, char *buffer, size_t size)
 {
-    if ((ks & ((unsigned long) ~0x1fffffff)) != 0) {
+    if ((ks & ((unsigned long) ~XKB_MAX_KEYSYM)) != 0) {
```
bluetech, line 67: This looks a bit weird, e.g. if the max was 535353 then it wouldn't have made sense to apply `~` to it. With a MAX constant it's clearer to do `if (ks > XKB_MAX_KEYSYM) { ... }`.

src/xkbcomp/expr.c
```diff
@@ -656,11 +656,25 @@ ExprResolveKeySym(struct xkb_context *ctx, const ExprDef *expr,
     if (!ExprResolveInteger(ctx, expr, &val))
         return false;
 
-    if (val < 0 || val >= 10)
```
bluetech, line 659: This looks like it regressed in 2e4933cda7c141c9efc8c62e7d72f591f2998836 compared to xkbcomp...

test/keysym.c
```diff
@@ -27,6 +27,9 @@
 #include "test.h"
 #include "keysym.h" /* For unexported is_lower/upper/keypad() */
 
+#define STRINGIFY_(x) #x
```
bluetech, line 30: Maybe move this to utils.h?

Also, maybe change `STRINGIFY_` -> `STRINGIFY` and `STRINGIFY` -> `STRINGIFY2`? Sometimes you don't want to expand (e.g. the one in `x11/keymap.c` doesn't.

bluetech (commented):
Good fixes!

I left some comments. Also, do you mind splitting the `Fix parsing of numeric keysyms` commit such that each fix is in its own commit? It will be easier to bisect etc. this way.

include/xkbcommon/xkbcommon-keysyms.h
wismill: Done

src/keysym.c
wismill: Yes. I went the lazy road to keep diff minimal.

test/keysym.c
wismill: Done

wismill:
@bluetech Commit split!

bluetech (approved):
Thanks!

include/xkbcommon/xkbcommon.h
```diff
@@ -303,6 +303,14 @@ typedef uint32_t xkb_led_mask_t;
 
 #define XKB_KEYCODE_MAX     (0xffffffff - 1)
 
+/**
+ * Keysyms are 32-bit integers with the 3 most significant bits always set
+ * to zero. See: Appendix A “KEYSYM Encoding” of the X Window System Protocol.
+ *
+ * @since 1.6.0
+ */
+#define XKB_KEYSYM_MAX      (0x1fffffff)
```
whot, line 312: aside from consistency with the ones above but - why is this in parenthesis? 

`XKB_KEYCODE_MAX` needs it (though it would be trivial to just specify it as the actual numeric value instead with a comment for the humand-readable one but here? I'm struggling to think of where the parens make a difference.

test/keysym.c
```diff
@@ -27,6 +27,9 @@
 #include "test.h"
 #include "keysym.h" /* For unexported is_lower/upper/keypad() */
 
+/* Needed because XKB_KEYSYM_MAX has parentheses */
+#define KEYSYM_MAX 0x1fffffff
```
whot, line 31: tbh I think it'd be better off using the actual strings in the tests instead of having a separate `#define` here. e.g
```c
    assert(test_string("0x1fffffff", XKB_KEYSYM_MAX));
```
is immediately understandable without the extra redirection.

whot (approved):
two nitpicks but this LGTM. though my eyes glaze over every time i look at the actual parser code :)

include/xkbcommon/xkbcommon.h
wismill: I used parentheses just for consistency with the other ones. Since the new constants has no special requirement, let’s drop them.

test/keysym.c
wismill: I dropped the parentheses so we can use XKB_KEYSYM_MAX for clarity.

wismill:
Rebased.

I added the constant `XKB_KEYSYM_MIN`. `xkb_keysym_t` is unsigned, but we do need to check min keysym in parser. Then we have the couple min/max.

I also added much more test related to keysyms, to reflect the current implementation.

include/xkbcommon/xkbcommon.h
```diff
@@ -303,6 +312,24 @@ typedef uint32_t xkb_led_mask_t;
...
 #define XKB_KEYCODE_MAX     (0xffffffff - 1)
 
+/**
+ * Minimum keysym value
+ *
+ * @since 1.6.0
+ * @sa XKB_KEYSYM_MAX, xkb_keysym_t
+ * @ingroup keysyms
+ */
+#define XKB_KEYSYM_MIN      0x00000000
```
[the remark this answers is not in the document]
wismill: Good point

Issue #37, opened 2016-06-23
Remove the 4 layout limit on Wayland.


ApostolosB:
Can something be done about it. I know on X it is a hard limit but can libxkbcommon be modified for allowing more than 4 keyboards on WL?

Thanks in advance.

bluetech:
This is definitely desirable. For the current status, see the comment here: https://github.com/xkbcommon/libxkbcommon/blob/d58fc90a2917e482470e74a662373f85ccb23f31/src/keymap.h#L91

I provide some additional details here on the problem, in case someone other than me reads this and wants to jump in.

First, the format of the "rules" file is described here: https://github.com/xkbcommon/libxkbcommon/blob/master/doc/rules-format.txt

Currently, the most used file is the "evdev" one, commonly found in `/usr/share/X11/xkb/rules/evdev`. This file is built up using some scripts as can be seen here: https://cgit.freedesktop.org/xkeyboard-config/tree/rules

The problem is that the format is really inflexible about the number of layouts (groups), and so it hardcodes and repeats 1,2,3,4. Hence, if we just lift the limit, some unexpected things may start to happen starting from the 5th group.

Some ways to proceed, from most preferable to least preferable:

- Fix up the rules format to not need to hardcode to group numbers. The linked bug (https://bugs.freedesktop.org/show_bug.cgi?id=14372) suggests an `:all` syntax, which will work transparently for all groups. This will require changes in both of xkbcommon and xkbcomp's implementations. Also, if the change is not backward compatible, there might be some transitionary issues (a possible workaround is to use a new file instead of "evdev").

- Keep hardcoding the numbers, but hardcode higher, e.g. to 8. This might also need fixing to work; I didn't try.

- Just make the change to xkbcommon, disregarding the above issues. For some scenarios, it will work just fine, I expect (didn't try either). Testing here will be appreciated.

That's it - needs a small effort to push through.

wismill:
#362 is a first step to support it in rules.

wismill:
#362 has been merged.

PR #362, opened 2023-07-13
Modern rules

wismill:
This is a replacement of #334.

There is a lot of repetition in the current rules files, because the MLVO mappings need to match on the exact layout/variant index. This also prevents to increase the number of maximum groups.

We introduces new special layout/variant indexes:
- “single”: matches a single layout; same as with no index.
- “first”: matches the first layout/variant, no matter how many layouts are in the RMLVO configuration.
- “later”: matches all but the first layout. This is an index range.
- “any”: matches layout at any position. This is an index range.

We also introduces the new `%i` expansion, which correspond to the index of the matched layout in a mapping with an index range. Example:

    layout[later] = symbols
    my_layout     = +my_symbols:%i
    *             = +%l[%i]:%i

Let’s have a look at concrete examples from xkeyboard-config:

    ! model     layout     = symbols
      *         *          = pc+%l%(v)
    ! model     layout[1]  = symbols
      *         *          = pc+%l[1]%(v[1])
    ! model     layout[2]  = symbols
      *         *          = +%l[2]%(v[2])
    ! model     layout[3]  = symbols
      *         *          = +%l[3]%(v[3])
    ! model     layout[4]  = symbols
      *         *          = +%l[4]%(v[4])

    ! layout    option     = symbols
      *         grp:toggle = +group(toggle)
    ! layout[1] option     = symbols
      *         grp:toggle = +group(toggle):1
    ! layout[2] option     = symbols
      *         grp:toggle = +group(toggle):2
    ! layout[3] option     = symbols
      *         grp:toggle = +group(toggle):3
    ! layout[4] option     = symbols
      *         grp:toggle = +group(toggle):4

With this commit we can now simplify it into:

    ! model    layout[first] = symbols
      *        *             = pc+%l[%i]%(v[%i])
    ! model    layout[later] = symbols
      *        *             = +%l[%i]%(v[%i]):%i

    ! layout[any] option     = symbols
      *           grp:toggle = +group(toggle):%i

The latter rule will work even if we increase XKB_MAX_GROUPS, whereas
the former would require to add the missing entries for the new groups.

In order to maintain consistent rules, we now disallow layout and variant to have different indexes. For example, the following mapping are now invalid:
- `layout variant[1]`
- `layout[1] variant[2]`
- `layout[1] variant`
- `layout[first] variant[1]`
- `layout[first] variant`
- `layout variant[any]`
- etc.

Some layout options require to be applied to every group to maintain consistency (e.g. a group switcher). Currently this must be done manually for all layout indexes. This is error prone and prevents the increase of the maximum group count.

We introduces the `:all` qualifier for `kccgst` values. When a rule with this qualifier applies, it will expands for every layout, e.g. `+group(toggle):all` would expand to `+group(toggle):1+group(toggle):2` if there are 2 layouts, etc.

__TODO:__
- [x] Update `docs/rules-format.md`
- [x] Add log entry
- [ ] Propose changes to `xkeyboard-config`

Related: #37 and #311


wismill:
@whot @bluetech  Rebased and improved the code. There are still redundancies, but for now I would like some feedback before to do further optimization.

Once the POC is validated, I can start to implement it into `libxkbfile`.

TODO: documentation

src/xkbcomp/rules.c
```diff
@@ -988,8 +988,45 @@ expand_rmlvo_in_kccgst_value(struct matcher *m, struct scanner *s,
 }
 
 /*
- * This function performs %-expansion on @value (see overview above),
- * and appends the result to @to.
+ * This function performs :all replacement on @value (see overview above),
+ * and appends the result to @expanded.
+ */
+static bool
```
bluetech, line 994: Since the `bool` return value is not used, would remove it.

src/xkbcomp/rules.c
```diff
@@ -988,8 +988,45 @@ expand_rmlvo_in_kccgst_value(struct matcher *m, struct scanner *s,
...
+expand_qualifier_in_kccgst_value(
+    struct matcher *m, struct scanner *s,
+    struct sval value, darray_char *expanded, unsigned int *i, unsigned int last_item_idx)
+{
+    const char *str = value.start;
+
+    if (*i + 2 >= value.len || str[*i] != 'a' || str[*i+1] != 'l' || str[*i+2] != 'l') {
```
bluetech, line 1001: Will be cleaner to turn this to a positive test and nest IMO. Then if there's another qualifier (theoretically) it will work the same.

src/xkbcomp/rules.c
```diff
@@ -988,8 +988,45 @@ expand_rmlvo_in_kccgst_value(struct matcher *m, struct scanner *s,
...
+        /* Not :all qualifier */
+        return true;
+    }
+    /* Found :all, check leftovers until the end of the qualifier */
+    if (*i + 3 <= value.len || str[*i + 3] == '+' || str[*i + 3] == '|') {
+        /* Add at least one layout */
```
bluetech, line 1007: Why?

bluetech (commented):
Awesome work @wismill. Now with X starting to go the way of the dodo, the fact that we're still stuck on max 4 layouts is a bit embarrassing. And the repetition in the rules format is even more embarrassing :)

I've read through the code, it looks good (I will write some nits later), but let's finalize the design first.

---

Regarding `first`, `later`, `any`.

I already forgot why the format insists on distinguishing `foo` (matches first only if single layout given) and `foo[1]` (matches first only if multiple layouts given). Is there an honest semantic reason for this, or just some historic weirdness?

I think what would help me quite a bit is to see `rules/evdev` before/after (if it's not too much effort). Then I can see more directly why each new addition is needed.

---

Idle thought: maybe instead of `first`, `later`, `any` we can use python slice syntax, i.e. `layout[2:]`, `layout[:]`. I realize the `first` case is complicated (see previous question).

---

`%i` SGTM, I think we can agree on this design.

---

`:all` SGTM, I think we can agree on this design. I left some nits on the commit. 

---

`docs/rules-format.md` will need an update, can do it later though.

---

> In order to maintain consistent rules, we now disallow layout and
variant to have different indexes.

This part LGTM, you can split & cherry-pick it to master already if you'd like unless you want to wait for libxkbfile to agree to this as well.

src/xkbcomp/rules.c
wismill: Done

src/xkbcomp/rules.c
wismill: This checks we have _exactly_ `all`. I refactored wrt to your other comment about positive test.

wismill:
@bluetech Thanks! I refactored based on your review. I also introduced `MAX_LAYOUT_INDEX_STR_LENGTH` to be more explicit.

> Regarding `first`, `later`, `any`.
> 
> I already forgot why the format insists on distinguishing `foo` (matches first only if single layout given) and `foo[1]` (matches first only if multiple layouts given). Is there an honest semantic reason for this, or just some historic weirdness?

I think this is due to some config model/layouts adding a second group, e.g:

```
! model		layout			=	symbols
  ataritt	$nonlatin		=	xfree68_vndr/ataritt(us)+%l%(v):2
```

Now I think it makes sense to add e.g. `layout[single]` to make it more explicit. What do you think?

> I think what would help me quite a bit is to see `rules/evdev` before/after (if it's not too much effort). Then I can see more directly why each new addition is needed.

I think this transformation would make sense even without this implementation, in order to simplify the rules. Let me take time to propose a MR in `xkeyboard-config`. Meanwhile, please have a look at the examples above.

> Idle thought: maybe instead of `first`, `later`, `any` we can use python slice syntax, i.e. `layout[2:]`, `layout[:]`. I realize the `first` case is complicated (see previous question).

I thought exactly about that! But then: what does it bring to us? Do we need e.g. `[3:]` or `[1:3]` or `[:2]`? If not, then I think 

> `docs/rules-format.md` will need an update, can do it later though.

Yes, it should be done in this MR. Let’s add a TODO.

> > In order to maintain consistent rules, we now disallow layout and
> > variant to have different indexes.
> 
> This part LGTM, you can split & cherry-pick it to master already if you'd like unless you want to wait for libxkbfile to agree to this as well.

I do not think we need to wait for libxkbfile.

wismill:
Rebased, added `[single]` special index and refactored a bit.

wismill:
Fixed some edge cases and ensure we are prepared to raise the group limit.

PR #364, opened 2023-07-21
Keysyms: improve generator

wismill:
Simplify evdev and XK_ substitution and improve alignment. Also, perform some additional XK_ substitutions in comments.

Motivation: see [MR in xorgproto](https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/merge_requests/80#note_2010176)


bluetech (approved):
I generated the before/after and I see the improvement, thanks.

bluetech:
@wismill I wanted to ask, when you open a PR as draft do you still want reviews or should we wait until it's not a draft?

wismill:
> @wismill I wanted to ask, when you open a PR as draft do you still want reviews or should we wait until it's not a draft?

Good question! My take on draft PR is: “I inform you that I work on this topic; the code is sufficiently advanced to open an MR but not quite finished; _you are welcome to review it_ but wait that I finish the draft if you do not have much time”.

scripts/makeheader
```diff
@@ -3,9 +3,41 @@ from __future__ import print_function
...
+
+def make_keysym_entry(m: re.Match[str]) -> str:
+    '''
+    Perform the substitutions
+    '''
+    if m.group('evdev'):
+        if m.group('define').startswith('#'):
+            # Replace the xorgproto _EVDEVK macro with the actual value
+            # 0x10081000 is the base, the evdev hex code is added to that.
+            # We replace to make parsing of the keys later easier.
```
whot, line 32: fwiw, this comment is now confusing since we're no longer replacing the line

scripts/makeheader
```diff
@@ -3,9 +3,41 @@ from __future__ import print_function
...
+            value = 0x10081000 + int(m.group('value'), 16)
+            value_str = f'{value:#x}    '
+        else:
+            value_str = f'''_EVDEVK({m.group('value')})'''
+    else:
+        value_str = m.group('value')
+    return f'''{m.group('define')} XKB_KEY_{m.group('prefix') or ''}{m.group('name')}{m.group('spacing')}{value_str}'''
```
whot: I think this may be a bit more readable if we move do a `foo = m.group('foo')` and then use those variables instead. A few more lines but makes this particular format easier to read

scripts/makeheader
```diff
@@ -3,9 +3,41 @@ from __future__ import print_function
...
+keysym_entry_pattern = re.compile(
+    r'''^
+    (?P<define>\#define|/\*\s+Use:)\s+
+    (?P<prefix>\w*)XK_(?P<name>\w+)(?P<spacing>\s+)
+    (?P<evdev>_EVDEVK\()?(?P<value>0x[0-9A-Fa-f]+)(?(evdev)\))
+    ''',
+    re.VERBOSE,
+)
+# Match remaining XK_ references
+xorgproto_keysym_prefix_pattern = re.compile(r'\b(?P<prefix>\w*)XK_(?!KOREAN\b)')
```
whot: got confused by this and had to check the various header files, would be nice to add something like:
```
# XK_KOREAN has ifdefs, ensure we handle those too
```

whot (approved):
a few minor nitpicks but this is a definitive improvement. would be nice to hook up black in the CI [1] to ensure this doesn't accidentally get different formatting in the future.

[1] possibly via pre-commit.ci

scripts/makeheader
wismill: Sorry, I do not understand:  we _do_ replace the macro when it is not commented.

scripts/makeheader
wismill: Done

scripts/makeheader
wismill: It is confusing, indeed! This pattern is intended to match remaining keysyms references in _comments_, e.g. replace:

```
#define XKB_KEY_SunCompose		0x0000FF20	/* Same as XK_Multi_key */
```

with:

```
#define XKB_KEY_SunCompose		0x0000FF20	/* Same as XKB_KEY_Multi_key */
```

I updated the comment.

Issue #361, opened 2023-07-07
Determining required dead keys


johngebbie:
Hello, I'm not sure how to programatically determine the required dead keys for a keysym. For example, I don't know how you'd get that you need to type dead_circumflex and e to type ecircumflex in the fr layout.

how-to-type doesn't seem to implement it, with this outputting no entries:
```
xkbcli how-to-type --layout fr --keysym ecircumflex
```

I think ecircumflex could be found under a level with `xkb_keymap_key_get_syms_by_level`, but (if that's even right) I'm not sure how to map the levels to the required dead keys.

My end goal is to have my uinput tool support typing symbols that require dead keys.

wismill:
Currently the tool `how-to-type` does not support _Compose_. It means it will _not_ look for keysyms that are the result of e.g. a sequence consisting of a dead key and a letter, such as: `<dead_circumflex> <e>`.

Unfortunately it does not look that there is a [public API](https://xkbcommon.org/doc/current/group__compose.html) to get a sequence from its result.

johngebbie:
Thank you, it's good to know I'm not just not seeing it. Shall I make this a feature request?

PR #367, opened 2023-07-25
Add iterator API for Compose table

### LINKED ISSUE #366: Add iteration API for compose
Hi! I'm a kbd project maintainer. I want to add xkbcommon support and generate kernel keymaps on fly. I made a proof-of-concept[1] a while back and it works quite well, but I ran into a problem with libxkbcommon. There is no way in the libxkbcommon to walkthrough a xkb_compose_table. I need this to load the compose table into the kernel (it's Just the sequences and the results). The Kernel API allows data to be loaded into kernelspace and then handles the input with everything itself.

In early 2022 I emailed Ran Benita about this issue and he even did an implementation [2], but unfortunately did not complete it.

Can you make something like xkb_keymap_key_for_each but for xkb_compose_table or maybe some other way to examine the table ?

[1] https://github.com/legionus/kbd/issues/72
[2] https://github.com/xkbcommon/libxkbcommon/compare/master...bluetech:libxkbcommon:compose-traverse

wismill:
This is a revival of the [`compose-traverse` branch](https://github.com/bluetech/libxkbcommon/tree/compose-traverse) from @bluetech.

Fixes #366
Could be also useful for #361.

### DIFF
--- a/include/xkbcommon/xkbcommon-compose.h
+++ b/include/xkbcommon/xkbcommon-compose.h
@@ -299,6 +299,143 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
 void
 xkb_compose_table_unref(struct xkb_compose_table *table);
 
+/**
+ * @struct xkb_compose_table_entry
+ * Opaque Compose table entry object.
+ *
+ * Represents a single entry in a Compose file in the iteration API.
+ * It is immutable.
+ *
+ * @sa xkb_compose_table_iterator_new
+ * @since 1.6.0
+ */
+struct xkb_compose_table_entry;
+
+/**
+ * Get the left-hand keysym sequence of a Compose table entry.
+ *
+ * For example, given the following entry:
+ *
+ * ```
+ * <dead_tilde> <space> : "~" asciitilde # TILDE
+ * ```
+ *
+ * it will return `{XKB_KEY_dead_tilde, XKB_KEY_space}`.
+ *
+ * @param[in]  entry The compose table entry object to process.
+ *
+ * @param[out] sequence_length Number of keysyms in the sequence.
+ *
+ * @returns The array of left-hand side keysyms.  The number of keysyms
+ * is returned in the @p sequence_length out-parameter.
+ *
+ * @memberof xkb_compose_table_entry
+ * @since 1.6.0
+ */
+const xkb_keysym_t *
+xkb_compose_table_entry_sequence(struct xkb_compose_table_entry *entry,
+                                 size_t *sequence_length);
+
+/**
+ * Get the right-hand result keysym of a Compose table entry.
+ *
+ * For example, given the following entry:
+ *
+ * ```
+ * <dead_tilde> <space> : "~" asciitilde # TILDE
+ * ```
+ *
+ * it will return `XKB_KEY_asciitilde`.
+ *
+ * The keysym is optional; if the entry does not specify a keysym,
+ * returns `XKB_KEY_NoSymbol`.
+ *
+ * @memberof xkb_compose_table_entry
+ * @since 1.6.0
+ */
+xkb_keysym_t
+xkb_compose_table_entry_keysym(struct xkb_compose_table_entry *entry);
+
+/**
+ * Get the right-hand result string of a Compose table entry.
+ *
+ * The string is UTF-8 encoded and NULL-terminated.
+ *
+ * For example, given the following entry:
+ *
+ * ```
+ * <dead_tilde> <space> : "~" asciitilde # TILDE
+ * ```
+ *
+ * it will return `"~"`.
+ *
+ * The string is optional; if the entry does not specify a string,
+ * returns the empty string.
+ *
+ * @memberof xkb_compose_table_entry
+ * @since 1.6.0
+ */
+const char *
+xkb_compose_table_entry_utf8(struct xkb_compose_table_entry *entry);
+
+/**
+ * @struct xkb_compose_table_iterator
+ * Iterator over a compose table’s entries.
+ *
+ * @sa xkb_compose_table_iterator_new()
+ * @since 1.6.0
+ */
+struct xkb_compose_table_iterator;
+
+/**
+ * Create a new iterator for a compose table.
+ *
+ * Intended use:
+ *
+ * ```c
+ * struct xkb_compose_table_iterator *iter = xkb_compose_table_iterator_new(compose_table);
+ * struct xkb_compose_table_entry *entry;
+ * while ((entry = xkb_compose_table_iterator_next(iter))) {
+ *     // ...
+ * }
+ * xkb_compose_table_iterator_free(iter);
+ * ```
+ *
+ * @returns A new compose table iterator, or `NULL` on failure.
+ *
+ * @memberof xkb_compose_table_iterator
+ * @sa xkb_compose_table_iterator_free()
+ * @since 1.6.0
+ */
+struct xkb_compose_table_iterator *
+xkb_compose_table_iterator_new(struct xkb_compose_table *table);
+
+/**
+ * Free a compose iterator.
+ *
+ * @memberof xkb_compose_table_iterator
+ * @since 1.6.0
+ */
+void
+xkb_compose_table_iterator_free(struct xkb_compose_table_iterator *iter);
+
+/**
+ * Get the next compose entry from a compose table iterator.
+ *
+ * The entries are returned in lexicographic order of the left-hand
+ * side of entries. This does not correspond to the order in which
+ * the entries appear in the Compose file.
+ *
+ * @attention The return value is valid until the next call to this function.
+ *
+ * Returns `NULL` in case there is no more entries.
+ *
+ * @memberof xkb_compose_table_iterator
+ * @since 1.6.0
+ */
+struct xkb_compose_table_entry *
+xkb_compose_table_iterator_next(struct xkb_compose_table_iterator *iter);
+
 /** Flags for compose state creation. */
 enum xkb_compose_state_flags {
     /** Do not apply any flags. */
--- a/src/compose/table.c
+++ b/src/compose/table.c
@@ -27,6 +27,7 @@
 #include "table.h"
 #include "parser.h"
 #include "paths.h"
+#include "xkbcommon/xkbcommon.h"
 
 static struct xkb_compose_table *
 xkb_compose_table_new(struct xkb_context *ctx,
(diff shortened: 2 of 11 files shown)


include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+
+/**
+ * Create a new iterator for a compose table.
+ *
+ * @memberof xkb_compose_table
+ * @sa xkb_compose_table_iterator_unref()
+ * @since 1.6.0
+ */
+struct xkb_compose_table_iterator *
+xkb_compose_table_iterator_new(struct xkb_compose_table *table);
```
bluetech, line 394: I'm thinking that if we ever want to modify this, e.g. add support for preserving the original entries order, then we should add a flags argument to future proof it. BUT, I think that's unlikely enough that it's not worth it. We would just add a `xkb_compose_table_iterator_new2` function.

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+xkb_compose_table_iterator_current(struct xkb_compose_table_iterator *iter);
+
+/**
+ * Get the next compose entry from a compose table iterator.
+ *
+ * The entries are returned in lexicographic order of the left-hand
+ * side of entries. This does not correspond to the order in which
+ * the entries appear in the Compose file.
+ *
+ * Returns NULL in case there is no more entries.
```
bluetech, line 421: It's unclear what `xkb_compose_table_iterator_current()` return when we reach the end. (the possibilities are: the last entry, or NULL).

tools/compose.c
```diff
@@ -29,34 +29,67 @@
...
     fprintf(fp,
+            "   --file - specify a file to load\n"
             "   --locale - specify the locale directly\n"
             "   --locale-from-env - get the locale from the LC_ALL/LC_CTYPE/LANG environment variables (falling back to C)\n"
             "   --locale-from-setlocale - get the locale using setlocale(3)\n"
     );
 }
 
+static void
+compose_table_print_cb(struct xkb_compose_table_entry *entry, void *data)
```
bluetech, line 50:
```suggestion
static void
print_compose_table_entry(const struct xkb_compose_table_entry *entry)
```

xkbcommon.map
```diff
@@ -109,3 +109,14 @@ global:
     xkb_utf32_to_keysym;
     xkb_keymap_key_get_mods_for_level;
 } V_0.8.0;
+
+V_1.6.0 {
+global:
+	xkb_compose_table_entry_sequence;
+	xkb_compose_table_entry_keysym;
+	xkb_compose_table_entry_utf8;
```
bluetech, line 117:
```suggestion
    xkb_compose_table_entry_sequence;
    xkb_compose_table_entry_keysym;
    xkb_compose_table_entry_utf8;
```

tools/compose.c
```diff
@@ -108,18 +144,40 @@ main(int argc, char *argv[])
...
+        }
     }
 
-    printf("Compiled successfully from locale %s\n", locale);
+
+    struct xkb_compose_table_iterator *iter = xkb_compose_table_iterator_new(compose_table);
+    while (xkb_compose_table_iterator_next(iter)) {
+        compose_table_print_cb(xkb_compose_table_iterator_current(iter), NULL);
+    }
+    xkb_compose_table_iterator_unref(iter);
```
bluetech, line 176: We should this iteration pattern as an example in the docs.

src/compose/table.c
```diff
@@ -226,3 +227,148 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
+    if (!sequence) {
+        free(iter);
+        return NULL;
+    }
+    iter->entry.sequence = sequence;
+    iter->entry.sequence_length = 0;
+
+    darray_init(iter->cursors);
+    cursor.direction = NODE_LEFT;
+    cursor.node_offset = 1;
```
bluetech, line 291:
```suggestion
    /* Offset 0 is a dummy null entry, skip it. */
    cursor.node_offset = 1;
```

src/compose/table.c
```diff
@@ -226,3 +227,148 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
+enum node_direction {
+    NODE_LEFT = 0,
+    NODE_DOWN,
+    NODE_RIGHT,
+    NODE_UP
+};
+
+struct xkb_compose_table_iterator_cursor {
+    uint32_t node_offset:30; /* WARNING: ensure it fits MAX_COMPOSE_NODES */
+    uint8_t direction:2;     /* enum node_direction */
```
bluetech, line 260: A comment explaining what `direction` means would be helpful. (could be the direction we came from, the direction we're going, etc.)

src/compose/table.c
```diff
@@ -226,3 +227,148 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
+        if (node->hikid) {
+            /* Right node */
+            new_cursor.node_offset = node->hikid;
+            darray_append(iter->cursors, new_cursor);
+            cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
+            continue;
+        } else {
+            /* Look for next node in parents */
+            while (iter->cursors.size > 1) {
+                iter->cursors.size--;
```
bluetech, line 363: Let's add a `darray_pop` operation (or something like it) for this, should try to not break the darray encapsulation.

src/compose/table.c
```diff
@@ -226,3 +227,148 @@ xkb_compose_table_new_from_locale(struct xkb_context *ctx,
...
+        iter->entry.sequence_length--;
+        if (node->hikid) {
+            /* Right node */
+            new_cursor.node_offset = node->hikid;
+            darray_append(iter->cursors, new_cursor);
+            cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
+            continue;
+        } else {
+            /* Look for next node in parents */
+            while (iter->cursors.size > 1) {
```
bluetech, line 362:
```suggestion
            while (darray_size(iter->cursors) > 1) {
```

bluetech (commented):
Thanks for picking this up @wismill, looks good!

I still want to read over `xkb_compose_table_iterator_next` more carefully, but left a few comments in the meantime.

bluetech:
I read over the `iterator_next` code. Nice work, it's not easy to convert recursion to an iterator like that.

I think code can be streamlined a bit to make it more directly analogous to the recursive version. Below is my suggestion. I didn't test it really beyond running your new tests and verifiying that the output stays the same on `en_US.UTF8/Compose`. Let me know what you think.

```diff
diff --git a/src/compose/table.c b/src/compose/table.c
index 1ff26b3..403309e 100644
--- a/src/compose/table.c
+++ b/src/compose/table.c
@@ -312,28 +312,47 @@ xkb_compose_table_iterator_current(struct xkb_compose_table_iterator *iter)
 XKB_EXPORT struct xkb_compose_table_entry *
 xkb_compose_table_iterator_next(struct xkb_compose_table_iterator *iter)
 {
+    /*
+     * This function takes the following recursive traversal function,
+     * and makes it non-recursive and resumable. The iter->cursors stack
+     * is analogous to the call stack, and cursor->direction to the
+     * instruction pointer of a stack frame.
+     *
+     *    traverse(xkb_keysym_t *sequence, size_t sequence_length, uint16_t p) {
+     *        if (!p) return
+     *        // cursor->direction == NODE_LEFT
+     *        node = &darray_item(table->nodes, p)
+     *        traverse(sequence, sequence_length, node->lokid)
+     *        // cursor->direction == NODE_DOWN
+     *        sequence[sequence_length++] = node->keysym
+     *        if (node->is_leaf)
+     *            emit(sequence, sequence_length, node->leaf.keysym, table->utf[node->leaf.utf8])
+     *        else
+     *            traverse(sequence, sequence_length, node->internal.eqkid)
+     *        sequence_length--
+     *        // cursor->direction == NODE_RIGHT
+     *        traverse(sequence, sequence_length, node->hikid)
+     *        // cursor->direction == NODE_UP
+     *    }
+     */
+
     struct xkb_compose_table_iterator_cursor *cursor;
-    struct xkb_compose_table_iterator_cursor new_cursor;
     const struct compose_node *node;
 
-    cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
-    new_cursor.direction = NODE_LEFT;
-
-    while (cursor->direction < NODE_UP) {
+    while (!darray_empty(iter->cursors)) {
+        cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
         node = &darray_item(iter->table->nodes, cursor->node_offset);
 
-        if (cursor->direction == NODE_LEFT) {
-            /* Left node */
+        switch (cursor->direction) {
+        case NODE_LEFT:
             cursor->direction = NODE_DOWN;
             if (node->lokid) {
-                new_cursor.node_offset = node->lokid;
+                struct xkb_compose_table_iterator_cursor new_cursor = {node->lokid, NODE_LEFT};
                 darray_append(iter->cursors, new_cursor);
-                cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
-                continue;
             }
-        }
-        if (cursor->direction == NODE_DOWN) {
-            /* Down node */
+            break;
+
+        case NODE_DOWN:
             cursor->direction = NODE_RIGHT;
             assert (iter->entry.sequence_length <= MAX_LHS_LEN);
             iter->entry.sequence[iter->entry.sequence_length] = node->keysym;
@@ -343,29 +362,23 @@ xkb_compose_table_iterator_next(struct xkb_compose_table_iterator *iter)
                 iter->entry.utf8 = &darray_item(iter->table->utf8, node->leaf.utf8);
                 return &iter->entry;
             } else {
-                new_cursor.node_offset = node->internal.eqkid;
+                struct xkb_compose_table_iterator_cursor new_cursor = {node->internal.eqkid, NODE_LEFT};
                 darray_append(iter->cursors, new_cursor);
-                cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
-                continue;
             }
-        }
-        cursor->direction = NODE_UP;
-        iter->entry.sequence_length--;
-        if (node->hikid) {
-            /* Right node */
-            new_cursor.node_offset = node->hikid;
-            darray_append(iter->cursors, new_cursor);
-            cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
-            continue;
-        } else {
-            /* Look for next node in parents */
-            while (iter->cursors.size > 1) {
-                iter->cursors.size--;
-                cursor = &darray_item(iter->cursors, iter->cursors.size - 1);
-                if (cursor->direction < NODE_UP) {
-                    break;
-                }
+            break;
+
+        case NODE_RIGHT:
+            cursor->direction = NODE_UP;
+            iter->entry.sequence_length--;
+            if (node->hikid) {
+                struct xkb_compose_table_iterator_cursor new_cursor = {node->hikid, NODE_LEFT};
+                darray_append(iter->cursors, new_cursor);
             }
+            break;
+
+        case NODE_UP:
+            iter->cursors.size--;
+            break;
         }
     }
 
```

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+ * returns the empty string.
+ *
+ * @memberof xkb_compose_table_entry
+ * @since 1.6.0
+ */
+const char *
+xkb_compose_table_entry_utf8(struct xkb_compose_table_entry *entry);
+
+/**
+ * Iterator for a compose table.
```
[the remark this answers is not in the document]
wismill: Done

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+ *
+ * @memberof xkb_compose_table
```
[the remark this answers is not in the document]
wismill: Done

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+ * @sa xkb_compose_table_iterator_new()
+ * @since 1.6.0
+ */
+struct xkb_compose_table_iterator;
+
+/**
+ * Create a new iterator for a compose table.
+ *
+ * @memberof xkb_compose_table
```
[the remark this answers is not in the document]
wismill: Done

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,133 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+xkb_compose_table_iterator_new(struct xkb_compose_table *table);
+
+/**
+ * Free a compose iterator.
+ *
+ * @memberof xkb_compose_table
+ * @since 1.6.0
+ */
+void
+xkb_compose_table_iterator_unref(struct xkb_compose_table_iterator *iter);
```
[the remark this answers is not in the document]
wismill: Good catch!

include/xkbcommon/xkbcommon-compose.h
wismill: What would be the use case?

tools/compose.c
wismill: Done

src/compose/table.c
wismill: Explicit is better. Done

src/compose/table.c
wismill: Done

src/compose/table.c
wismill: I added `darray_remove_last`, as pop returns an element that we would not use.

src/compose/table.c
wismill: Good idea

wismill:
Reabse. @bluetech your version is slightly slower in the new benchmark. But I think we can still do better (faster), as the current implementation with “direction” is maybe a bit naive. I have something in the works, I will upload it soon.

include/xkbcommon/xkbcommon-compose.h
wismill: fixed

tools/compose.c
wismill: fixed

xkbcommon.map
wismill: fixed

wismill:
@bluetech Let’s continue the experience: I added a new implemention and restore the other ones, guarded by `#if` macros (look for `ITER_IMPL` and `USE_FOREACH` to choose).

| Implementation | Execution time (s) for 10,000 iterations |
| - | - |
| Foreach (reference) | 1.200935 |
| Iterator 1 (with bit field) | 2.425992 |
| Iterator 1 (without bit field) | 2.319385 |
| Iterator 2 | 2.829882 |
| Iterator 3 (with bit field) | 2.042240 |
| Iterator 3 (without bit field) | 1.683946 |

As you can see, the latest implementation is the closest to the reference `foreach` one. Because we need to keep track of the current position, we cannot beat the foreach implementation. I wonder how we could get closer though.

include/xkbcommon/xkbcommon-compose.h
bluetech: Something like support for preserving the original entries order. But as I said we can omit the flags now as it's probably not going to be needed.

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,186 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+ * @since 1.6.0
+ */
+struct xkb_compose_table_iterator;
+
+/**
+ * Create a new iterator for a compose table.
+ *
+ * Intended use:
+ *
+ * ```c
```
bluetech, line 428: Is this equivalent to the following?

```c

struct xkb_compose_table_iterator *iter = xkb_compose_table_iterator_new(compose_table);
struct xkb_compose_table_entry *entry;
while (entry = xkb_compose_table_iterator_next(iter)) {
    //...
}
xkb_compose_table_iterator_free(iter);
```

tools/compose.c
```diff
@@ -29,34 +29,67 @@
...
+
 int
 main(int argc, char *argv[])
 {
     int ret = EXIT_FAILURE;
     struct xkb_context *ctx = NULL;
     struct xkb_compose_table *compose_table = NULL;
     const char *locale = NULL;
+    const char *path = NULL;
+    /* TODO: add option */
```
bluetech, line 83: Not needed while we only have one

bluetech (commented):
I liked the Iterator 2 solution because it corresponds directly to the obviously-correct recursive code, so was easiest to understand and assess correctness.

I haven't read the Iterator 3 solution in detail yet. Seems like its the fastest, though all solutions are plenty fast for the task at hand, so my suggestion is to prefer to maintainability/correctness aspects over performance.

If you think Iterator 3 is cleaner than Iterator 2 then let me know and I'll read it over. If not, my suggestion is to go with Iterator 2. But I'll leave the decision to you :)

include/xkbcommon/xkbcommon-compose.h
wismill: Yes, I understand the preserving order part, but where would it be _useful_? One use case I see is for cleaning/normalizing compose files, but as we do not keep comments nor white spaces, such a hypothetical tool will still have to parse and compare resulting entries with xkbcommon.

include/xkbcommon/xkbcommon-compose.h
wismill: I think so, because `xkb_compose_table_iterator_next` returns the next entry of NULL is there is none. Now I wonder if `xkb_compose_table_iterator_current` is needed. I added because the iterator API model I consulted had one, but keeping our API minimal is probably a good idea!

wismill:
> If you think Iterator 3 is cleaner than Iterator 2 then let me know and I'll read it over. If not, my suggestion is to go with Iterator 2. But I'll leave the decision to you :)

@bluetech I understand your argument. My POV is that once I have a correct implementation, I do like to improve it to make it efficient. Implementation 3 runs in about 40% less time that implementation 2, which is not negligible.

I have the feeling that implementation 3 has good perf but is still too verbose. I would appreciate a review, but I know we all have limited time for this project.

So I propose the following: we can go with implementation 2 for now, and iterate on better implementation in a follow-up PR. This way we ensure it will make for 1.6.0 release without delaying it further.

tools/compose.c
wismill: Fixed locally. I will upload this in the next substantial update, in order to save a bit of CI.

include/xkbcommon/xkbcommon-compose.h
bluetech: Right, I think we should omit `current` if we can do without it

bluetech:
> So I propose the following: we can go with implementation 2 for now, and iterate on better implementation in a follow-up PR. This way we ensure it will make for 1.6.0 release without delaying it further.

:+1:

include/xkbcommon/xkbcommon-compose.h
wismill: Done

wismill:
@bluetech I removed alternative implementations and squashed all the commits. I left your original commit atm. Are you ok to squash it as well? I used the `Co-authored-by` in the last commit with your name.

include/xkbcommon/xkbcommon-compose.h
```diff
@@ -299,6 +299,141 @@ xkb_compose_table_ref(struct xkb_compose_table *table);
...
+xkb_compose_table_iterator_free(struct xkb_compose_table_iterator *iter);
+
+/**
+ * Get the next compose entry from a compose table iterator.
+ *
+ * The entries are returned in lexicographic order of the left-hand
+ * side of entries. This does not correspond to the order in which
+ * the entries appear in the Compose file.
+ *
+ * Returns `NULL` in case there is no more entries.
```
bluetech, line 429: Let's mention the lifetime of the return value, specifically, that's it's valid until the next call to `_next()`.

bluetech (approved):
Sure, feel free to squash.

include/xkbcommon/xkbcommon-compose.h
wismill: Done.

wismill:
Squashed and rebased.

Remaining items of the original commit’s TODO:

- Convert tools/compose.c to an `xkbcli compose`:
    - Better names for the `--locale` arguments.
    - Add `--format` (currently only `text-v1`).
    - Check return value of `printf`s.
    - Add to `xkbcli`.
    - Test it in `test/tool-option-parsing.py`.
    - Write manpage.

I think this belongs to another PR, as this tool was not introduced by the current PR.

wismill:
Reworked the commit message. Will merge when test passes.

PR #370, opened 2023-08-05
parser.y update for deprecated %pure-parser

mckellygit:
Hi,
Thank you for libxkbcommon.
I get this when building -
../src/xkbcomp/parser.y:88.1-12: warning: deprecated directive: ‘%pure-parser’, use ‘%define api.pure’ [-Wdeprecated]
and changed parser.y to use newer method.
Not sure if changing causes more harm than good or if we can version this somehow.
I am using bison 3.5.1
thx

### DIFF
--- a/src/xkbcomp/parser.y
+++ b/src/xkbcomp/parser.y
@@ -85,7 +85,7 @@ resolve_keysym(const char *name, xkb_keysym_t *sym_rtrn)
 #define param_scanner param->scanner
 %}
 
-%pure-parser
+%define api.pure
 %lex-param      { struct scanner *param_scanner }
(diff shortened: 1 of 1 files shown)


bluetech:
Yes this is annoying, the reason that I haven't changed it so far is that we also support [byacc](https://www.invisible-island.net/byacc/byacc.html), which didn't support `@define api.pure`. BUT, I just checked again and

```
2023-05-16  Thomas E. Dickey  <dickey@invisible-island.net>
	* reader.c:
	recognize bison's "%define api.pure" as an alternative to "%pure-parser"
```

and a quick test shows that it works. So we can merge this, thanks!

PR #374, opened 2023-09-19
Disallow producing NULL with escape sequences & prevent overflow

### LINKED ISSUE #363: Octal escape allows \0..\777
Working on #332, I realize that we allow to insert NULL characters in string literals.

This is probably not a good idea. In fact, what use case requires to use control characters in our string literals?

Also, the octal parser accept the range `\0..\777`. The result is cast to `char` which will silently overflow. We should at least check that the upper bound is `\377`, i.e. 0xff.

As there is no specified encoding, in principle we do not have to limit to ASCII and the input file could use 0x80..0xff to encode _explicitly_ e.g. in utf-8. 🤔 But that sounds silly!

wismill:
NULL usually terminates the strings; allowing to produce it via escape sequences may lead to undefined behaviour.

- Make null escape sequences (e.g. `\0` and `\x0`) invalid
- Introduce the new message: XKB_WARNING_INVALID_ESCAPE_SEQUENCE.
- Prevent overflow (e.g. `\400`, `\777`)
- Improve log messages.
- Add corresponding tests.

Fixes #363


src/utils.h
```diff
@@ -65,6 +65,12 @@
 #define STRINGIFY(x) #x
 #define STRINGIFY2(x) STRINGIFY(x)
 
+/* Check if a char is valid in a string literal */
+static inline bool
+is_valid_char(char c) {
```
bluetech, line 70: The style (though I don't really like it myself...) is to put the `{` of functions in its own line

test/data/keymaps/invalid-escape-sequence.xkb
```diff
@@ -0,0 +1,9 @@
+xkb_keymap {
+    // The following include statement has an octal escape sequence that
+    // must be ignored. Else it would insert a NULL character and thus
```
bluetech, line 3: NULL -> NUL

bluetech (approved):
Great catches, thanks.

src/utils.h
wismill: Sorry

test/data/keymaps/invalid-escape-sequence.xkb
wismill: Actually, the character name is NULL. `NUL` is the [ASCII abreviation](https://en.wikipedia.org/wiki/C0_and_C1_control_codes#ASCII). I prefer the name, but if you prefer the abbreviation then I suggest we use `\NUL` instead.

wismill:
Rebased with review fixes.

@bluetech @whot this PR is only about `\NUL` character, but I wonder if it make sense to allow the other control characters.

bluetech:
> this PR is only about \NUL character, but I wonder if it make sense to allow the other control characters.

Is there a reason to prevent it?

wismill:
Well, I wonder if some strings are exposed and where. They might be exposed by some of the log entries, currently unescaped. They could be exposed by functions like `xkb_keymap_layout_get_name` and those in `xkbregistry`, but I would have to investigate this carefully. My point is after discovering the `\NUL` improper handling, there may be insufficient filtering and/or escaping that the users of the API are unaware of. Also, because there is no explicit encoding, I am not sure how the `char*` are interpreted. In `xkeyboard-config` they are all ASCII if I am not mistaken, but there could be a way to craft malicious XKB files. It would not be an issue for xkbcommon itself, but could create problems for its users.

Just thinking out loud here; we should deal with it in a dedicated issue if relevant.

bluetech:
> there could be a way to craft malicious XKB files

I have thought about this topic before.

My assumption is that xkbcommon, being a huge custom parser written in C in the 90s, is not safe from malicious input. So the question is whether there is a threat model in which xkbcommon is fed untrusted keymaps. In the cases I know and could think of, xkb keymap input comes trusted sources, like a compositor, X server, `/usr/share`, user's $HOME, xkbcli flag. If any of those are compromised then something has already gone wrong. (Maybe a malicious person posting a "Try my awesome keymap" post on the internet, causing users to download and "execute" it)?

Of course this is no excuse to make xkbcommon full of holes, and we even have a rudimentary fuzzer, but it does make me weigh compatibility more than malicious input.

wismill:
Rebased & improved commit message. Will merge when CI is successful.

PR #379, opened 2023-09-21
Messages (part 2)

wismill:
Follow-up of #332 

- Merge log functions with and without message code.
- Add more messages entries.


src/messages-codes.h.jinja
```diff
@@ -6,6 +6,14 @@
 
 #include <stdint.h>
 
+/**
+ * Special case when no message identifier is defined.
+ *
+ * @added 1.6.0
```
bluetech, line 12: This is not public API so the `@added` is not needed

bluetech (approved):
Excellent work. Everything looks good to me.

src/messages-codes.h.jinja
wismill: Right! I also removed it from `enum xkb_message_code`.

wismill:
Applied review fixes and rework the commit messages.

PR #384, opened 2023-09-29
Add bash completions

wismill:
WIP. This is the first time I write bash completion, be kind!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -417,6 +417,7 @@ man_pages = []
 # Tools
 build_tools = get_option('enable-tools') and cc.has_header_symbol('getopt.h', 'getopt_long', prefix: '#define _GNU_SOURCE')
 if build_tools
+    # Common resources
     libxkbcommon_tools_internal_sources = [
         'tools/tools-common.h',
         'tools/tools-common.c',
@@ -432,10 +433,27 @@ if build_tools
         dependencies: dep_libxkbcommon,
     )
 
+    # Tool: xkbcli
     executable('xkbcli', 'tools/xkbcli.c',
                dependencies: tools_dep, install: true)
     install_man('tools/xkbcli.1')
 
+    if get_option('enable-bash-completion')
+        bash_completion_path = get_option('bash-completion-path')
+        if bash_completion_path == ''
+            bash_completion = dependency('bash-completion', required: false)
+            if bash_completion.found()
+                bash_completion_path = bash_completion.get_variable(pkgconfig: 'completionsdir')
+            else
+                bash_completion_path = get_option('datadir') / 'bash-completion/completions'
+            endif
+        endif
+        install_data('tools/xkbcli-bash-completion.sh',
+                     rename: 'xkbcli',
+                     install_dir: bash_completion_path)
+    endif
+
+    # Tool: compile-keymap
     xkbcli_compile_keymap = executable('xkbcli-compile-keymap',
                                        'tools/compile-keymap.c',
                                        dependencies: tools_dep,
@@ -450,12 +468,16 @@ if build_tools
                c_args: ['-DENABLE_PRIVATE_APIS'],
                include_directories: [include_directories('src', 'include')],
                install: false)
+
+    # Tool: compose
     executable('compose',
                'tools/compose.c',
                dependencies: tools_dep,
                include_directories: [include_directories('src', 'include')],
                install: false)
     configh_data.set10('HAVE_XKBCLI_COMPILE_KEYMAP', true)
+
+    # Tool: how-to-type
     executable('xkbcli-how-to-type',
                'tools/how-to-type.c',
                dependencies: tools_dep,
@@ -463,6 +485,8 @@ if build_tools
                install_dir: dir_libexec)
     install_man('tools/xkbcli-how-to-type.1')
     configh_data.set10('HAVE_XKBCLI_HOW_TO_TYPE', true)
+
+    # Tool: interactive-evdev
     if cc.has_header('linux/input.h')
         executable('xkbcli-interactive-evdev',
                    'tools/interactive-evdev.c',
@@ -481,6 +505,8 @@ if build_tools
                 include_directories: [include_directories('src', 'include')],
                 install: false)
     endif
+
+    # Tool: interactive-x11
     if get_option('enable-x11')
         x11_tools_dep = declare_dependency(
             link_with: libxkbcommon_x11,
@@ -498,6 +524,8 @@ if build_tools
         install_man('tools/xkbcli-interactive-x11.1')
         configh_data.set10('HAVE_XKBCLI_INTERACTIVE_X11', true)
     endif
+
+    # Tool: interactive-wayland
     if get_option('enable-wayland')
         wayland_client_dep = dependency('wayland-client', version: '>=1.2.0', required: false)
         wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.12', required: false)
@@ -534,6 +562,7 @@ You can disable the Wayland xkbcli programs with -Denable-wayland=false.''')
         configh_data.set10('HAVE_XKBCLI_INTERACTIVE_WAYLAND', true)
     endif
 
+    # Tool: list
     if get_option('enable-xkbregistry')
         configh_data.set10('HAVE_XKBCLI_LIST', true)
         executable('xkbcli-list',
@@ -544,6 +573,7 @@ You can disable the Wayland xkbcli programs with -Denable-wayland=false.''')
         install_man('tools/xkbcli-list.1')
     endif
 
+    # Tool: check-messages
     executable('xkb-check-messages',
                'tools/check-messages.c',
                'tools/messages.c',
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -14,6 +14,11 @@ option(
     description: 'The X locale root [default=$datadir/X11/locale]',
 )
 option(
+    'bash-completion-path',
+    type: 'string',
+    description: 'Directory for bash completion scripts'
+)
+option(
     'default-rules',
     type: 'string',
     value: 'evdev',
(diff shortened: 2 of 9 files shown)


scripts/completion.sh
```diff
@@ -0,0 +1,72 @@
+# bash completion support for xkbcli.
+
+___xkbcli_main()
+{
+    _init_completion -s || return
```
bluetech, line 5: Can you add a comment (maybe at the top of the file) that this comes from the [bash-completion](https://github.com/scop/bash-completion) package? It took me a bit of effort to find it, turns out it's deprecated in the yet-unreleased version 2.12 (though without emitting a deprecation messages).

The function [doc](https://github.com/scop/bash-completion/blob/fdf4456186eb4548ef628e65fb1be73d8e4695e9/bash_completion.d/000_bash_completion_compat.bash#L205) says "Before calling this function, make sure cur, prev, words, and cword are local, ditto split if you use -s.", but it's not done here. Is this a problem?

scripts/completion.sh
```diff
@@ -0,0 +1,72 @@
...
+            -*) ;;
+            *)
+            cmd="$s"
+            break
+            ;;
+        esac
+        (( i++ ))
+    done
+
+    local subcommands=(
```
bluetech, line 23: It would be good to avoid duplicating this list here, since it at risk of getting out of sync. Possible alternatives (in my order of preference) are to parse the `--help` output, add a hidden `xkbcli` subcommand/flag which lists the subcommands and use that here, or fill using a template.

bluetech (commented):
Nice, I've missed this a few times. I only ever wrote [one](https://github.com/systemd/systemd/blob/main/shell-completion/bash/systemctl.in) completion script but I forgot almost everything about it.

wismill:
Free desktop Gitlab is down atm.

bluetech:
Great work as always. For the meson hookup you can check out [systemd's implementation](https://github.com/systemd/systemd/blob/main/shell-completion/bash/meson.build) for reference.

wismill:
Added meson install entry and reworked the commit messages.

bluetech (approved):

meson.build
```diff
@@ -432,10 +433,29 @@ if build_tools
...
     )
 
+    # Tool: xkbcli
     executable('xkbcli', 'tools/xkbcli.c',
                dependencies: tools_dep, install: true)
     install_man('tools/xkbcli.1')
 
+    bash_completion_path = get_option('bash-completion-path')
+    if bash_completion_path == ''
+        bash_completion = dependency('bash-completion', required : false)
```
whot, line 443:
```suggestion
        bash_completion = dependency('bash-completion', required: false)
```
mostly basing this on the context of this diff, maybe we have spaces before `:` elsewhere? if so we should probably normalize this

meson.build
```diff
@@ -432,10 +433,29 @@ if build_tools
...
+        if bash_completion.found()
+            bash_completion_path = bash_completion.get_variable(pkgconfig : 'completionsdir')
```
whot, line 445:
```suggestion
            bash_completion_path = bash_completion.get_variable(pkgconfig: 'completionsdir')
```

meson.build
```diff
@@ -432,10 +433,29 @@ if build_tools
...
+            bash_completion_path = get_option('datadir') / 'bash-completion/completions'
+        endif
+    endif
+    if bash_completion_path != 'no'
+        install_data(
+            'tools/xkbcli-completion.sh',
+            rename: 'xkbcli',
+            install_dir: bash_completion_path
+        )
+    endif
```
whot, line 456: handwavy: I'm wondering if it's better to have a `bash-completion` as meson feature and the path as a separate meson option that is either `''` or an explicit path. Mostly because while I can see the point of disabling bash completion, setting the path is probably quite a specific niche case. And setting a path string as `no` feels like an old autotools leftover, meson options cost very little in comparison.

This is a suggestion only, feel free to ignore.

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+#       thus any new subcommand or option will be supported, as long as it has its
+#       entry in the help messages. This should result in low maintenancei effort.
+
+___xkbcli_main()
+{
+    # Initialization: https://github.com/scop/bash-completion/blob/fdf4456186eb4548ef628e65fb1be73d8e4695e9/bash_completion.d/000_bash_completion_compat.bash#L205
+    local cur prev words cword
+    _init_completion -s || return
+
+    local i=1 cmd
```
whot, line 14:
```suggestion
    local cur prev words cword cmd
    _init_completion -s || return

    local i=1
```
Mixing uninitialized and initialized declarations seems messy

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+    do
+        local s="${COMP_WORDS[i]}"
+        case "$s" in
+            -*) ;;
+            *)
+            cmd="$s"
+            break
+            ;;
+        esac
+        (( i++ ))
```
whot, line 27: huh, I learned something here, I always thought it had to be `i=$((i + 1))` but this is much nicer.

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+    done
+
+    # Parse available subcommands
+    local help_output line is_command_list=false subcommands=()
+    help_output=$(xkbcli --help)
+    while IFS= read -r line; do
```
whot, line 33:
```suggestion
    while IFS='' read -r line; do
```
for clarity? or is this not the same? my bash scripting is quite rusty

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+                subcommands+=("${BASH_REMATCH[1]}")
+            # Detect end of subcommand list based on indentation
+            elif [[ "$line" =~ ^[[:graph:]] ]]; then
+                is_command_list=false
+            fi
+        # Detect start of subcommand list
+        elif [[ "$line" == "Commands:" ]]; then
+            is_command_list=true
+        fi
+    done <<< "$help_output"
```
whot, line 47:
```suggestion
    done < <(xkbcli --help)
```
If you want to get fancy and skip the `help_output` variable.

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+        local opts
+        # Doc for _parse_help: https://github.com/scop/bash-completion/blob/fdf4456186eb4548ef628e65fb1be73d8e4695e9/bash_completion.d/000_bash_completion_compat.bash#L311
+        opts=$(_parse_help xkbcli)
+        local cur="${COMP_WORDS[COMP_CWORD]}"
+        COMPREPLY=($(compgen -W "${subcommands[*]} $opts" -- "$cur"))
+        return
+    fi
+
+    # Found a supported subcommand: proceed to completion
+    if [[ " ${subcommands[*]} " =~ " $cmd " ]]
```
whot, line 61: are the spaces here on purpose?

also, `if [[ ]]; then` for consistency please

tools/xkbcli-completion.sh
```diff
@@ -0,0 +1,97 @@
...
+            fi
+        # Detect start of subcommand list
+        elif [[ "$line" == "Commands:" ]]; then
+            is_command_list=true
+        fi
+    done <<< "$help_output"
+
+    # No previous subcommand or incomplete: completion for root xkbcli command
+    if [[ "$i" -eq "$COMP_CWORD" ]]
+    then
```
whot, line 51:
```suggestion
    if [[ "$i" -eq "$COMP_CWORD" ]]; then
```
For consistency with the above

whot (approved):
This LGTM, thanks. A few style nitpicks but that's about it, feel free to merge when those are addressed (or not :)

meson.build
wismill: Fixed

meson.build
wismill: Fixed

meson.build
wismill: I agree. I adapted the meson script from `systemd` as Ran proposed and we can continue improving on it.

tools/xkbcli-completion.sh
wismill: Fixed

tools/xkbcli-completion.sh
wismill: > are the spaces here on purpose?

I admit I adapted a copy/paste and did not pay much attention because it worked 😅. So the spaces are needed because we need to match _whole_ subcommands. `${subcommands[*]}` provide a space-separated list, but we need space before and after.

I switched to an explicit pattern. There may be better solution than pattern matching though.

meson_options.txt
```diff
@@ -13,6 +13,11 @@ option(
     type: 'string',
     description: 'The X locale root [default=$datadir/X11/locale]',
 )
+option(
+    'bash-completion-path',
+    type: 'string',
+    description: 'directory for bash completion scripts'
```
whot, line 19:
```suggestion
    description: 'Directory for bash completion scripts'
```

tools/xkbcli-bash-completion.sh
```diff
@@ -0,0 +1,96 @@
...
+___xkbcli_main()
+{
+    # Initialization: https://github.com/scop/bash-completion/blob/fdf4456186eb4548ef628e65fb1be73d8e4695e9/bash_completion.d/000_bash_completion_compat.bash#L205
+    local cur prev words cword cmd
+    _init_completion -s || return
+
+    # Find subcommand
+    local i=1
+    while [[ "$i" -lt "$COMP_CWORD" ]]
+    do
```
whot, line 17:
```suggestion
    while [[ "$i" -lt "$COMP_CWORD" ]]; do
```

whot (approved):
LGTM, thanks!

tools/xkbcli-bash-completion.sh
wismill: Fixed

meson_options.txt
wismill: Fixed

PR #389, opened 2023-10-25
Prevent recursive includes of keymap components

wismill:
Prevent recursive includes of keymap components:

- Add file name property to `XkbFile`. This allow better identification of the file being processed.
- Add check for recursive includes of keycodes, key types, compat and symbols keymap components.
- Add tests for recursive includes.

I would like to get the design right wrt further work on #390.

### DIFF
--- a/src/xkbcomp/ast-build.c
+++ b/src/xkbcomp/ast-build.c
@@ -534,8 +534,8 @@ err:
 }
 
 XkbFile *
-XkbFileCreate(enum xkb_file_type type, char *name, ParseCommon *defs,
-              enum xkb_map_flags flags)
+XkbFileCreate(enum xkb_file_type type, const char *file_name, char *name,
+              ParseCommon *defs, enum xkb_map_flags flags)
 {
     XkbFile *file;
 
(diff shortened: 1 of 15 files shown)


wismill:
There seems to be memory leaks in _key types_ code that make the test suite failed when runing with valgrind.

wismill:
Fixed memory leak. I found some other leaks and grouped them in #391.

This PR will need to be rebase once #391 is merged.

bluetech:
A simpler and arguably more robust method for detecting include loops is to have an include depth limit. Have you considered it?

wismill:
@bluetech I did, but then I thought it would be cleaner to do an exact check that:
- avoids cycling completely: check occurs _before_ loading the recursive file.
- provides useful log message.

I guess this issue is unlikely enough that we could just do as you propose. I have no idea what is the depth threshold in `xkeyboard-config`. I guess __15__ is a reasonable threshold, with plenty of margin for keymaps in the wild.

wismill:
Created an alternative design in #392.

PR #391, opened 2023-10-25
Fix some memory leaks

wismill:
I stumble on these memory leaks while working on #389.


src/xkbcomp/types.c
```diff
@@ -180,6 +180,11 @@ MergeIncludedKeyTypes(KeyTypesInfo *into, KeyTypesInfo *from,
 {
     if (from->errorCount > 0) {
         into->errorCount += from->errorCount;
+        // Free unused types
```
bluetech, line 183: If I'm reading correctly, the "contract" of `MergeIncludedKeyTypes` is that it steals/empties out `from` from within, but leaves it in tact; the caller still needs to clear `from`.  And `HandleIncludeKeyTypes` does call `ClearKeyTypesInfo` after each `MergeIncludedKeyTypes` call.

So I suppose the question is, shouldn't `ClearKeyTypesInfo` take care of freeing `types`?

bluetech (commented):
Good catches! I have a question on the key types compilation commit, the other two look good to me.

src/xkbcomp/types.c
wismill: That was my first thought, but then I ran into double-free issues.

tools/compile-keymap.c
```diff
@@ -244,8 +244,10 @@ print_keymap(struct xkb_context *ctx, const struct xkb_rule_names *rmlvo)
     if (keymap == NULL)
         return false;
 
-    printf("%s\n", xkb_keymap_get_as_string(keymap,
-                                            XKB_KEYMAP_FORMAT_TEXT_V1));
+    char *buf = xkb_keymap_get_as_string(keymap, XKB_KEYMAP_FORMAT_TEXT_V1);
```
whot, line 247: @bluetech are we happy to go C99? Because just these late declarations are going to make the code better in a lot of places but let's make this official.

src/xkbcomp/include.c
```diff
@@ -267,11 +267,13 @@ FindFileInXkbPath(struct xkb_context *ctx, const char *name,
...
             if (pathRtrn) {
                 *pathRtrn = buf;
-                buf = NULL;
+            } else {
+                free(buf);
             }
             *offset = i;
             goto out;
         }
+        free(buf);
```
whot, line 276: possibly other option:
```c
char buf[PATH_MAX];

sprintf(buf, ...)

if (everything is fine)
    pathRtrn = strdup(buf);
```

which means we don't need to track buf itself.

whot (approved):

src/xkbcomp/types.c
bluetech: Yep but I think that's what we should fix. I think we need this:

```diff
diff --git a/src/xkbcomp/types.c b/src/xkbcomp/types.c
index 34e1028..9f499ad 100644
--- a/src/xkbcomp/types.c
+++ b/src/xkbcomp/types.c
@@ -201,6 +201,7 @@ MergeIncludedKeyTypes(KeyTypesInfo *into, KeyTypesInfo *from,
             if (!AddKeyType(into, type, false))
                 into->errorCount++;
         }
+        darray_init(from->types);
     }
 }
```

tools/compile-keymap.c
bluetech: No problem with using C99 late declarations from me.

src/xkbcomp/include.c
wismill: Done. I introduced the new message `XKB_ERROR_INSUFFICIENT_BUFFER_SIZE`.

src/xkbcomp/types.c
wismill: So I re-analyze the issue and I think I get it right this time. I left comments in the code to explain it, as it was difficult to detect for me.

wismill: And I fixed another leak in `HandleKeyTypeDef`.

src/xkbcomp/types.c
```diff
@@ -192,6 +195,7 @@ MergeIncludedKeyTypes(KeyTypesInfo *into, KeyTypesInfo *from,
 
     if (darray_empty(into->types)) {
         into->types = from->types;
+        /* Types stealen via shallow copy, so reinitialze the array */
```
bluetech, line 198:
```suggestion
        /* Types stolen via shallow copy, so reinitialize the array */
```

bluetech (approved):

src/xkbcomp/types.c
wismill: 😳 sorry

PR #392, opened 2023-10-26
Prevent recursive includes of keymap components (using include depth)

wismill:
Prevent recursive includes of keymap components by limiting the include depth. This is an alternative design of #389.

This PR has a commit that fixes a memory leak, but should be rebased once #391 is merged.


wismill:
Added corresponding new log message `recursive-include`.

bluetech (approved):
Thanks, looks great to me.

wismill:
Rebased.

PR #394, opened 2023-10-29
Skip heading UTF-8 encoded BOM

### LINKED ISSUE #393: xkbcommon chokes on UTF-8 BOM
I am on Ubuntu 23.10. I am developing with Rust and Cargo. I am not using xkbcommon directly, so I am not sure how to tell the exact version of libxkbcommon I am using, but if you tell me I can look it up.

I compile and run (with `cargo run`) *any* Rust program using winit. For example https://github.com/mcclure/webgpu-tutorial-rs or example "Editor" from https://github.com/iced-rs/iced.

For any such app, at boot it prints the line:

```
xkbcommon: ERROR: /home/mcc/.XCompose:1:1: unrecognized token
```

Additionally, I find XCompose sequences do not work.

The error is perplexing to me because my .XCompose begins with `<Multi_key> <minus> <minus>: "—"`. I try changing the first line to `include "%L"` on a recommendation; this does nothing for this particular error..

Then it hits me. I run:

```
mcc@Anthy:~/work/r/other/iced/examples/editor$ file ~/.XCompose 
/home/mcc/.XCompose: C source, Unicode text, UTF-8 (with BOM) text
```

Oh no. 
I use Sublime Text to remove the BOM from ~/.XCompose.
Immediately, the problem goes away and XCompose sequences work.

The [BOM](https://en.wikipedia.org/wiki/Byte_order_mark), as you may already know, is an element of the Unicode Standard for identifying which Unicode representation a particular file is using (UTF-8, UTF-16 LE, or UTF-16 BE). This BOM is conventional in UTF-16, but relatively rare in UTF-8. However it is *legal* in UTF-8 (Wikipedia, with cites: "The Unicode Standard permits the BOM in UTF-8[4] but does not require or recommend its use.[5]"), and some editors will add BOMs to UTF-8 files as a matter of course.

Because a .XCompose file will often contain UTF-8, a UTF-8 BOM is a reasonable thing to include. In my testing, a UTF-8 BOM is supported and correctly parsed (which is to say: it is ignored) in .XCompose by many other applications, including Firefox and Chrome.

My "Expected Behavior" is that xkbcommon, on encountering a UTF-8 BOM file, should silently ignore the BOM and parse the file as UTF-8.

For extra credit, if xkbcommon encounters a *UTF-16* BOM, it should print a useful error message; currently it prints a bunch of "unrecognized token"s, but a better error message would be "ERROR: File is UTF-16".

wismill:
Fixes #393

### DIFF
--- a/src/compose/parser.c
+++ b/src/compose/parser.c
@@ -534,6 +534,10 @@ initial:
     production.mods = 0;
     production.modmask = 0;
 
+    /* Skip UTF-8 encoded BOM (U+FEFF) */
+    /* See: https://www.unicode.org/faq/utf_bom.html#bom5 */
+    scanner_str(s, "\xef\xbb\xbf", 3);
+
     /* fallthrough */
 
 initial_eol:
--- a/src/xkbcomp/rules.c
+++ b/src/xkbcomp/rules.c
@@ -1099,6 +1099,10 @@ read_rules_file(struct xkb_context *ctx,
 
     scanner_init(&scanner, matcher->ctx, string, size, path, NULL);
 
+    /* Skip UTF-8 encoded BOM (U+FEFF) */
+    /* See: https://www.unicode.org/faq/utf_bom.html#bom5 */
+    scanner_str(&scanner, "\xef\xbb\xbf", 3);
+
     ret = matcher_match(matcher, &scanner, include_depth, string, size, path);
 
     unmap_file(string, size);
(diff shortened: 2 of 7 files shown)


test/data/rules/bom
```diff
@@ -0,0 +1,21 @@
+﻿! model         = keycodes
+  my_model      = my_keycodes
+  *             = default_keycodes
```
whot, line 3: any chance we can put a comment in this file somewhere near the top to say that it starts with a bom? Otherwise it's a bit hard to tell, given the thing isn't supposed to render :)

whot (approved):
LGTM, thanks!

test/data/rules/bom
wismill: Good point! I added a comment and changed the file name to a more explicit one.

wismill:
Review fixes & improved commit messages.

PR #397, opened 2023-10-30
Detect invalid file encoding early

wismill:
Follow-up of #393 and #394. It provides a useful error message for common invalid file encodings.

Specific target use case: UTF-16 (for Windows users).


mcclure:
I haven't tested this, but reading the code these error messages look very helpful.

src/xkbcomp/rules.c
```diff
@@ -1101,13 +1101,37 @@ read_rules_file(struct xkb_context *ctx,
...
+    if (scanner.s[0] == '\0' || scanner.s[1] == '\0') {
+        if (scanner.s[0] != '\0')
+            scanner.token_column++;
+        scanner_err(&scanner, "unexpected NULL character.");
+        goto encoding_error;
+    }
+    if (!is_ascii(scanner.s[0])) {
+        scanner_err(&scanner, "unexpected non-ASCII character.");
+        goto encoding_error;
+    }
```
whot, line 1118: tbh, this code is bit confusing with the interleaved gotos. the `out` label is not (no longer?) needed, we can return from the error condition after map_file.

but this here should be a helper function so the code reads as something like
```c
if (is_supported_encoding()) {
   scanner_err(&scanner, "yuck!");
   return false;
}

ret = matcher_match(...);
```

that should make it possible to skip all the gotos, or most of them. I think the same applies to the other patches with similar approaches.

Note there's also a bug afaict, skipping to `encoding_error` never calls `unmap_file()` in 1122.

whot (commented):
LGTM for the approach but some of the code is a bit convoluted, I'd appreciate it if we could streamline this a bit.

test/buffercomp.c
```diff
@@ -31,6 +31,59 @@
...
+
+    /* Reject UTF-16LE encoded string */
+    const char *utf16_le =
+        "x\0k\0b\0_\0k\0e\0y\0m\0a\0p\0 \0{\0\n\0"
+        " \0 \0x\0k\0b\0_\0k\0e\0y\0c\0o\0d\0e\0s\0 \0{\0 \0i\0n\0c\0l\0u\0d\0e\0 \0\"\0e\0v\0d\0e\0v\0\"\0 \0}\0;\0\n\0"
+        " \0 \0x\0k\0b\0_\0t\0y\0p\0e\0s\0 \0{\0 \0i\0n\0c\0l\0u\0d\0e\0 \0\"\0c\0o\0m\0p\0l\0e\0t\0e\0\"\0 \0}\0;\0\n\0"
+        " \0 \0x\0k\0b\0_\0c\0o\0m\0p\0a\0t\0 \0{\0 \0i\0n\0c\0l\0u\0d\0e\0 \0\"\0c\0o\0m\0p\0l\0e\0t\0e\0\"\0 \0}\0;\0\n\0"
+        " \0 \0x\0k\0b\0_\0s\0y\0m\0b\0o\0l\0s\0 \0{\0 \0i\0n\0c\0l\0u\0d\0e\0 \0\"\0p\0c\0\"\0 \0}\0;\0\n\0"
+        "}\0;\0";
+    keymap = test_compile_buffer(ctx, utf16_le, 156);
```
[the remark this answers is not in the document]
wismill: Good advice! In fact the number was not even correct before!

src/xkbcomp/rules.c
wismill: Good advice! That’s why I marked the PR as draft.

Refactored. I chose `check_supported_char_encoding` over `is_supported_encoding` because of the BOM detection, which in fact _consumes_ the input.

I also improved the comments.

src/xkbcomp/scanner.c
```diff
@@ -193,16 +193,49 @@ _xkbcommon_lex(YYSTYPE *yylval, struct scanner *s)
...
+        return false;
+    }
+    /* The first character must be ASCII: whitespace, section, comment */
+    if (!is_ascii(scanner->s[0])) {
+        scanner_err(scanner, "unexpected non-ASCII character.");
+        return false;
+    }
+
+    return true;
+}
```
whot, line 221: I think the three `check_supported_char_encoding` are the same, right? If so let's move them to `utils.c`? Or make them part of the scanner API as `scanner_check_...`?


After a *very* quick peek at the code, I think we *may* be able to even put this into `scanner_init()` and just fail that for unexpected encodings?

src/xkbcomp/rules.c
whot: thanks, looking much better now, a bit of deduplication [^1] would be nice though if possible (see my comment below)

[^1]: detriplication, to be technical :)

src/xkbcomp/scanner.c
wismill: Good point. They only differ from their comment about the first character. 

`scanner_init` is also used in `matcher_include` in rules, so I do not see including it there, in order to keep separation of concerns.

I moved it in `scanner-utils.h` as `scanner_check_supported_char_encoding` and moved the aforementioned comments before its use.

whot (approved):
thanks!

PR #398, opened 2023-11-01
Compose: allow overlapping sequences

wismill:
Compose: allow overlapping sequences

Currently xkbcommon does not support Compose overlapping sequences,
unlike GTK and ibus. It means a sequence that is a prefix to a longer
one is simply discarded with a warning.

This is unfortunate:
- There is discrepency for users of xkbcommon, e.g. Qt-based apps.
- It is impossible to have a sequence that is the prefix of another one.
  As a consequence, if one imports e.g. the system locale Compose file
  (as many – most ? – custom Compose files do), some sequences become
  impossible.

  Example: I used `<Multi_key> <minus> <period> : "•" U2022` but
  upstream change in libX11 added `<Multi_key> <minus> <period> <e> : "ė̄"`,
  so I had to introduce an alternative sequence for Qt apps, although
  the previous one works in GTK-based apps.

This commit introduces the new following API:
- `xkb_compose_compile_flags`:
  `XKB_COMPOSE_COMPILE_OVERLAPPING_SEQUENCES`: Allow overlapping sequences
- `xkb_compose_status`:
  `XKB_COMPOSE_CANDIDATE`: A complete sequence has been matched, but a
  longer sequence also exists.

**NOTE:** I intend to add the first two commits in another PR.


wismill:
The idea is to create nodes that contain the overlapping results using a previously unused slot. 

Possible enhancements:
- Naming
- Compose sequence iterator
- Doc
- Integration in toolkits (Qt in particular)

It would be great to be able to test this in Qt apps to test the design, but I will not do it myself.

wismill:
I asked [feedback](https://bugreports.qt.io/browse/QTBUG-118698) to Qt devs.

wismill:
Not too happy with the `XKB_COMPOSE_CANDIDATE_ACCEPTED` state. But if we return just `XKB_COMPOSE_COMPOSED`, then the client should handle the last keysym as not part of the sequence.

wismill:
I sent a [merge request](https://codereview.qt-project.org/c/qt/qtbase/+/516630) in Qt.

PR #399, opened 2023-11-01
Compose: add support to all interactive tools

wismill:
Currently only `interactive-evdev` has support for compose. Let’s add X11 and Wayland.


bluetech (approved):
Nice!

PR #402, opened 2023-11-07
Promote tools/compose to xkbcli-compile-compose

### LINKED ISSUE #381: Make tools/compose part of the xkbcli commands
This is a follow-up of #367.

Convert `tools/compose.c` to `xkbcli compose`:

- Better names for the `--locale` arguments.
- ~Add `--format` (currently only text-v1).~ add it only when there is a new format?
- Check return value of `printfs`.
- Add to `xkbcli`.
- Test it in `test/tool-option-parsing.py`.
- Write `man` page

wismill:
Fixes #381

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -460,6 +460,7 @@ if build_tools
                                        install: true,
                                        install_dir: dir_libexec)
     install_man('tools/xkbcli-compile-keymap.1')
+    configh_data.set10('HAVE_XKBCLI_COMPILE_KEYMAP', true)
     # The same tool again, but with access to some private APIs.
     executable('compile-keymap',
                'tools/compile-keymap.c',
@@ -470,12 +471,13 @@ if build_tools
                install: false)
 
     # Tool: compose
-    executable('compose',
-               'tools/compose.c',
+    executable('xkbcli-compile-compose',
+               'tools/compile-compose.c',
                dependencies: tools_dep,
-               include_directories: [include_directories('src', 'include')],
-               install: false)
-    configh_data.set10('HAVE_XKBCLI_COMPILE_KEYMAP', true)
+               install: true,
+               install_dir: dir_libexec)
+    install_man('tools/xkbcli-compile-compose.1')
+    configh_data.set10('HAVE_XKBCLI_COMPILE_COMPOSE', true)
 
     # Tool: how-to-type
     executable('xkbcli-how-to-type',
--- a/src/compose/dump.h
+++ b/src/compose/dump.h
@@ -1,0 +1,95 @@
+/*
+ * Copyright © 2023 Pierre Le Marre <dev@wismill.eu>
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice (including the next
+ * paragraph) shall be included in all copies or substantial portions of the
+ * Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
+ * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+ * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+ * DEALINGS IN THE SOFTWARE.
+ */
+
+
+#ifndef COMPOSE_DUMP_H
+#define COMPOSE_DUMP_H
+
+#include "config.h"
+
+#include <stdlib.h>
+
+#include "src/utils.h"
+
+/* Ad-hoc escaping for UTF-8 string
+ *
+ * Note that it only escapes the strict minimum to get a valid Compose file.
+ * It also escapes hexadecimal digits after an hexadecimal escape. This is not
+ * strictly needed by the current implementation: "\x0abcg" parses as "␊bcg",
+ * but better be cautious than sorry and produce "\x0a\x62\x63g" instead.
+ * In the latter string there is no ambiguity and no need to know the maximum
+ * number of digits supported by the escape sequence.
+ */
+static inline char*
+escape_utf8_string_literal(const char *from)
+{
+    const size_t length = strlen(from);
+    /* Longest escape is converting ASCII character to "\xNN" */
+    char* to = calloc(4 * length + 1, sizeof(to));
+    if (!to)
+        return NULL;
+
+    size_t t = 0;
+    bool previous_is_hex_escape = false;
+    uint8_t nbytes = 0;
+    for (size_t f = 0; f < length;) {
+        if ((unsigned char) from[f] < 0x80) {
+            /* ASCII */
+            if (from[f] <= 0x10 || from[f] == 0x7f ||
+                (is_xdigit(from[f]) && previous_is_hex_escape))
+            {
+                /* Control character or
+                   hexadecimal digit following an hexadecimal escape */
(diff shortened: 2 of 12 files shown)


tools/compile-compose.c
```diff
@@ -69,10 +147,17 @@ print_compose_table_entry(struct xkb_compose_table_entry *entry)
...
     if (*utf8 != '\0') {
-        printf(" \"%s\"", utf8);
+        if (!escape_utf8_string(utf8, buf, sizeof(buf))) {
+            fprintf(stderr,
+                    "ERROR: Cannot escape the string: "
+                    "unsufficient buffer size.\n");
+            printf("\"\"");
+        } else {
+            printf(" \"%s\"", buf);
+        }
```
whot, line 156: Re-using `buf` here feels naughty. Any reason we can't just return an allocated buffer from `escape_utf8_string()` and worry about greater things in life than bufsizes? :)


If you alloc a buffer `4 * strlen(from) + 1` then our string will definitely fit and we can realloc to the smaller size before return.

tools/xkbcli-bash-completion.sh
```diff
@@ -95,6 +95,8 @@ ___xkbcli_subcommand()
     # • https://github.com/scop/bash-completion/blob/fdf4456186eb4548ef628e65fb1be73d8e4695e9/bash_completion.d/000_bash_completion_compat.bash#L311
     # We need both as the current help messages adopt both GNU and BSD styles.
     opts=$(_parse_usage xkbcli "$1 --help")
+    opts+="
+    "
```
whot, line 99: Can you add a comment to the commit message to explain what it fixes? Some error message maybe?

meson.build
```diff
@@ -460,6 +460,7 @@ if build_tools
                                        install: true,
                                        install_dir: dir_libexec)
     install_man('tools/xkbcli-compile-keymap.1')
+    configh_data.set10('HAVE_XKBCLI_COMPILE_KEYMAP', true)
```
whot, line 463: This needs to be a separate commit, it seems unrelated to the compose tool here.

whot (approved):
nitpicks, overall LGTM, thanks. @bluetech may have a comment on why the locale handling was implemented like that first though.

tools/compile-compose.c
```diff
@@ -56,6 +59,81 @@ usage(FILE *fp, char *progname)
...
+            f += 4;
+            previous_hex_escape = false;
+        }
+    }
+    if (t >= size)
+        return false;
+    to[t] = '\0';
+    return true;
+}
+
```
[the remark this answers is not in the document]
wismill: Done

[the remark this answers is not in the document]
wismill: Done. I added a test for random strings plus a few static strings.

wismill (commented):
@whot thanks for the review. I fixed the code, but I am not sure where to put the escape utility function.

whot:
> I am not sure where to put the escape utility function.

based on the filename `src/utf8.c` seems like a good match? :)

wismill:
I realized I did not push my changes…

> based on the filename `src/utf8.c` seems like a good match? :)

@whot I would prefer to keep `src/utf8.c` only related to low-level encoding. So I chose `src/compose/dump.h`. Note that it could be adapted to be used in other contexts: it would need a flag to indicate whether to use hexadecimal escapes  (compose-only) or octal ones.

src/compose/dump.h
```diff
@@ -0,0 +1,98 @@
...
+#ifndef COMPOSE_DUMP_H
+#define COMPOSE_DUMP_H
+
+#include "config.h"
+
+#include <stdlib.h>
+
+#include "src/utils.h"
+
+static char*
```
whot, line 44:
```suggestion
static inline char*
```

Not needed right now but it's a common workaround to stop compiler warnings about unused statics if this header file is included somewhere but not all functions in it are used.

src/compose/dump.h
```diff
@@ -0,0 +1,98 @@
...
+            nbytes = 4;
+        memcpy(&to[t], &from[f], nbytes);
+        t += nbytes;
+        f += nbytes;
+        previous_hex_escape = false;
+    }
+    to[t++] = '\0';
+    if (t < max_new_length)
+        return realloc(to, t);
+    return to;
```
whot, line 95:
```suggestion
   return realloc(to, t);
```
simpler and for 99.9% of all real-world strings we realloc anyway.

test/compose.c
```diff
@@ -769,20 +772,111 @@ test_traverse(struct xkb_context *ctx)
...
+
+    /* Test round-trip of random strings */
+#   define SAMPLE_SIZE 1000
+#   define MIN_CODE_POINT 0x0001
+#   define MAX_CODE_POINTS_COUNT 15
+    char buf[1 + MAX_CODE_POINTS_COUNT * 4];
+    for (int ascii = 1; ascii >= 0; ascii--) {
+        for (size_t s = 0; s < SAMPLE_SIZE; s++) {
+            /* Create the string */
+            size_t length = 1 + (rand() % MAX_CODE_POINTS_COUNT);
```
whot, line 827: if we're using `rand()` in tests, best to print the seed to the logs so it can be reproduced - otherwise you just get an error with no idea how to get to that same point again. So something like this in `main`

```c
int seed = time();
printf("Using random seed: %d\n", seed);
srand(seed);
```

whot (commented):
sorry, got no focus today so only two comments but I trust you enough that this will be fine.

test/compose.c
```diff
@@ -769,20 +772,111 @@ test_traverse(struct xkb_context *ctx)
...
+                /* Handle invalid code point in UTF-8 */
+                if (!count)
+                    /* Use replacement character: � */
+                    count = utf32_to_utf8(0xfffd, &buf[c]);
+                c += count - 1;
+                assert(c <= sizeof(buf) - 1);
+            }
+            assert(buf[c] == '\0');
+            assert(strlen(buf) == c);
+            assert(is_valid_utf8(buf, c));
```
whot, line 841: unrelated to this series: we should have something like
```c
#define assert_printf(cond, ...) \
   if (!(cond)) { \
      printf("Assertion failure: __VA_ARGS__); \
      assert(cond); \
   }
```
to aid with debugging assertion failures.

wismill (commented):
Fixed reviews comments and added a few comments.

@bluetech could you check my refactoring of locale handling?

tools/compile-compose.c
```diff
@@ -48,12 +48,8 @@ usage(FILE *fp, char *progname)
             " --file FILE\n"
             "    Specify a Compose file to load\n"
             " --locale LOCALE\n"
-            "    Specify the locale directly\n"
-            " --locale-from-env\n"
```
bluetech: The reason for adding this was mostly for testing vs. setlocale, but I agree it doesn't serve a purpose anymore.

bluetech (approved):
Nice, it would be nice to have `xkbcli compile-compose`.

I left a few minor comments, but LGTM (feel free to merge after fixing).

src/compose/dump.h
```diff
@@ -0,0 +1,95 @@
+/*
+ * Copyright © 2013 Pierre Le Marre <dev@wismill.eu>
```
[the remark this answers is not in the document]
wismill: 😅

src/compose/dump.h
```diff
@@ -0,0 +1,95 @@
...
+ * It also escapes hexadecimal digits after an hexadecimal escape. This is not
+ * strictly needed by the current implementation: "\x0abcg" parses as "␊bcg",
+ * but better be cautious than sorry and produce "\x0a\x62\x63g" instead.
+ * In the latter string there is no ambiguity and no need to know the maximum
+ * number of digits supported by the escape sequence.
+ */
+static inline char*
+escape_utf8_string_literal(const char *from)
+{
+    const int length = strlen(from);
```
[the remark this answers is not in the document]
wismill: 👍

src/compose/parser.c
```diff
@@ -468,6 +468,22 @@ resolve_modifier(const char *name)
     return XKB_MOD_INVALID;
 }
 
+/* For testing purposes */
```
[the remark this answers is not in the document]
wismill: Good idea

tools/compile-compose.c
```diff
@@ -69,10 +69,17 @@ print_compose_table_entry(struct xkb_compose_table_entry *entry)
...
         }
     }
-    printf(":");
+    printf(" : ");
     const char *utf8 = xkb_compose_table_entry_utf8(entry);
     if (*utf8 != '\0') {
-        printf(" \"%s\"", utf8);
+        char *escaped = escape_utf8_string_literal(utf8);
+        if (!escaped) {
+            fprintf(stderr, "ERROR: Cannot escape the string: allocation error\n");
```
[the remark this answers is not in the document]
wismill: Good point! And I realized that the current exit code is always… non-zero! Fixed.

tools/compile-compose.c
wismill: Duly noted. I modified the commit message to explain that.

PR #406, opened 2023-11-16
registry: Parse “popularity” attribute

### LINKED ISSUE #278: registry ignores XML popularity attribute
[registry.c](https://github.com/xkbcommon/libxkbcommon/blob/master/src/registry.c) ignores the [popularity attribute](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/blob/master/rules/xkb.dtd) in XML, instead setting popularity exotic if the file name contains 'extras'.

Backwards compatibility suggests to continue to assume exotic in extras, but an improvement could be to read the popularity attribute in base.

See also https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/76

wismill:
Previously the attribute “popularity” was completely ignored. It also did not respect the modified DTD, because its default value depends if we are loading an “extras” rules file.

Fixed:
- Always parse the popularity attribute.
- Change the DTD to reflect that the default value is implied.

Fixes #278

### DIFF
--- a/src/registry.c
+++ b/src/registry.c
@@ -708,44 +708,73 @@ extract_text(xmlNode *node)
     return NULL;
 }
 
+/* Data from “configItem” node */
+struct config_item {
+    char *name;
+    char *description;
+    char *brief;
+    char *vendor;
+    enum rxkb_popularity popularity;
+};
+
+#define config_item_new(popularity_) { \
+    .name = NULL, \
+    .description = NULL, \
+    .brief = NULL, \
+    .vendor = NULL, \
+    .popularity = popularity_ \
+}
+
+static void
+config_item_free(struct config_item *config) {
+    free(config->name);
+    free(config->description);
+    free(config->brief);
+    free(config->vendor);
+}
+
 static bool
-parse_config_item(struct rxkb_context *ctx,
-                  xmlNode *parent,
-                  char **name,
-                  char **description,
-                  char **brief,
-                  char **vendor)
+parse_config_item(struct rxkb_context *ctx, xmlNode *parent,
+                  struct config_item *config)
 {
     xmlNode *node = NULL;
     xmlNode *ci = NULL;
 
     for (ci = parent->children; ci; ci = ci->next) {
         if (is_node(ci, "configItem")) {
-            *name = NULL;
-            *description = NULL;
-            *brief = NULL;
-            *vendor = NULL;
+            /* Process attributes */
+            xmlChar *raw_popularity = xmlGetProp(ci, (const xmlChar*)"popularity");
+            if (raw_popularity) {
+                if (xmlStrEqual(raw_popularity, (const xmlChar*)"standard"))
+                    config->popularity = RXKB_POPULARITY_STANDARD;
+                else if (xmlStrEqual(raw_popularity, (const xmlChar*)"exotic"))
+                    config->popularity = RXKB_POPULARITY_EXOTIC;
+                else
+                    log_err(ctx,
+                            "xml:%d: invalid popularity attribute: expected "
+                            "'standard' or 'exotic', got: '%s'\n",
+                            ci->line, raw_popularity);
+            }
+            xmlFree(raw_popularity);
 
+            /* Process children */
             for (node = ci->children; node; node = node->next) {
                 if (is_node(node, "name"))
-                    *name = extract_text(node);
+                    config->name = extract_text(node);
                 else if (is_node(node, "description"))
-                    *description = extract_text(node);
+                    config->description = extract_text(node);
                 else if (is_node(node, "shortDescription"))
-                    *brief = extract_text(node);
+                    config->brief = extract_text(node);
                 else if (is_node(node, "vendor"))
-                    *vendor = extract_text(node);
+                    config->vendor = extract_text(node);
                 /* Note: the DTD allows for vendor + brief but models only use
                  * vendor and everything else only uses shortDescription */
             }
 
-            if (!*name || !strlen(*name))  {
+            if (!config->name || !strlen(config->name))  {
                 log_err(ctx, "xml:%d: missing required element 'name'\n",
                         ci->line);
-                free(*name);
-                free(*description);
-                free(*brief);
-                free(*vendor);
+                config_item_free(config);
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


src/registry.c
```diff
@@ -721,6 +722,20 @@ parse_config_item(struct rxkb_context *ctx,
 
     for (ci = parent->children; ci; ci = ci->next) {
         if (is_node(ci, "configItem")) {
+            xmlChar *raw_popularity;
+            raw_popularity = xmlGetProp(ci, (const xmlChar*)"popularity");
```
whot, line 726:
```suggestion
            xmlChar *raw_popularity = xmlGetProp(ci, (const xmlChar*)"popularity");
```

src/registry.c
```diff
@@ -762,7 +777,9 @@ parse_model(struct rxkb_context *ctx, xmlNode *model,
 {
     char *name, *description, *brief, *vendor;
 
-    if (parse_config_item(ctx, model, &name, &description, &brief, &vendor)) {
+    if (parse_config_item(ctx, model, &popularity, &name,
+                          &description, &brief, &vendor))
+    {
```
whot, line 782:
```suggestion
                          &description, &brief, &vendor)) {
```

same for the others below

src/registry.c
```diff
@@ -711,6 +711,7 @@ extract_text(xmlNode *node)
 static bool
 parse_config_item(struct rxkb_context *ctx,
                   xmlNode *parent,
+                  enum rxkb_popularity *popularity,
                   char **name,
                   char **description,
                   char **brief,
```
whot, line 717: not directly related to your patch but with all the args it would be better to have a 

```c
struct config_item {
   char *name;
   ...
}
```
and return that, together with a `config_item_free()` that moves the code into a single location. This will clean up the code a bit.

src/registry.c
```diff
@@ -1146,9 +1169,11 @@ validate(struct rxkb_context *ctx, xmlDoc *doc)
...
     xmlDtd *dtd = NULL;
     xmlParserInputBufferPtr buf = NULL;
-    /* This is a modified version of the xkeyboard-config xkb.dtd. That one
-     * requires modelList, layoutList and optionList, we
-     * allow for any of those to be missing.
+    /* This is a modified version of the xkeyboard-config xkb.dtd:
+     * • xkeyboard-config requires modelList, layoutList and optionList,
+     *   but we allow for any of those to be missing.
+     * • xkeyboard-config sets default value of “popularity” to “standard”,
+     *   but we set this value depending if we are loading an “extras” rule file.
```
whot, line 1176:
```suggestion
     *   but we set this value depending if we are currently parsing an “extras” rule file.
```
slightly less ambiguous

test/registry.c
```diff
@@ -74,17 +75,35 @@ struct test_option_group {
     struct test_option options[10];
 };
 
+#define POPULARITY_NONE 0
+static const char *
+popularity_attr(enum rxkb_popularity popularity)
+{
+    switch ((int) popularity) {
```
whot, line 82:
```suggestion
    switch ((int) popularity) {
```
is the `(int)` required here? technically this removes a little bit of type-safety and we have little enough of that in C as it is :)

whot (approved):
side-note: the reason I skipped this in the registry loading is because at the time xkeyboard-config didn't always set it (iirc) and it's somewhat superfluous since xk-c also guarantees that there's no exotic in the default base.xml file, see the `test_rules.py` there.

A few style nitpicks but this LGTM to me otherwise. The `struct config_item` change would be nice if you find the time for it.

src/registry.c
wismill: Yep, I am all for cleaning up!

src/registry.c
wismill: Done

src/registry.c
wismill: Fixed

test/registry.c
wismill: I love type-safety, but here I faced the issue that `POPULARITY_NONE` is not a member of the enum and raises a compiler warning. Well, the next option is to deal with it in the `default` case.

wismill (commented):
Fixed review comments.

test/registry.c
whot: hmm, I wonder if it makes sense to work around this by doing a 
```
enum popularity {
   POP_STANDARD = POPULARITY_STANDARD,
   POP_EXOTIC = POPULARITY_EXOTIC,
   POP_UNDEFINED = PROP_EXOTIC + 1
}
```
and using that. That would also allow you to return a `struct config_item` from `parse_config_item()` and override the popularity if `POP_UNDEFINED`. All without leaking the undefined value into the public API.

Then again, the public API was explicitly designed so we can use `0` as none case, so all this could be over the top :)

wismill: I propose the following:
- Use the idea of the new enum in test. We ensure we keep up-to-date with assertions.
- Using an extended enum in `struct config_item` would require to also check we keep up-to-date, then add a test at _each_ call of `parse_config_item`. I think we can do simpler: just initialize all members of `struct config_item` in the callers of `parse_config_item`, and not just popularity as previously.

I updated the code to implement this. What do you think?

bluetech (approved):
Only had a cursory look (I defer to @whot), but LGTM.

src/registry.c
```diff
@@ -956,17 +971,14 @@ parse_layout(struct rxkb_context *ctx, xmlNode *layout,
...
         list_init(&l->iso3166s);
-        l->name = name;
+        l->name = config.name;
         l->variant = NULL;
-        l->description = description;
-        l->brief = brief;
-        l->popularity = popularity;
+        l->description = config.description;
+        l->brief = config.brief;
+        l->popularity = config.popularity;
```
whot, line 978: we totally need a [g_steal_pointer](https://docs.gtk.org/glib/func.steal_pointer.html) or copy libei's [steal](https://gitlab.freedesktop.org/libinput/libei/-/blob/main/src/util-mem.h#L92) macro for the same effect. would allow us to just use an unconditional cleanup path.

whot (approved):
Thanks, this LGTM!

src/registry.c
wismill: I will have a look, but let’s do this in another PR.

wismill: I gave it a try in #407.

PR #407, opened 2023-11-21
Utils: steal `steal` from libei

wismill:
Follow-up of [Peter’s suggestion](https://github.com/xkbcommon/libxkbcommon/pull/406#discussion_r1398695593).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -364,6 +364,7 @@ if get_option('enable-xkbregistry')
         'src/utils.c',
         'src/util-list.h',
         'src/util-list.c',
+        'src/util-mem.h',
     ]
     libxkbregistry_link_args = []
     libxkbregistry_link_deps = []
--- a/meson.build
+++ b/meson.build
@@ -228,6 +228,7 @@ libxkbcommon_sources = [
     'src/text.h',
     'src/utf8.c',
     'src/utf8.h',
+    'src/util-mem.h',
     'src/utils.c',
     'src/utils.h',
 ]
--- a/src/registry.c
+++ b/src/registry.c
@@ -34,6 +34,7 @@
 #include "xkbcommon/xkbregistry.h"
 #include "utils.h"
 #include "util-list.h"
+#include "util-mem.h"
 
 struct rxkb_object;
 
@@ -803,9 +804,9 @@ parse_model(struct rxkb_context *ctx, xmlNode *model,
 
         /* new model */
         m = rxkb_model_create(&ctx->base);
-        m->name = config.name;
-        m->description = config.description;
-        m->vendor = config.vendor;
+        m->name = steal(&config.name);
+        m->description = steal(&config.description);
+        m->vendor = steal(&config.vendor);
         m->popularity = config.popularity;
         list_append(&ctx->models, &m->base.link);
     }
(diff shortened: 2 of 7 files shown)


whot (approved):
quite disappointed you didn't say "steal steal from libei" but oh well :)

LGTM, thanks!

whot:
what's the compiler warning that is introduced here?

bluetech:
I think we should only steal `steal`, since I doubt `__attribute__((cleanup(_x)))` works on msvc, and we don't use `x*alloc` functions. I'm also surprised `__typeof__` works, but I see we're already using it in `util-list` and we didn't get any complaints, so it's OK.

wismill:
> quite disappointed you didn't say "steal steal from libei" but oh well :)

@whot Good one! Fixed 😉

> what's the compiler warning that is introduced here?

Which one?

> I think we should only steal `steal`

@bluetech I removed any function in `util-mem.h` that we do not use. Now I am not sure we should have a separate file with only one public macro.

@whot if we were to merge the now tiny `util-mem.h` or migrate memory-related function to it, should we keep the current license or would you agree to use license in e.g. `utils.h` and add you as author?

whot:
> > what's the compiler warning that is introduced here?
> 
> Which one?

You added the compiler warning label to this MR, so I figured there may be a compiler warning :)

> would you agree to use license in e.g. utils.h and add you as author?

the licenses are the same anyway but if there's a minor difference somewhere - yes, I agree to relicense to whatever specific wording libxkbcommon uses.

wismill:
> You added the compiler warning label to this MR, so I figured there may be a compiler warning :)

@whot Ah ok, now I understand. Well, there are _some_ warnings, but they are unrelated. I chose this label because there is currently none for memory handling improvement. Now I see this is confusing.

About the new file `util-mem.h`: I think it makes sense to keep it and move the memory-related functions `utils.h` to `util-mem.h` (in another PR).

wismill:
I added a new commit that use `steal` in `xkbcomp` files. I did not find other occurrences of the pattern:
```c
a = b;
b = NULL;
```

but of course the simple regex I used cannot spot all the occurrences.

OTOH a recurrent pattern I saw is stealing a `darray` and re-initializing the source. We do have `darray_steal`, but it only steals the array, not the whole struct. I think it is worth another PR:
1. Rename `darray_steal` to e.g. `darray_steal_array`
2. Create a new macro `darray_steal` that steals the `darray` and re-initialize the source.

wismill:
> I'm also surprised `__typeof__` works, but I see we're already using it in `util-list` and we didn't get any complaints, so it's OK.

@bluetech it does not work 😕 We did not catch this in the CI because it is only used in the registry, which is not built on Windows… I am looking for a fix.

whot:
> We loose some type safety, but is there another option other than switching to C23?

IMO this doesn't matter too much here anyway because the type-safety still exists on other platforms so any issues we'll pick up there.

whot:
> but of course the simple regex I used cannot spot all the occurrences.

if you have too much time on your hand, you can learn [coccinelle](https://www.kernel.org/doc/html/v4.19/dev-tools/coccinelle.html) :)

whot (approved):

bluetech (approved):

PR #415, opened 2023-12-05
Keysyms: fix xkb_keysym_is_modifier

### LINKED ISSUE #413: Compose sequence terminates after pressing ISO_LEVEL5_SHIFT
Hi y'all,

I have an issue with the compose key when using neo layout. The compose sequence stops when pressing Alt Gr/<> aka ISO_LEVEL5_SHIFT. This happens when I use kitty on sway but not in gkx (The gnome console) on sway.
This makes all entries in the compose map requiring LEVEL5 impossible to type.

The issues becomes apparent when using "xkbcli interactive-evdev --layout de --variant neo --enable-compose" on a tty:
The key combo (I use Qwerty notation here): Left Caps + Tab, a,e produces æ.
The combo Left Caps + Tab, Left Shift(press), Left Shift(release), a, e produces also æ i.e. the "accidental" shift press does not terminate the compose sequence.
However, Caps + Tab, Alt Gr(press), Alt Gr(release),a produces immediately an 'a' like the composing was terminated. Pressing 'e' afterward just adds an 'e' as one would expect then.

My guess is that xkb somehow intercepts the Alt Gr or tries to interpret it as part of a sequence instead of a pure shift modifier.
The same happens by the way if instead of "Alt Gr" the key "<>" is used, which are the two LEVEL5_SHIFT keys in neo.

I'd really like to be able to use my whole .XCompose in all applications. Is it possible by just a config change?

Regards

wismill:
Fixes #413 and add corresponding tests. Also add some tests for keypad keysyms.

This is based on #414; we should wait for it to be merged.

### DIFF
--- a/test/keysym.c
+++ b/test/keysym.c
@@ -23,9 +23,85 @@
 #include "config.h"
 
 #include <locale.h>
+#include <stdbool.h>
 
 #include "test.h"
 #include "keysym.h" /* For unexported is_lower/upper/keypad() */
+
+/* Explicit ordered list of modifier keysyms */
+static const xkb_keysym_t modifier_keysyms[] = {
+    XKB_KEY_ISO_Lock,
+    XKB_KEY_ISO_Level2_Latch,
+    XKB_KEY_ISO_Level3_Shift,
+    XKB_KEY_ISO_Level3_Latch,
+    XKB_KEY_ISO_Level3_Lock,
+    /* XKB_KEY_ISO_Group_Shift == XKB_KEY_Mode_switch */
+    XKB_KEY_ISO_Group_Latch,
+    XKB_KEY_ISO_Group_Lock,
+    XKB_KEY_ISO_Next_Group,
+    XKB_KEY_ISO_Next_Group_Lock,
+    XKB_KEY_ISO_Prev_Group,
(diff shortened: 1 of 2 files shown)


test/keysym.c
```diff
@@ -23,10 +23,93 @@
 #include "config.h"
 
 #include <locale.h>
+#include <stdbool.h>
 
 #include "test.h"
 #include "keysym.h" /* For unexported is_lower/upper/keypad() */
 
+static const xkb_keysym_t modifier_keysyms[] = {
+
```
whot, line 32:
```suggestion
```

whot (approved):
Only looked at the top two commits (and only skim-reviewed those) but they LGTM, thanks!

bluetech:
Thanks @wismill, the modifier fixes LGTM. Apologies for getting it wrong 10 years ago :)

Can you explain why this depends on #414? I see `assert(!xkb_keysym_is_assigned(0xfe10));` but I don't think all of the complexity of #414 is worth it just for that...

wismill:
@bluetech the fix itself does not depend on #414, but the tests do: they require the keysym iterator. I did not expect #414 to be problematic. I should probably reverse the PR dependency then.

wismill:
Rebased.

PR #416, opened 2023-12-05
Compose: explicit result string size

wismill:
In the spirit than #414, add explicit bound for compose result string.

This is based on #414; we should wait for it to be merged.


whot (commented):
Same argument here as in #414 , I don't think a buffer size should be part of the API but do feel free to use it as internal constant.

wismill:
I agree. Rebased & fixed.

wismill:
Rebased.

whot (approved):

PR #417, opened 2023-12-07
Keysyms: Fix case mapping

wismill:
There are a few mistakes in the function `XConvertCase`.

Fixed the upper case mappings for:
- `XKB_KEY_ydiaeresis`
- `XKB_KEY_mu`
- `XKB_KEY_ssharp`

Also added ICU optional test.

Notes:
- `xkb_keysym_is_upper` is interpreted as matching the disjunction of the Unicode character properties “Uppercase” or “Titlecase”.
- In Unicode, the upper case of “ß” (`U+00DF`) is *not* “ẞ” (`U+1E9E`) but “SS”. “ẞ” is reserved for text in capitals.

This MR depends on #414 and should be rebased on it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -651,10 +651,18 @@ if get_option('enable-x11')
         ],
     )
 endif
+# TODO: version range?
+keysyms_test_dep = [test_dep]
+keysyms_test_c_args = ['-DENABLE_PRIVATE_APIS']
+icu_dep = dependency('icu-uc', required: false)
+if icu_dep.found()
+    keysyms_test_dep += [icu_dep]
+    configh_data.set10('HAVE_ICU', true)
+endif
 test(
     'keysym',
-    executable('test-keysym', 'test/keysym.c', dependencies: test_dep,
-               c_args: ['-DENABLE_PRIVATE_APIS']),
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -651,10 +651,18 @@ if get_option('enable-x11')
         ],
     )
 endif
+# TODO: version range?
+icu_dep = dependency('icu-uc', required: false)
+keysyms_test_dep = [test_dep]
+keysyms_test_c_args = ['-DENABLE_PRIVATE_APIS']
+if icu_dep.found()
+    keysyms_test_dep += [icu_dep]
+    keysyms_test_c_args += ['-DHAVE_ICU']
```
whot, line 660: I think this should be 
```
configh_data.set('HAVE_ICU', 1)
```
and it then gets pulled in via `config.h`

(we also seem to have a mix of `#if` and `#ifdef` for our own constants...)

whot (approved):
minor meson.build-related nitpick, LGTM otherwise, thanks

meson.build
wismill: Rebased & fixed.

> (we also seem to have a mix of `#if` and `#ifdef` for our own constants...)
I opted for just `#if`, as it is the most used.

wismill:
Rebased

PR #419, opened 2023-12-12
Update meson.build error message when enable-wayland enabled

arfoll:
Fix error message version to match the version check

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -535,7 +535,7 @@ if build_tools
         wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.12', required: false)
         wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
         if not wayland_client_dep.found() or not wayland_protocols_dep.found() or not wayland_scanner_dep.found()
-            error('''The Wayland xkbcli programs require wayland-client >= 1.2.0, wayland-protocols >= 1.12 which were not found.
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -535,7 +535,7 @@ if build_tools
         wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.12', required: false)
         wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
         if not wayland_client_dep.found() or not wayland_protocols_dep.found() or not wayland_scanner_dep.found()
-            error('''The Wayland xkbcli programs require wayland-client >= 1.2.0, wayland-protocols >= 1.7 which were not found.
+            error('''The Wayland xkbcli programs require wayland-client >= 1.2.0, wayland-protocols >= 1.12 which were not found.
```
whot, line 538:
```suggestion
            error('''The Wayland xkbcli programs require wayland-client and wayland-protocols which were not found.
```
I'd argue to drop the versions here because meson will print something like this anyway:
```
Dependency wayland-protocols found: NO found 1.11 but need: '>=1.12'
```

arfoll:
Absolutely fair, your suggestion is easier to maintain - I just stumbled along the issue and am happy with whatever that reduces the confusion that this caused me :)

whot:
squashed and merged, thanks

PR #421, opened 2023-12-18
test: add a test for multiple keysyms (and some minimal docs)

whot:
I couldn't find any reference to *how* the keymap format actually needs to look like if you want multiple keysyms per level. So let's add a test for it and a minimal documentation entry.

### DIFF
--- a/doc/keymap-format-text-v1.md
+++ b/doc/keymap-format-text-v1.md
@@ -910,6 +910,17 @@ and separated by commas. Each element of the symbol arrays corresponds to a
 different modifier level. In this example, the symbol (keysym) `XKB_KEY_q` for
 level 1 and `XKB_KEY_Q` for level 2.
 
+As an extension to the XKB format, libxkbcommon supports multiple key symbols
+per level.
+
+    key <AD01> { [ {a, b}, Q ] };
+
+In this example, the keycode `<AD01>` produces two symbols on level 1
+(`XKB_KEY_a` and `XKB_KEY_b`) and one symbol (`XKB_KEY_Q`) on level 2.
+
+@warning Keymaps containing multiple key symbols per level are not supported
+by the various X11-related tools (`setxkbmap`, `xkbcomp`, etc.).
+
 #### Actions
 
 @todo how to bind key actions
--- a/test/keymap.c
+++ b/test/keymap.c
@@ -198,12 +198,45 @@ test_numeric_keysyms(void)
     xkb_context_unref(context);
 }
 
+static void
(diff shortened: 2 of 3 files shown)


doc/keymap-format-text-v1.md
```diff
@@ -910,6 +910,16 @@ and separated by commas. Each element of the symbol arrays corresponds to a
 different modifier level. In this example, the symbol (keysym) `XKB_KEY_q` for
 level 1 and `XKB_KEY_Q` for level 2.
 
+As an extension to the XKB format, libxkbcommon supports multiple key symbols
+per level.
+
+    key <AD01> { [ {a, b}, Q ] };
+
+In this example, the keycode `<AD01>` produces two symbols on level 1
+(`XKB_KEY_a` and `XKB_KEY_b1)` and one symbol (`XKB_KEY_Q`) on level 2.
```
wismill, line 919:
```suggestion
(`XKB_KEY_a` and `XKB_KEY_b)` and one symbol (`XKB_KEY_Q`) on level 2.
```

doc/keymap-format-text-v1.md
```diff
@@ -910,6 +910,16 @@ and separated by commas. Each element of the symbol arrays corresponds to a
...
+Note that keymaps containing multiple key symbols per level are not supported
```
wismill, line 920:
```suggestion
@warning Keymaps containing multiple key symbols per level are not supported
```

test/keymap.c
```diff
@@ -198,12 +198,45 @@ test_numeric_keysyms(void)
     xkb_context_unref(context);
 }
 
+static void
+test_multiple_keysyms(void)
```
wismill, line 202:
```suggestion
test_multiple_keysyms_per_level(void)
```

wismill (approved):
Good idea! I read your comment on `xkeyboard-config`: we should check multiple actions per level too. Probably another MR.

whot:
All fixed, thanks!

>  we should check multiple actions per level too. Probably another MR.

yes, we should. but not here because I'm afraid of having to add this to my todo list if it doesn't work and that one is already exploding :)

wismill:
Actually multiple actions per level is not currently supported by the parser: not in symbols nor in compatibility files.

bam80:
Thanks for addressing this!

Referencing original issue:
https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/428

> Actually multiple actions per level is not currently supported by the parser: not in symbols nor in compatibility files.

Any hope for this to be implemented?

PR #423, opened 2023-12-27
Fix crash when an element of SymbolsBody is empty

chiro:
An empty element is allowed in SymbolsBody definition, so the following keymap is gramatically correct.

```
xkb_keymap {
  xkb_symbols "sym" {
    key <SPC> {, [Space] };
  };
};
```

However, the current parser crashes with the keymap due to null pointer access.
This change fixes the crash.

### DIFF
--- a/test/data/keymaps/empty-symbol-decl.xkb
+++ b/test/data/keymaps/empty-symbol-decl.xkb
@@ -1,0 +1,10 @@
+xkb_keymap {
+    xkb_keycodes  { include "evdev+aliases(qwerty)" };
+    xkb_types     { include "complete" };
+    xkb_compat    { include "complete" };
+    xkb_symbols "sym" {
+        // This syntax caused the crash in parser before, and is not accepted
+        // anymore.
+        key <SPC> { , [Space] };
+    };
+};
--- a/test/filecomp.c
+++ b/test/filecomp.c
@@ -54,6 +54,7 @@ main(void)
     assert(!test_file(ctx, "keymaps/bad.xkb"));
     assert(!test_file(ctx, "keymaps/syntax-error.xkb"));
     assert(!test_file(ctx, "keymaps/syntax-error2.xkb"));
+    assert(!test_file(ctx, "keymaps/empty-symbol-decl.xkb"));
     assert(!test_file(ctx, "does not exist"));
 
     /* Test response to invalid flags and formats. */
--- a/src/xkbcomp/parser.y
+++ b/src/xkbcomp/parser.y
@@ -465,7 +465,6 @@ SymbolsBody     :       SymbolsBody COMMA SymbolsVarDecl
                         { $$.head = $1.head; $$.last->common.next = &$3->common; $$.last = $3; }
                 |       SymbolsVarDecl
                         { $$.head = $$.last = $1; }
-                |       { $$.head = $$.last = NULL; }
                 ;
 
 SymbolsVarDecl  :       Lhs EQUALS Expr         { $$ = VarCreate($1, $3); }


wismill:
Thank you for the bug report.

I confirm this syntax is accepted by `xkbcomp` and is interpreted as if there was no leading comma.

But I think this is an oversight, because this syntax is useless. As [Ivan Pascal wrote](https://web.archive.org/web/20190723233834/http://pascal.tsu.ru/en/xkb/gram-symbols.html#brackets), the intended syntax would rather be:

```c
key <SPC> {[], [Space] };
```

which parses correctly in `xkbcommon`.

Anyway, a segmentation fault is definitly not acceptable!

test/data/keymaps/empty-symbol-decl.xkb
```diff
@@ -0,0 +1,5 @@
+xkb_keymap {
+  xkb_symbols "sym" {
+    key <SPC> { [Space] };
+  };
+};
```
wismill, line 5: The test suite currently works because this is not a valid _keymap_ file, only a symbols file. Also, we want to check that the syntax that caused segfault previously is now parsed correctly.

```suggestion
xkb_keymap {
    xkb_keycodes  { include "evdev+aliases(qwerty)" };
    xkb_types     { include "complete" };
    xkb_compat    { include "complete" };
    xkb_symbols "sym" {
        key <SPC> { , [Space] };
    };
};
```

test/filecomp.c
```diff
@@ -54,6 +54,7 @@ main(void)
     assert(!test_file(ctx, "keymaps/bad.xkb"));
     assert(!test_file(ctx, "keymaps/syntax-error.xkb"));
     assert(!test_file(ctx, "keymaps/syntax-error2.xkb"));
+    assert(!test_file(ctx, "keymaps/empty-symbol-decl.xkb"));
```
wismill, line 57:
```suggestion
    assert(test_file(ctx, "keymaps/empty-symbol-decl.xkb"));
```

The test should pass! Please move this above.

wismill (changes requested):
@chiro The fix LGTM, but the test needs some fixes, please look at my suggestions.

@whot @bluetech this PR would allow to parse input currently accepted by `xkbcomp`, but I think it really should not be accepted there either.

test/data/keymaps/empty-symbol-decl.xkb
chiro: Oops, thank you for pointing it out! Fixed.

test/filecomp.c
chiro: Fixed!

chiro:
@wismill Thanks for the prompt review! I fixed the test. PTAL again.

I agree that the parser rule is not useful. However changing it introduces backward incompatibility, so I decided to fix the crash for now.

test/data/keymaps/empty-symbol-decl.xkb
```diff
@@ -0,0 +1,8 @@
+xkb_keymap {
+    xkb_keycodes  { include "evdev+aliases(qwerty)" };
+    xkb_types     { include "complete" };
+    xkb_compat    { include "complete" };
+    xkb_symbols "sym" {
+        key <SPC> { , [Space] };
```
bluetech, line 6:
```suggestion
        // This dubious syntax is accepted by xkbcomp so we accept it for
        // backward compatibility. The preferred way to write it is:
        // key <SPC> { [], [Space] };
        key <SPC> { , [Space] };
```

bluetech (approved):
Thanks @chiro, LGTM.

Agree it's not worth breaking backward compat over.

I'll leave it to @wismill to make the final decision.

test/data/keymaps/empty-symbol-decl.xkb
wismill: Actually, these syntaxes are not equivalent. Using X11 `xkbcomp`, I get that 
```c
key <SPC> { [Space] };
key <SPC> { , [exclam] };
```
is equivalent to 
```c
key <SPC> { [exclam] };
```
whereas  
```c
key <SPC> { [Space] };
key <SPC> { [], [exclam] };
```
is equivalent to 
```c
key  <SPC> {
    symbols[Group1]= [          Space ],
    symbols[Group2]= [          exclam ]
};
```
That’s why I wrote it was probably an oversight.

chiro: Thanks for the clarification. then, is there any other changes needed?

wismill, line 6:
```suggestion
        // This dubious syntax is accepted by xkbcomp so we accept it for
        // backward compatibility. The preferred way to write it is:
        // key <SPC> { [Space] };
        key <SPC> { , [Space] };
```

Small fix to Ran’s suggestion.

wismill: @chiro  I wonder: where did you encounter this syntax in the first place? Unless there is a real-world proof, I think we should fix just the segfault, but disallow this syntax. It is useless per se and confusing. It is not used in `xkeyboard-config`.

I do not see to make `key <SPC> { , [exclam] };` equivalent to `key <SPC> { [], [exclam] };` either: XKB has C-style, so it would not be intuitive.

whot: Fully agree with @wismill here - if this is somewhere in the xkb maps we should fix it in xkeyboard-config but allowing this syntax seems like the wrong approach.

chiro: @wismill It was not found in an actual in-use layout file. It's detected in testing. As far as I know, there is no real-world use case of this syntax. I have no objection to disallowing this syntax.

chiro: > @wismill It was not found in an actual in-use layout file. It's detected in testing. As far as I know, there is no real-world use case of this syntax. I have no objection to disallowing this syntax.

Just to clarify, I couldn't find any real-world use case, but it doesn't mean there is no use case in the world. That's why I made the CL just to fix the crash.

wismill: @chiro we are grateful for uncovering the segfault. This is a nasty bug, and it would not occur to me to test this syntax. Well done!

I am curious, could you share how you stumble on this? Maybe we could then improve our test suite further.

Now, given that:
- AFAIK the syntax is useless and undocumented;
- `xkeyboard-config` compiles just fine with xkbcommon;
- no one opened a ticket before.

then I really doubt there are use cases in the world. While we strive to have maximum compatibility with X11, I think this is one _edge_ case we should not support. Probably worth a comment, at least in the code.

chiro: It was reported as a potential security issue to the project I'm working on (Actually it was not because only layout files in xkeyboard-config are used). I don't know the details of how it was discovered, but it should be by fuzzing.

whot: If it's not used in any existing xkeyboard-config files, can we just make it return a parser error? That seems like the best option here - making this intentionally backwards-compatible means we have to support it forever when there seems to be little point to it.

chiro: Thanks for the comment (and the reaction)! I changed this PR not to allow the syntax. Could you review it again? Thanks.

whot (approved):
LGTM now, @wismill can you merge it if you're happy with it? Thanks.

wismill:
Thanks @chiro !

chiro:
Thank you all for reviewing!

PR #427, opened 2024-01-10
Add ENABLE_PRIVATE_APIS guard for internal API

wismill:
Follow-up of #414. See [@whot comments](https://github.com/xkbcommon/libxkbcommon/pull/414#discussion_r1421835848)

### DIFF
--- a/src/keysym.c
+++ b/src/keysym.c
@@ -119,6 +119,8 @@ xkb_keysym_get_name(xkb_keysym_t ks, char *buffer, size_t size)
     return snprintf(buffer, size, "0x%08x", ks);
 }
 
+#ifdef ENABLE_PRIVATE_APIS
+
 bool
 xkb_keysym_is_assigned(xkb_keysym_t ks)
 {
@@ -217,6 +219,8 @@ xkb_keysym_iterator_next(struct xkb_keysym_iterator *iter)
     }
     return true;
 }
+
+#endif
 
 /*
  * Parse the numeric part of a 0xXXXX and UXXXX keysym.
--- a/src/keysym.h
+++ b/src/keysym.h
@@ -79,6 +79,8 @@
 /** Maximum keysym name length */
 #define XKB_KEYSYM_NAME_MAX_SIZE  27
 
+#ifdef ENABLE_PRIVATE_APIS
+
 bool
 xkb_keysym_is_assigned(xkb_keysym_t ks);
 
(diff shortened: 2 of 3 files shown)


whot (commented):
I had a look at `-DENABLE_PRIVATE_APIS` again (which apparently I introduced) and the original reason for it was that we have the same tool but with different functionality - the installed one links against libxkbcommon but the one in the builddir statically compiles against it. So that `#define` doesn't actually enable any private APIs, it allows the tools to use such an API (the difference might be hair-splitting).

More importantly though: this doesn't affect libxkbcommon.so because that one uses symbol visibility anyway and we shouldn't ever leak any symbols.

So my opinion here is still that we don't really need this, the only user of those symbols is the test which also statically links libxkbcommon.

I'm not sure if saving the 500 bytes is really worth it. I know floppies are no longer produced but still ;)

wismill:
@whot I agree the semantics may not be great. Maybe `ENABLE_TEST_API` is better? 

> I'm not sure if saving the 500 bytes is really worth it. I know floppies are no longer produced but still ;)

Sorry but I cannot resist: with this kind of reasoning we would end with fat libraries all around. Libraries that are downloaded millions (billions?) of times. I _do_ agree in our case the gain is small, but I do not understand why so much resistance. I mean, the patch is trivial and mean no real maintenance.

whot:
Developer mental load, primarily :) One more thing to keep track of even though it has effectively zero effect since the symbol is private anyway. Which means guessing why the ifdef is there when it has no effect, looking up the git history (if available, too many times now I find myself in code where that is not the case...), etc.

bluetech:
I'm surprised unused functions increase the .so size. I suspect it would go away if we use `--gc-sections`, e.g. from systemd (I just guessed they do this and indeed they do...):

```meson
if get_option('buildtype') != 'debug'
        possible_cc_flags += [
                '-ffunction-sections',
                '-fdata-sections',
        ]

        possible_link_flags += '-Wl,--gc-sections'
endif
```

https://github.com/systemd/systemd/blob/978e7d166c362302ce2302ac36efe9717d642d90/meson.build#L482-L489

wismill:
@bluetech I tried your suggestion in #508.

wismill:
Closing, as @whot [suggests](https://github.com/xkbcommon/libxkbcommon/pull/508#issuecomment-2363078727) that neither this PR nor #508 really fixes the issue.

PR #430, opened 2024-01-11
keysyms: Fast and complete case mappings (Unicode 15.1)

wismill:
The current code to handle keysym case mappings is quite complex and slow. It is also incomplete, as it does not cover recent Unicode database.

It would be easier if we were to use only a lookup table, but a trivial implementation would lead to a huge array: the cased characters range from `U+0041` to `U+1E943`, i.e. a span of 125 187 elements.

Thus we need some tricks to compress the lookup table. We base our work on the post: https://github.com/apankrat/notes/blob/3c551cb028595fd34046c5761fd12d1692576003/fast-case-conversion/README.md

The compression algorithm is roughly:
1. Compute the delta between the characters and their mappings.
2. Split the delta array in chunk of a given size.
3. Rearrange the order of the chunks in order to optimize consecutive
   chunks overlap.
4. Create a data table with the reordered chunks and an index table that
   maps the original chunk index to its offset in the data table.

Trivial example (chunk size: 4, from step 2):
```python
[1, 2, 3, 4, 2, 3, 4, 5, 0, 1, 2, 3]          # source data
-> [[1, 2, 3, 4], [2, 3, 4, 5], [0, 1, 2, 3]] # make chunks
-> [[0, 1, 2, 3], [1, 2, 3, 4], [2, 3, 4, 5]] # rearrange
-> ([0, 1, 2, 3, 4, 5], [1, 2, 0]) # overlap chunks & compute their offsets
```

The compression algorithm is then applied a second time to the previous index table.

The complete algorithm optimizes the two chunk sizes in order to get the lowest total data size.

The mappings were generated using CPython 3.12, PyICU 2.12, PyYaml 6.0.1and icu 73.2.

**TODO:**
- [x] Write independent doc in the code
- [ ] Simplify Python code


wismill:
Note that the Python script is probably a bit too complex, because I experimented various techniques before choosing the best one.

wismill:
CI fixed

wismill:
I would rather have the new release out soon, so let’s postpone it for the following release.

wismill:
Rebased & documented.

The Python script could be simplified; in the current state it is probably too general. But I am running out of time for a script that is probably not going to be run anytime soon.

wismill:
Rebased

wismill:
Rebased, fixed title-case handling, improved tests and added log entry.

PR #434, opened 2024-01-16
compose: Fix iterator for empty tables

wismill:
The current `xkb_compose_table_iterator_next` segfaults when used with an empty table. Indeed, in this case we initialize cursors in `xkb_compose_table_iterator_new` with the dummy node and the direction `NODE_LEFT`, but the dummy node is a leaf!

Fixed by initializing with no cursors when the table is has no non-dummy nodes.


wismill:
Sorry, I do not understand how I could forget to check this…

whot:
belated :+1:  :)

PR #439, opened 2024-02-05
doc: remove obsolete HTML_TIMESTAMP

whot:
warning: Tag 'HTML_TIMESTAMP' at line 44 of file '/home/whot/xorg/lib/libxkbcommon/build/Doxyfile' has become obsolete.


wismill:
`HTML_TIMESTAMP` was deprecated in [Doxygen 1.9.8](https://www.doxygen.nl/manual/changelog.html#log_1_9_8). Should we ensure CI is up to date before this ?

whot:
We don't need to care in this case, the default was `NO`, it's been replaced by just `TIMESTAMP` (also default no), so the behaviour shouldn't change regardless of the version.

wismill (approved):

PR #443, opened 2024-02-07
interactive-evdev: reduce the space allocated for keysyms

whot:
In commit 8cca3a7bfb185876994d6ec3d25cda88e5640a4d the buffer for the keysym was extended to accommodate for up to XKB_COMPOSE_MAX_STRING_SIZE bytes. This caused the printf to expand to the same width for the keysym alone, making the output less useful. Drop this back down to the same width it was before.

cc @wismill


wismill (changes requested):
Good catch!

PR #445, opened 2024-02-08
Prepare for 1.7.0

wismill:
I think we are ready for a new release. There are a few MRs ( #391, #415, #434, #440) that would make a release worth on their own.

Updated NEWS.

TODO:
- [ ] Merge remaining MR planned for 1.7.0

Few questions:
- Is the changelog too detailed?
- Should the keysyms update goes to “breaking changes”? I consider it more as a fix, but it does break anything that rely on the reverted canonical names.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,118 @@
+libxkbcommon 1.7.0 - XXX
+==================
+
+TODO: date
+
+API
+---
+
+### New
+
+- Added early detection of invalid encodings and BOM for keymaps, rules & Compose.
+  Also added a hint that the expected encoding must be UTF-8 compatible.
+
+### Fixes
+
+- Updated keysyms using latest [xorgproto] (commit: `cd33097fc779f280925c6d6bbfbd5150f93ca5bc`):
+
+  For the sake of compatibility, this reintroduces some deleted keysyms and
+  postpones the effective deprecation of others, that landed in xkbcommon 1.6.0.
+
+  - Additions (reverted removal):
+
+    - `XKB_KEY_dead_lowline`
+    - `XKB_KEY_dead_aboveverticalline`
+    - `XKB_KEY_dead_belowverticalline`
+    - `XKB_KEY_dead_longsolidusoverlay`
+
+  - The following keysyms names remain deprecated, but are set again (i.e. as
+    before xkbcommon 1.6.0) as the reference names for their respective keysyms,
+    in order to ensure the transition to the newer names that replace them. This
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,3 +1,95 @@
...
+
+### New
+
+- `xkbcli compile-compose`: added new CLI utility to test Compose files.
+- `xkbcli interactive-x11`: added support for Compose
+- `xkbcli interactive-wayland`: added support for Compose
+
+### Fixes
+
+- `xkbcli interactive-evdev`: fixed help.
```
whot, line 87: fwiw, tiny fixes like this don't IMO need to be mentioned in the release notes since you'd mostly expect every release to have fixes like this all over the place. No need to remove though, this is just a fwiw.

whot (approved):
lgtm, thanks, but I leave the final ack to @bluetech

NEWS
wismill: Agreed. Git history has all the details.

wismill:
@whot suggested that we should downgrade parse error on `Mod6-Mod32` to mere warning ASAP, in preparation of future addition to real modifiers. See #447  for a PoC and discussion.

whot:
fwiw, given how #447 is going to be a bit more complex I reckon it might be best punting that to the next release.

bluetech:
> Is the changelog too detailed?

I think it's good!

> Should the keysyms update goes to “breaking changes”? I consider it more as a fix, but it does break anything that rely on the reverted canonical names.

I think the current changelog makes things pretty clear.

wismill:
@whot @bluetech only #465 remains for this release.

wismill:
Note: Release notes should be updated if #465 is merged.

I am testing [towncrier](https://pypi.org/project/towncrier/) as [proposed by Peter](https://github.com/xkbcommon/libxkbcommon/pull/445#issuecomment-1933593824). Sounds a good tool for the job, will send MR soon for the next release.

wismill:
I added entry for #465. Please check my wording: my intention is to warn of the inconsistency but at the same time to reassure it was an issue with relatively only few keysyms.

wismill:
Added entry for #468 (pending).

wismill:
@bluetech @whot I think we are ready for a new release. The `NEWS.md` file requires a proper date.

whot:
LGTM!

bluetech (approved):
Thanks! Added an entry for d455e805f9ee21059dbbff01502f0f41a242cf3f, will do the release now.

Issue #432, opened 2024-01-13
Add support for conditional comments


wismill:
(EDIT: This need reformulation. See my [comment](https://github.com/xkbcommon/libxkbcommon/issues/432#issuecomment-1891023169) below.
# TLDR;

I propose to add support for __conditional comments__ in the keymap text format, in order to introduce _non_-backward compatible features, in files parsable by _both_ legacy XOrg tools and xkbcommon.

In a gist (_syntax subject to change_):

```c
xkb_symbols "example-1" {
    // Comments starting with “@” are treated as special comments.
    // They are ignored by XOrg tools and previous versions of xkbcommon.
    
    // The following conditional comment starts a conditional block.
    // It guards XKB code for xkbcommon with support for multiple keysyms.
    
    //@if support-multiple-keysyms
    
    // Next comment is commented normal XKB code, parsed and interpreted only by 
    // xkbcommon versions that support both conditional comments and multiple keysyms output.
    // This allow us to support both legacy and modern XKB next to each other.
    
    //@  key <AD01> { [ {a, b}, Q ] };
    
    // Next comment is a conditional comment to guard the fallback code.
    
    //@else
    
    // The following line is normal code, always parsed by XOrg tools and old xkbcommon versions,
    // but parsed by newer version of xkbcommon only without support for multiple keysyms.
         key <AD01> { [ q, Q ] };
        
    // Next comment ends the conditional block.
        
    //@endif
}
```

# Context

I started to look at pretty old issues such as:
- [Need to kick hotkeys on release, not press](https://gitlab.freedesktop.org/xorg/xserver/-/issues/258)
- [Option to apparently unlock the Lock modifier on press of Caps Lock](https://gitlab.freedesktop.org/xorg/xserver/-/issues/312)

I am not going to discuss them in details here. What they have in common is that they need to act _against_ the XKB protocol. Or rather, they need to _extend_ the protocol. Also: __they are _not_ going to be fixed in XOrg[^1]__.

[^1]: Well, they could, in the _unlikely_ case somebody stands up.

XKB devs were pretty clear that going against the XKB specification is risky in general. Hard-coding ad-hoc solutions for the issues mentioned before does not seem a good idea either.

Instead, I think we should _extend_ the protocol and the keymap text syntax. Again, I am not going to share a solution here, but rather raise a question in the following section.

Note that we _already_ introduced a backward-incompatible syntax for multiple keysyms:

```c
xkb_symbols "example-2" {
[...]

PR #446, opened 2024-02-08
Experimental: Add XKB extension for locks

wismill:
Following the discussion in #432, I have done a little experiment that implements two new options:

- `lockOnRelease` for `LockGroup` action: intended to fix [Need to kick hotkeys on release, not press](https://gitlab.freedesktop.org/xorg/xserver/-/issues/258).
- `unlockOnPress` for `LockMods` action: intended to fix [Option to apparently unlock the Lock modifier on press of Caps Lock](https://gitlab.freedesktop.org/xorg/xserver/-/issues/312).

This is backward-incompatible as it introduces a new syntax for the keymap.

I see at least 4 points to discuss here:
1. Is the behavior of the new options correct?
2. How do we make this change more compatible?
    I think there are 2 aspects to consider:
    1. Compatibility with older versions of xkbcommon.  We could propose an _additional_ syntax using the `Private` action, useful for exchange/ouput. Older xkbcommon would just ignore this action and thus we avoid a syntax error.
    2. Compatibility with X11. This feature could be implemented as well in the Xorg universe. Anyway, we should probably use `Private` as presented in the previous section, in order to maximize compatibility.
3. Should these options extended to their respective actions? I would say yes for symmetry, but we could also simply wait to see it there is a need.
4. Does it need a new version of the corresponding Wayland protocol? While I would like to have version negociation, I also hope that the spec is lax enough to avoid to have to wait for this.

I think that if we answer 1) and 2), then we could publish it as a tech preview, with a big warning. That could get us feedback before engaging with protocol updates.

@whot @bluetech @fooishbar

### DIFF
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -152,6 +152,7 @@ enum xkb_action_flags {
     ACTION_ACCEL = (1 << 8),
     ACTION_SAME_SCREEN = (1 << 9),
     ACTION_LOCK_ON_RELEASE = (1 << 10),
+    ACTION_UNLOCK_ON_PRESS = (1 << 11),
 };
 
 enum xkb_action_controls {
--- a/src/state.c
+++ b/src/state.c
@@ -395,9 +395,17 @@ xkb_filter_mod_lock_new(struct xkb_state *state, struct xkb_filter *filter)
 {
     filter->priv = (state->components.locked_mods &
                     filter->action.mods.mods.mask);
-    state->set_mods |= filter->action.mods.mods.mask;
-    if (!(filter->action.mods.flags & ACTION_LOCK_NO_LOCK))
-        state->components.locked_mods |= filter->action.mods.mods.mask;
+    if (filter->priv && (filter->action.mods.flags & ACTION_UNLOCK_ON_PRESS)) {
+        /* XKB extension: Unlock on second press */
+        state->clear_mods |= filter->action.mods.mods.mask;
+        if (!(filter->action.mods.flags & ACTION_LOCK_NO_UNLOCK))
+            state->components.locked_mods &= ~filter->priv;
+        filter->priv = 0;
+    } else {
+        state->set_mods |= filter->action.mods.mods.mask;
+        if (!(filter->action.mods.flags & ACTION_LOCK_NO_LOCK))
+            state->components.locked_mods |= filter->action.mods.mods.mask;
+    }
 }
 
 static bool
--- a/src/xkbcomp/action.c
+++ b/src/xkbcomp/action.c
@@ -101,6 +101,7 @@ enum action_field {
     ACTION_FIELD_KEYCODE,
     ACTION_FIELD_MODS_TO_CLEAR,
     ACTION_FIELD_LOCK_ON_RELEASE,
+    ACTION_FIELD_UNLOCK_ON_PRESS,
 };
 
 ActionsInfo *
@@ -168,6 +169,7 @@ static const LookupEntry fieldStrings[] = {
     { "clearmods",        ACTION_FIELD_MODS_TO_CLEAR   },
     { "clearmodifiers",   ACTION_FIELD_MODS_TO_CLEAR   },
     { "lockOnRelease",    ACTION_FIELD_LOCK_ON_RELEASE },
+    { "unlockOnPress",    ACTION_FIELD_UNLOCK_ON_PRESS },
     { NULL,               0                            }
 };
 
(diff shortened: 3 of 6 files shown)


whot:
> Is the behavior of the new options correct?

The code LGTM, but tests would answer that question :)

> How do we make this change more compatible?

Adding `XKB_KEYMAP_FORMAT_TEXT_V2` solves the case of `xkb_keymap_as_string()`.

And we have a pretty clear split between compositor-side `xkb_keymap_new_from_names` and client-side `xkb_keymap_new_from_string()`. As long as `xkb_keymap_as_string()` spits out the right version, we shouldn't have much of an issue in actual deployments because we can mostly assume that compositors generate from the keymap files (and thus will handle v2 if they want) and clients receive from the compositor and thus will get v1 (old compositors) or v2 (new compositors + client saying they want it).

I don't think we need to worry about X11, that ship has sailed and is listing.

So what we need is the Wayland negotiation which I at minimum needs a `wl_keyboard` (and `wl_seat/pointer/touch`) version bump, a `wl_keyboard.done` event and then the documentation that if the client has version $whatever the compositor may send multiple keymaps for the client to pick-and-choose:

- `wl_seat.get_keyboard` - unchanged
- `wl_keyboard.keymap(xkb_v1)` - unchanged
- `wl_keyboard.keymap(xkb_v2)` <- new in $version
- `wl_keyboard.done`  <- new in $version

Not sure whether we need a `wl_keyboard.use_keymap(xkb_v2)` request to confirm to the compositor which keymap we ended up using. The problem is that every time we bump the XKB version we need to bump the wayland protocol too.

A potentially better approach is to add a subversion so we use `xkb_v1` as identifier for our general format (`"xkb_keymap {...}"`) and use numeric versions within:
- `wl_seat.get_keyboard` - unchanged
- `wl_keyboard.keymap(xkb_v1)` - unchanged or not sent at all if client >= $version
- `wl_keyboard.keymap_version(xkb_v1, 1)` <- optional, same as `wl_keyboard.keymap(xkb_v1)`
- `wl_keyboard.keymap_version(xkb_v1, 2)` <- new in $version
- `wl_keyboard.done` <- new in $version

This way we can introduce future XKB versions like this without having to bump the wayland protocol for each of them. Note that doing so decouples `XKB_KEYMAP_FORMAT_TEXT_V2` from Wayland's `xkb_v1` which needs to be documented to avoid confusion. 

We could do a subversion here too like `xkb_keymap_as_string_with_version(XKB_KEYMAP_FORMAT_TEXT_V1, 2)` but not sure it provides the same benefits.


> Should these options extended to their respective actions? I would say yes for symmetry, but we could also simply wait to see it there is a need.

Can you rephrase/expand on this, I'm not sure what you mean here.

> Does it need a new version of the corresponding Wayland protocol?

Yes, we can only send keymaps to the client that are compatible v1.

wismill:
> >  Should these options extended to their respective actions? I would say yes for symmetry, but we could also simply wait to see it there is a need.

> Can you rephrase/expand on this, I'm not sure what you mean here.

@whot I meant to add also `lockOnRelease` to `LockMods` and `unlockOnPress` to `LockGroup`. However, I do not think it is sound for the latter. And I am not sure there is a use case for the former… So let’s forget about symmetry here.

bluetech:
Just a quick question, what actually happens on old xkbcommon and xkbcomp if you set these options? Do they count towards the fatal error count, or just issue a log and ignored?

bluetech:
Regarding

> Is the behavior of the new options correct?

Do you think you can document the new flags in keymap-format-text-v1.md? (Mostly so I can remind myself about these things :grinning: but also because it seems better to discuss a "spec" change before a code change).

(Also, link to a previous Xorg proposal https://lists.x.org/archives/xorg-devel/2012-November/034427.html by Andreas Wettstein)

wismill:
> Just a quick question, what actually happens on old xkbcommon and xkbcomp if you set these options? Do they count towards the fatal error count, or just issue a log and ignored?

xkbcommon 1.6 (xkbcli compile-keymap)
```
xkbcommon: ERROR: Unknown field name unlockOnPress
xkbcommon: ERROR: Failed to compile xkb_compatibility
xkbcommon: ERROR: Failed to compile keymap
Couldn't create xkb keymap
```

xkbcomp 1.4.7:
```
Warning:          Unsupported maximum keycode 569, clipping.
                  X11 cannot support keycodes above 255.
Keycodes above 372 (e.g. <I372>) are not supported by X and are ignored
Error:            Unknown field name unlockOnPress
```

So error in both cases.

wismill:
> Do you think you can document the new flags in keymap-format-text-v1.md? (Mostly so I can remind myself about these things 😀 but also because it seems better to discuss a "spec" change before a code change).

@bluetech I tried to adapt the fix of Andreas (thanks for the link) to xkbcommon. I agree we should ideally discuss spec before the code, but since I have limited time and we are not going to merge this fix before handling the compatibility question, I though I would leave the nice doc for now.

- `lockOnRelease` locks group on release rather than press as in the protocol. Intended to fix issue for group lock on e.g. `shift+alt`.
- `unlockOnPress` unlocks modifier on second press rather than the second release, as in the protocol. Intended to fix issue with `CapsLock` for fast typists.

PR #447, opened 2024-02-11
Experiment: Add new real modifiers

wismill:
In X11 we are limited to 8 _real_ modifiers and it seems just fine in general. But they are some use cases where it would be useful to have some more:

- Have true separate `Meta` and `Hyper` modifiers;
- Have an `Fn` layer;
- Possibly an accessibility modifier;
- Use latch modifiers rather than dead keys, because then the layout is completely defined in XKB;
- Differentiate right/left modifiers.

It is already possible to do this, but at the cost of other modifiers.

This PR adds 5 new modifiers (`Mod6`..`Mod10`) for a total of **13** (count subject to change). Why not add more? Good question! We encode the mask on 32 bits, but _real_ and _virtual_ modifiers share the same mask. So we cannot have 32 real modifiers because we need to keep some bits (at least 16) to encode the virtual modifiers as well.

This PR is a PoC to open the discussion. 

*Warning:* The current implementation provides no backward-compatibility (X11, older xkbcommon).

@whot @bluetech

### DIFF
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -31,8 +31,11 @@
 static void
 update_builtin_keymap_fields(struct xkb_keymap *keymap)
 {
-    /* Predefined (AKA real, core, X11) modifiers. The order is important! */
+    /* Predefined (AKA real) modifiers. The order is important!
+     * Indeed, the modifier position defines the bit used to encode it.
+     * We use explicit indexes here to facilitate debugging. */
     static const char *const builtin_mods[] = {
+        /* Core/X11 modifiers */
         [0] = "Shift",
         [1] = "Lock",
         [2] = "Control",
@@ -40,8 +43,16 @@ update_builtin_keymap_fields(struct xkb_keymap *keymap)
         [4] = "Mod2",
         [5] = "Mod3",
         [6] = "Mod4",
-        [7] = "Mod5"
+        [7] = "Mod5",
+        /* xkbcommon extension */
+        [8] = "Mod6",
+        [9] = "Mod7",
+        [10] = "Mod8",
+        [11] = "Mod9",
+        [12] = "Mod10"
     };
+
+    _Static_assert(ARRAY_SIZE(builtin_mods) == MOD_REAL_MAX);
 
     for (unsigned i = 0; i < ARRAY_SIZE(builtin_mods); i++) {
         keymap->mods.mods[i].name = xkb_atom_intern(keymap->ctx,
--- a/test/data/keymaps/real-modifiers.xkb
+++ b/test/data/keymaps/real-modifiers.xkb
@@ -26,8 +26,8 @@ xkb_keycodes "test" {
 
 xkb_types "complete" {
     // Define a buch of virtual modifier to test mappings to Mod5-Mod10 real
-    // modifiers. WonderMod will be mapped to Mod5+Mod6.
-    virtual_modifiers MyMod5,MyMod6,MyMod6,MyMod7,MyMod8,MyMod9,MyMod10,WonderMod;
+    // modifiers. VMod11 will be mapped to Mod5+Mod6.
+    virtual_modifiers VMod5,VMod6,VMod6,VMod7,VMod8,VMod9,VMod10,VMod11;
 
     type "ONE_LEVEL" {
         modifiers= none;
@@ -40,16 +40,16 @@ xkb_types "complete" {
 		level_name[2]= "Shift";
 	};
     type "SOOO_MANY_LEVELS" {
-        modifiers = Shift+MyMod5+MyMod6+MyMod6+MyMod7+MyMod8+MyMod9+MyMod10+WonderMod;
+        modifiers = Shift+VMod5+VMod6+VMod6+VMod7+VMod8+VMod9+VMod10+VMod11;
         map[Shift] = 2;
-        map[MyMod5] = 3;
-        map[MyMod6] = 4;
-        map[MyMod6+Shift] = 5;
-        map[MyMod7] = 6;
-        map[MyMod8] = 7;
-        map[MyMod9] = 8;
-        map[MyMod10] = 9;
-        map[WonderMod] = 10;
+        map[VMod5] = 3;
+        map[VMod6] = 4;
+        map[VMod6+Shift] = 5;
+        map[VMod7] = 6;
+        map[VMod8] = 7;
+        map[VMod9] = 8;
+        map[VMod10] = 9;
+        map[VMod11] = 10;
         level_name[1] = "1";
         level_name[2] = "2";
         level_name[3] = "3";
@@ -67,7 +67,7 @@ xkb_compatibility "complete" {
     interpret.repeat= False;
     interpret.locking= False;
 
-    virtual_modifiers MyMod5,MyMod6,MyMod7,MyMod8,MyMod9,MyMod10,WonderMod;
+    virtual_modifiers VMod5,VMod6,VMod7,VMod8,VMod9,VMod10,VMod11;
 
     interpret Any+AnyOf(all) {
         action = SetMods(modifiers=modMapMods,clearLocks);
@@ -76,43 +76,43 @@ xkb_compatibility "complete" {
     // Define some letter keysyms as custom modifiers
     interpret Q {
         useModMapMods = level1;
-        virtualModifier = MyMod5;
-        action = SetMods(modifiers=MyMod5);
+        virtualModifier = VMod5;
+        action = SetMods(modifiers=VMod5);
     };
     interpret W {
         useModMapMods = level1;
-        virtualModifier = MyMod6;
-        action = SetMods(modifiers=MyMod6);
+        virtualModifier = VMod6;
+        action = SetMods(modifiers=VMod6);
     };
     interpret E {
         useModMapMods = level1;
-        virtualModifier = MyMod7;
-        action = SetMods(modifiers=MyMod7);
+        virtualModifier = VMod7;
+        action = SetMods(modifiers=VMod7);
     };
     interpret R {
         useModMapMods = level1;
-        virtualModifier = MyMod8;
-        action = SetMods(modifiers=MyMod8);
+        virtualModifier = VMod8;
+        action = SetMods(modifiers=VMod8);
     };
     interpret T {
         useModMapMods = level1;
-        virtualModifier = MyMod9;
-        action = SetMods(modifiers=MyMod9);
+        virtualModifier = VMod9;
+        action = SetMods(modifiers=VMod9);
     };
     interpret Y {
         useModMapMods = level1;
-        virtualModifier = MyMod10;
-        action = SetMods(modifiers=MyMod10);
+        virtualModifier = VMod10;
+        action = SetMods(modifiers=VMod10);
     };
     interpret U {
         useModMapMods = level1;
-        virtualModifier = WonderMod;
-        action = SetMods(modifiers=WonderMod);
+        virtualModifier = VMod11;
+        action = SetMods(modifiers=VMod11);
     };
     interpret I {
         useModMapMods = level1;
-        virtualModifier = WonderMod;
-        action = SetMods(modifiers=WonderMod);
+        virtualModifier = VMod11;
+        action = SetMods(modifiers=VMod11);
     };
 };
 xkb_symbols {
@@ -124,31 +124,31 @@ xkb_symbols {
     key <LFSH> { [Shift_L] };
     modifier_map Shift { <LFSH> };
 
-    // virtual modifier: MyMod5
+    // virtual modifier: VMod5
     key <AD01> { [Q] };
     modifier_map Mod5 { <AD01> };
 
-    // virtual modifier: MyMod6
+    // virtual modifier: VMod6
     key <AD02> { [W] };
     modifier_map Mod6 { <AD02> }; // real modifier not supported by X11
 
-    // virtual modifier: MyMod7
+    // virtual modifier: VMod7
     key <AD03> { [E] };
     modifier_map Mod7 { <AD03> }; // real modifier not supported by X11
 
-    // virtual modifier: MyMod8
+    // virtual modifier: VMod8
     key <AD04> { [R] };
     modifier_map Mod8 { <AD04> }; // real modifier not supported by X11
 
-    // virtual modifier: MyMod9
+    // virtual modifier: VMod9
     key <AD05> { [T] };
     modifier_map Mod9 { <AD05> }; // real modifier not supported by X11
 
-    // virtual modifier: MyMod10
+    // virtual modifier: VMod10
     key <AD06> { [Y] };
     modifier_map Mod10 { <AD06> }; // real modifier not supported by X11
 
-    // virtual modifier: WonderMod (Mod5+Mod6)
+    // virtual modifier: VMod11 (Mod5+Mod6)
     key <AD07> { [U] };
     modifier_map Mod5 { <AD07> };
     key <AD08> { [I] };
--- a/test/data/keymaps/real-modifiers.xkb
+++ b/test/data/keymaps/real-modifiers.xkb
@@ -1,0 +1,151 @@
+xkb_keymap {
+xkb_keycodes "test" {
+    minimum = 8;
+    maximum = 255;
+	<LVL3>  = 92;
+    <LFSH>  = 50;
+	<RTSH>  = 62;
+	<LALT>  = 64;
+	<RALT>  = 108;
+	<LWIN>  = 133;
+	<RWIN>  = 134;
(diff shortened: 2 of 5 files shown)


test/data/keymaps/real-modifiers.xkb
```diff
@@ -0,0 +1,151 @@
...
+        useModMapMods = level1;
+        virtualModifier = WonderMod;
+        action = SetMods(modifiers=WonderMod);
+    };
+};
+xkb_symbols {
+    name[group1]="Test";
+
+    key <LFSH> { [Shift_L] };
+    modmap Shift { <LFSH> };
```
whot, line 119: can we use `modifier_map` here please? A git grep on xkeyboard-config shows `modmap` isn't used anywhere in the official layouts and while we should have a test that it still works as `modmap` (or `modMap`) I think it's better to stick to the common ones for tests that look at other things. Otherwise it's too confusing for occasional reviewer because it looks like yet another instruction.

also: inconsistent spelling, the ones below use `modMap`.

test/data/keymaps/real-modifiers.xkb
```diff
@@ -0,0 +1,151 @@
...
+        modifiers= none;
+        level_name[Level1]= "Any";
+    };
+	type "TWO_LEVEL" {
+		modifiers= Shift;
+		map[Shift]= 2;
+		level_name[1]= "Base";
+		level_name[2]= "Shift";
+	};
+    type "SO_MANY_LEVELS" {
```
whot, line 40:
```suggestion
    type "SOOO_MANY_LEVELS" {
```
:wink:

test/modifiers.c
```diff
@@ -151,10 +157,67 @@ test_modmap_none(void)
...
+                        KEY_O, BOTH, XKB_KEY_i, NEXT,
+                        KEY_Y, UP,   XKB_KEY_y, NEXT,
+                        KEY_U, DOWN, XKB_KEY_u, NEXT,
+                        KEY_O, BOTH, XKB_KEY_j, NEXT,
+                        KEY_U, UP,   XKB_KEY_u, NEXT,
+                        KEY_Q, DOWN, XKB_KEY_q, NEXT,
+                        KEY_W, DOWN, XKB_KEY_w, NEXT,
+                        KEY_O, BOTH, XKB_KEY_j, NEXT,
+                        KEY_W, UP,   XKB_KEY_w, NEXT,
+                        KEY_Q, UP,   XKB_KEY_q, FINISH));
```
whot, line 210: staring at `test_key_seq` - it doesn't actually test any of the modifiers being set here. which arguably here is quite important ;)

test/modifiers.c
```diff
@@ -151,10 +157,67 @@ test_modmap_none(void)
...
+    xkb_keycode_t keycode;
+
+    keymap = test_compile_file(context, "keymaps/real-modifiers.xkb");
+    assert(keymap);
+
+    keycode = xkb_keymap_key_by_name(keymap, "AD01");
+    assert(keycode != XKB_KEYCODE_INVALID);
+    key = XkbKey(keymap, keycode);
+    assert(key->modmap == Mod5Mask);
+    assert(key->vmodmap == Mod10Mask << 1);
```
whot, line 175: to make sure I understand this right: `Mod10Mask` is the highest real one we have, and this key is mapped to the first vmod defined and thus must have `highest-real-mask << 1` set? Might be worth adding a comment to that effect, took me too long to work that out :)

src/keymap-priv.c
```diff
@@ -31,18 +31,29 @@
...
         [1] = "Lock",
         [2] = "Control",
         [3] = "Mod1",
         [4] = "Mod2",
         [5] = "Mod3",
         [6] = "Mod4",
-        [7] = "Mod5"
+        [7] = "Mod5",
+        /* xkbcommon extension */
+        [8] = "Mod6",
```
whot, line 48: Is it worth naming those `HMod1` through to `HMod5`? For "high modifier", or whatever other name we pick, just to make sure these mods are different to `Mod1`-`5`?

test/data/keymaps/real-modifiers.xkb
```diff
@@ -0,0 +1,151 @@
...
+        level_name[9] = "9";
+        level_name[10] = "10";
+    };
+};
+xkb_compatibility "complete" {
+    interpret.useModMapMods= AnyLevel;
+    interpret.repeat= False;
+    interpret.locking= False;
+
+    virtual_modifiers MyMod5,MyMod6,MyMod7,MyMod8,MyMod9,MyMod10,WonderMod;
```
whot, line 68: does `WonderMod` have some specific meaning that is not yet tested for in this draft?

test/modifiers.c
```diff
@@ -151,10 +157,67 @@ test_modmap_none(void)
...
+    assert(keycode != XKB_KEYCODE_INVALID);
+    key = XkbKey(keymap, keycode);
+    assert(key->modmap == Mod6Mask);
+    assert(key->vmodmap == Mod10Mask << 2);
+
+    assert(test_key_seq(keymap,
+                        KEY_O, BOTH, XKB_KEY_a, NEXT,
+                        KEY_Q, DOWN, XKB_KEY_q, NEXT,
+                        KEY_O, BOTH, XKB_KEY_c, NEXT,
+                        KEY_Q, UP,   XKB_KEY_q, NEXT,
```
whot, line 187: Is there any specific reason for this whole test sequence, with some keys down/up and others as `BOTH`? It looks special enough but I can't see it. Other than the up/down keys are our new modifiers but we don't actually check for modifier state anywhere.

Also, I must be missing something but: why do `a`, `c`, ... have a keysym assigned? our keymap doesn't use `include` so shouldn't all those be `NoSymbol`?

whot (commented):
Very nice, thanks! I stared at this over the weekend too and came up with the same approach (except naming them `HMod1...` as suggested).

I *think* we can add 8 modifiers immediately, XKB only allows for 16 virtual modifiers so we do have 8 bits available. But only adding 5 is the better approach, it kicks the can down the road for a bit and leaves us 3 bits for some potential emergency hack.

whot:
So in terms of backwards-incompatibility:
- dumping the keymap will include the new modifiers
- the keys' modmap states will include bitmasks earlier clients don't know about?

Both of these could be handled via xkb context creation flags, I think. Anything else?


Also cc @fooishbar in case he remembers anything that would prevent this approach.

test/data/keymaps/real-modifiers.xkb
wismill: aha I was about to use emojis in the name but better to keep test ascii.

test/modifiers.c
wismill: Well we do: the various `KEY_Q..KEY_I` are mapped to virtual modifiers through the keysyms `q..i` (now `Q..I`). The interesting result is what we get from `KEY_O` (`<AD09>`), which uses our special key type `SOOO_MANY_LEVELS`.

test/modifiers.c
wismill: Good point! Fixed

src/keymap-priv.c
wismill: Continuing the sequence seems more natural to me.

If we were to start a new sequence, then I suggest we use a longer prefix, e.g. `HigherMod1`, in order to make clear it is not a typo. I am not sure “higher” is a good description though. Maybe “extended” is better, such as `ExtMod1` or `ModExt1`?

test/data/keymaps/real-modifiers.xkb
wismill: Its secret and special characteristic is to be mapped to **two** real modifiers `Mod5+Mod6`. I could name it `MyMod5Plus6` if you prefer. Joke apart, I added comment to make this clear.

wismill:
I improve the MR with comments.

> Both of these could be handled via xkb context creation flags, I think. Anything else?

@whot maybe a new keymap format for `xkb_keymap_get_as_string` is enough? E.g. `XKB_KEYMAP_FORMAT_TEXT_V1_X11` or `XKB_KEYMAP_FORMAT_TEXT_V1_COMPAT`?

whot:
Random-ish comment because right now this is a thought bubble because I haven't really wrapped my head around this: is it possible to make virtual modifiers *actual* virtual modifiers (as opposed to just name aliases for real modifiers?). They're not visible in the API at the moment but the information itself seems to be largely there?

e.g. remove the line `modifier_map Mod2 { <NMLK> };` from your keymap and then run:
```
$ sudo ./build/interactive-evdev --keymap $keymapfile --print-modmaps
...
keycode [ NMLK ] modmap [         ] vmodmap [ NumLock  ] keysyms [ Num_Lock                    ] unicode [   ] layout [ English (US) (0) ] level [ 0 ] mods [ ] leds [ ]
```

The public api (`mods [ ]`) doesn't know about it but the internal API knows there is a virtual modifier even though it's not bound to anything. If we can add that modifier in the public API we may be able to get away with all this without backwards incompatible keymaps.

The API/behaviour may be backwards incompatible but I'm not sure that matters too much.

whot:
> maybe a new keymap format for xkb_keymap_get_as_string is enough? E.g. XKB_KEYMAP_FORMAT_TEXT_V1_X11 or XKB_KEYMAP_FORMAT_TEXT_V1_COMPAT?

So this will be for the exchange format[^1] in Wayland which would be the next step. But right now Wayland has no way of announcing which format the client supports (and even with that added we'll have clients supporting the v1 format only). So we'll still need some way to tell libxkbcommon to export v1 format (i.e. without the incompatible new `Mod6`), and a newer libxkbcommon to behave like the old libxkbcommon (i.e. it can parse but never set the bits for `Mod6` as seen by the public API). Not sure how much the latter is required, it feels like a bug if an application is confused by that.

[^1]: and IMO wee should just go with `_V2` and document the permitted differences instead of trying to add semantic names. It's hard to run out of numbers after all and names will make it more confusing.

test/data/keymaps/real-modifiers.xkb
whot: nah, the name is fine, makes it stick out. I just didn't see the mapping. Thanks for the clarification and comments.

src/keymap-priv.c
whot: yeah, naming things is hard. Either way, I'm not going to fight hard for this so unless we come up with some outstanding name, `Mod6` and friends it is.

test/data/keymaps/real-modifiers.xkb
```diff
@@ -0,0 +1,164 @@
...
+	<AD06> = 29;
+	<AD07> = 30;
+	<AD08> = 31;
+	<AD09> = 32;
+};
+
+xkb_types "complete" {
+    // Define a buch of virtual modifier to test mappings to Mod5-Mod10 real
+    // modifiers. WonderMod will be mapped to Mod5+Mod6.
+    virtual_modifiers MyMod5,MyMod6,MyMod6,MyMod7,MyMod8,MyMod9,MyMod10,WonderMod;
```
whot, line 30: Another nitpick because I'm getting confused reading this: how about `VMod5` etc. so it's instantly readable as "this is a virtual modifier?

test/modifiers.c
whot: *audible click sound in the background* 

ah, now i get it, thanks!

whot (commented):
Thanks, this is a lot easier to understand now with the comments.

fooishbar:
Yeah, we thought a lot about it at the time and decided not to. The main reason was that it was far too early to break compatibility with X11. If this were to be done, you would either have to make the new modifiers purely virtual, or if they were real modifiers, to just no-op anything where a real modifier was active or set to be invoked for X11.

I don't think it's feasible to split real modifiers, e.g. turn the single Meta modifier (implying at least one of LAlt/RAlt virtual modifiers) into two separate modifiers. That would play havoc with all the existing keymaps, and also very likely clients who depend on, e.g. Meta active with A pressed to be interpreted as Alt+A, since they would instead see A pressed with some weird new left-alt modifier active. Similarly adding new real modifiers won't work; A pressed with Meta+LAlt real modifiers would be seen as different by clients to Meta+A when processing actions.

Adding new virtual modifiers is a great idea though. We did explicitly make room in the API for doing this, and it would be great to have. The API is probably incomplete and could use some description (implying more changes in keymap files), but the idea is sound.

> So this will be for the exchange format[1](https://github.com/xkbcommon/libxkbcommon/pull/447#user-content-fn-1-4b4eb1f8fb96fe69145c8a3e17283092) in Wayland which would be the next step. But right now Wayland has no way of announcing which format the client supports (and even with that added we'll have clients supporting the v1 format only).

... yeah, you've picked this problem out. I got this badly wrong in the original wl_keyboard design. It paints clients into the same corner as X11. We'd need to fix that, and I'll gladly help with review etc.

wismill:
@fooishbar IIRC the modifier mechanics, the mapping of virtual modifiers to real modifiers is quite essential. In this case, having virtual modifiers without such mapping will not work at all without a serious change in the modifier handling.

> I don't think it's feasible to split real modifiers, e.g. turn the single Meta modifier (implying at least one of LAlt/RAlt virtual modifiers) into two separate modifiers. That would play havoc with all the existing keymaps, and also very likely clients who depend on, e.g. Meta active with A pressed to be interpreted as Alt+A, since they would instead see A pressed with some weird new left-alt modifier active. Similarly adding new real modifiers won't work; A pressed with Meta+LAlt real modifiers would be seen as different by clients to Meta+A when processing actions.

I am not sure I follow you. I remember having played with separate Alt/Meta and Super/Hyper without issue, but unfortunately I do not find the file. Also I would like to clarify that I do not propose to do that by default nor do I even propose to add it the option to `xkeyboard-config`. Anyway these are only some examples of use cases. There is another use case that I just added: some people prefer to use latch modifiers rather than dead keys, because then the layout is completely defined in XKB. If real modifiers mapping is required, then one can lack modifiers quickly!

---

I am all in favor to allow virtual modifiers to work without the need for a mapping to real modifiers. In this case, as you wrote, we do not even need any new real modifiers. If we go that route:
- We should maybe change the vocabulary we use: _real_/_virtual_ modifiers is a confusing terminology. Real modifier would be mere _compatibility_ modifiers – aka X11 core modifiers – but I do not have a better name for the virtual modifiers.
- We should add new warnings in order to forbid missing real modifiers mappings in `xkeyboard-config`.
- Ensure we have enough tests of the current API.

test/data/keymaps/real-modifiers.xkb
wismill: Let’s do this.

wismill:
For reference, what I understood from the modifiers machinery in the section [“Define and use a modifier”](https://xkbcommon.org/doc/current/keymap-text-format-v1.html#autotoc_md47) of the doc.

whot:
>  IIRC the modifier mechanics, the mapping of virtual modifiers to real modifiers is quite essential. In this case, having virtual modifiers without such mapping will not work at all without a serious change in the modifier handling.

Yes, the mapping of modifiers is essential atm. It's a bit confusing because `virtual` modifiers are really just name aliases, they're not actually modifiers (in the spec, anyway).

If we were to add true virtual modifiers the main breakage we would get is from keymaps that define interpret actions but rely on the virtual modifier being mapped or not to trigger those actions. There are only a handful of virtual modifiers currently that aren't in the default `modifier_map` statements:
- OLPC's `Square,Cross,Triangle,Circle`, there is no section that uses those, so I'm not sure how those would ever be mapped (with our existing keymaps). I think we can ignore those
- `Kana_Lock` used in `symbols/sun_vendor/jp` where it's mapped to `Mod3`. Only exposed via option `japan:kana_lock`.
- `Compose` in `compat/ledcompose` where it's used for an indicator, exposed via option ` mod_led:compose`
- `ScrollLock` in `compat/misc` which is dragged into the default keymaps so we all have it but no-one has it *by default*. There's no option for it, only `br(abnt2)` maps it but that's a bug ([xkeyboard-config!662](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/662)).

So the immediate fallout of implementing purely virtual modifiers will likely be seen in the `Scroll_Lock` modifier which everyone has and will suddenly show up as modifier. At least that provides a somewhat widespread test bed for what will blow up :)

> We should add new warnings in order to forbid missing real modifiers mappings in xkeyboard-config.

I'm not sure I understand this point, can you expand on it?

wismill:
> >  We should add new warnings in order to forbid missing real modifiers mappings in xkeyboard-config.

> I'm not sure I understand this point, can you expand on it?

I mean that we should ensure that we do _not_ rely on the new behavior in files used by XOrg, i.e. remove or forget real modifiers mappings that will remain mandatory for X11 systems.

wismill:
Closing in favor of #450. This is a much simpler approach that require no syntax change, although the discussion here  about backward compatibility will be very useful.

PR #449, opened 2024-02-14
Some improvements to test debugging

whot:
`test_key_seq` takes a list of key operations and if one fails it can be a bit tricky to figure out why (e.g. in my case I knew it was RALT but didn't realise it was the up, not the down that triggered the error).

Let's help with this by printing the operation and the keymap state just like e.g. `interactive-evdev` does.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -434,6 +434,7 @@ if build_tools
         link_with: libxkbcommon_tools_internal,
         dependencies: dep_libxkbcommon,
     )
+    configh_data.set10('HAVE_TOOLS', true)
 
     # Tool: xkbcli
     executable('xkbcli', 'tools/xkbcli.c',
@@ -586,6 +587,8 @@ You can disable the Wayland xkbcli programs with -Denable-wayland=false.''')
                dependencies: [tools_dep],
                include_directories: [include_directories('src', 'include', 'tools')],
                install: false)
+else
+    tools_dep = declare_dependency()
 endif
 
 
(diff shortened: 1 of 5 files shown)


wismill:
@whot how do you make `meson test` capture `stdout`? I do not see the output of `tools_print_keycode_state` in the testlog.

whot:
meson captures it separately (similar to e.g. pytest) and prints them separately into the testlog. If you find the `keyseq` test you should see the stdout bits first and then the stderr bits after. Lines 2287 and 2575 in my testlog.txt, should be approximately the same line numbers for you.

(not super-useful for the meson runs but good enough for local test runs which is where you'd need this the most anyway)

wismill:
There is something fishy: I get __*no*__ `stdout` when there is an error, which admittedly is the opposite of what we want. Does it have to do with buffering? Is it a meson bug?

wismill:
@whot It is about buffering: using `setbuf(stdout, NULL);` at the beginning of `main` allows to capture `stdout` correctly. Maybe we should use in in every test?

Note to myself: using `--no-stdsplit` with `meson test` is useful to get everything in context.

whot:
>  when there is an error

Oh, I didn't realize. I only ran this manually for the actual testing and assumed the testlog just worked. Good find!

> Maybe we should use in in every test?

Yes, definitely. I'm surprised meson doesn't do this for all captured output by default.

whot:
I've now added `setbuf` in a slightly hacky way. Ideally we want some `test_init()` in all tests but we can do this as a follow-up somewhen?

wismill (approved):
LGTM, thanks

PR #453, opened 2024-02-18
xkbcomp: Require newline after !include line in rules files

### LINKED ISSUE #452: ! include line in rules file does not require a newline
I have noticed that rules file parser is pretty lenient about !include files. In particular, it does not require a newline after the include, so the following rules file is parsed without errors:
```
! include %S/inc-dst-simple ! include %S/inc-dst-simple ! layout = symbols
```
I guess this is an oversight, and nobody relies on this?

dottedmag:
Rules file parser allows constructs like
```
!include "foo" !include "bar" !layout = symbols
```
This is most likely an oversight in original code.

Closes #452

### DIFF
--- a/src/xkbcomp/rules.c
+++ b/src/xkbcomp/rules.c
@@ -978,6 +978,14 @@ include_statement:
     switch (tok = gettok(m, s)) {
     case TOK_IDENTIFIER:
         matcher_include(m, s, include_depth, m->val.string);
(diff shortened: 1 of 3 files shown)


whot (approved):
This LGTM thanks, cc @wismill and @bluetech for a second set of eyes

wismill (approved):
LGTM, thank you!

PR #454, opened 2024-02-20
Add test_unit to all tests

wismill:
Currently it only ensure we do not buffer `stdout`.

Follow-up of #449 (see [this comment](https://github.com/xkbcommon/libxkbcommon/pull/449#issuecomment-1951707681)).


whot (approved):
thanks!

Issue #424, opened 2024-01-03
symbol map fails with recent lld (aka broken test for -Wl,--version-script)


bapt:
with lld 1.17, linking with a symbol map which contains symbols which actually do not exists now result in an error, which means the test in meson.build to check if -Wl,--version-script works always fails. because it tries to use the general map file with "int main(void) {}".

Either it should be tested with an empty map file, or another strategy should be defined

Or something like the following patch: https://cgit.freebsd.org/ports/tree/x11/libxkbcommon/files/patch-meson.build should be used.

whot:
To save us from having to reproduce this - which of the symbols doesn't exist? This may just be a typo/copy/paste error?

bapt:
Sorry for the delay, the problem is not with the map file, but with the test in meson in create a binary with no symbols and apply the xkbcommon.map version script on it, which results in failure. I am surprised there is an actual need for this test at all

whot:
Looks like this was added for #24 back when we still used autotools and then taken over to meson. A quick google points me to https://github.com/mesonbuild/meson/issues/3047 so meson doesn't appear to have native version-script support yet.

Your patch looks good on a first glance and appears to work locally on my F39 machine, mind filing a pull request for that?

PR #459, opened 2024-02-27
meson: inform ld to accept undefined symbols for version-scripts

bapt:


whot:
Thanks. I copied the comment from #424 into your commit message so we have a bit more info here.

glechapelain:
I'll verify here but this seems to be causing grieves to pkgsrc using GNU:

https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58272

meson.build
```diff
@@ -145,7 +145,7 @@ configh_data.set('WIN32_LEAN_AND_MEAN', 1)
 # Supports -Wl,--version-script?
 have_version_script = cc.links(
     'int main(){}',
-    args: '-Wl,--version-script=' + meson.current_source_dir()/'xkbcommon.map',
+    args: '-Wl,--undefined-version,--version-script=' + meson.current_source_dir()/'xkbcommon.map',
```
eli-schwartz, line 148: Certainly it is a bug if libxkbcommon depends on linker options that only work with specific bleeding-edge versions of binutils.

I am surprised that this broken PR was ever merged, at least in the state it was in. This is an lld-specific issue, so it should have been gated on `cc.get_linker_id() == 'lld'`.

wismill: Let’s focus on this issue in #481.

PR #458, opened 2024-02-24
Do not accept statements like garbage.level_name in types files

dottedmag:
Fix parser accepting clearly nonsensical type definitions like

type "ONE_LEVEL" {
    garbage.modifiers = None;
    garbage.map[None] = Level1;
    garbage.level_name[Level1] = "Any";
};

and ignoring the garbage part.

Closes #457

PR #460, opened 2024-02-29
Fix global default statements

### LINKED ISSUE #457: Types parser allows type definitions with garbage
The following types definition is accepted by .xkb parser and emitted back without the `garbage` tokens:
```
default xkb_types "garbage" {
    type "ONE_LEVEL" {
	garbage.modifiers = None;
	garbage.map[None] = Level1;
	garbage.level_name[Level1] = "Any";
    };
}
```

wismill:
This is a more general version of #458. It fixes the handling of global default statements (`x.y = z;`) when they are in the wrong scope or when the qualifier (`x.` in previous example)  is invalid.

Thanks to @dottedmag for spotting the issue and proposing the original fix.

Closes #457.


doc/message-registry.md
```diff
@@ -6,7 +6,7 @@ NOTE: This file has been generated automatically by “update-message-registry.p
 -->
 
 This page lists the warnings and errors generated by xkbcommon.
-There are currently 54 entries.
```
whot, line 9: Is this correct? Looks like 428 was replaced with a different error so shouldn't the number be the same?

whot (approved):
This is way too deep in the internals for me but code-wise it LGTM, thanks.

doc/message-registry.md
wismill: I just re-run the script:  it is correct. It seems that it was not updated properly in 00e3058e7b027c3d698b24415fd2ac105cd72246. Probably because of an incomplete rebase.

whot: might be worth splitting that particular update out if you have the time and motivation to do it, otherwise just note it in the commit message please, thanks.

src/xkbcomp/compat.c
```diff
@@ -668,18 +668,18 @@ HandleInterpBody(CompatInfo *info, VarDef *def, SymInterpInfo *si)
     ExprDef *arrayNdx;
 
     for (; def; def = (VarDef *) def->common.next) {
-        if (def->name && def->name->expr.op == EXPR_FIELD_REF) {
+        ok = ExprResolveLhs(info->ctx, def->name, &elem, &field, &arrayNdx);
+        if (!ok)
+            continue;
+        if (elem) {
```
bluetech, line 674: This also puts out the error on array refs with element (`x.y[z] = foo`), but that's a good thing, seems like the `x` was was just completely ignored in this case (though would still fail in `SetInterpField` due to the array index).

bluetech (approved):
LGTM, thanks @wismill and @dottedmag!

doc/message-registry.md
wismill: Good point. Done

src/xkbcomp/compat.c
wismill: Well, I have to confess I missed this! 😳 Thank you for pointing it. I checked again that this does not further have side effects. I think we are good, but I will wait a few days before merging.

PR #461, opened 2024-03-01
tools: remove unused `#incldue <ctype.h>`

bluetech:
We avoid this header due to its locale dependence. This include seems like a leftover in 44029221e8423f1ca93470952542a0517a208d42.


wismill (approved):
Good catch! I guess I wanted to use `iscntrl` but changed my mind. Not sure we need to improve the hack much with standard functions.

PR #465, opened 2024-03-05
keysyms: Fix inconsistent case-insensitive name lookup

### LINKED ISSUE #463: xkb_keysym_from_name returns upper case keysym for case-insensitive match in a few cases
`xkb_keysym_from_name` is documented to return a lowercase keysym if asked to do a case-insensitive match.

This fails (uppercase keysym is returned) for the following names:
```
dead_I
dead_i
dead_O
dead_o
dead_SCHWA
dead_schwa

kana_TSU
kana_tsu
kana_YA
kana_ya
kana_YO
kana_yo
```

wismill:
`xkb_keysym_from_name` has inconsistent behavior when used with the flag `XKB_KEYSYM_CASE_INSENSITIVE`:

```c
xkb_keysym_from_name("a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_a;
xkb_keysym_from_name("A", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_a;
xkb_keysym_from_name("dead_a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_A;
xkb_keysym_from_name("dead_A", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_A;
xkb_keysym_from_name("dead_o", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_o;
xkb_keysym_from_name("dead_O", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_o;
xkb_keysym_from_name("KANA_tsu", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_tsu;
xkb_keysym_from_name("KANA_TSU", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_tsu;
xkb_keysym_from_name("KANA_ya", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_YA;
xkb_keysym_from_name("KANA_YA", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_YA;
xkb_keysym_from_name("XF86Screensaver", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_XF86ScreenSaver;
xkb_keysym_from_name("XF86ScreenSaver", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_XF86ScreenSaver;
```

So currently, if two keysym names differ only by case, then the lower-case *keysym* is returned, not the keysym corresponding to the lower-case keysym *name*. Indeed, `xkb_keysym_from_name` uses `xkb_keysym_is_lower` to test if a keysym is a lower-case keysym.

Let’s look at the example for keysyms `a` and `A`: we get the keysym `a` not because its name is lower case, but because `xkb_keysym_is_lower(XKB_KEY_a)` returns true and `xkb_keysym_is_lower(XKB_KEY_A)` returns false.

So the results are correct according to the doc:

- Katakana is not a bicameral script, so e.g. `kana_tsu` is not the lower case of `kana_TSU`.
- As for the `dead_*` keysyms, they are not cased either because they do not correspond to characters.

But this is also very counter-intuitive: `xkb_keysym_is_lower` is not the right function to use in this case, because one would expect to check only the name, not the corresponding character case:

```c
xkb_keysym_from_name("KANA_YA", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_ya;
xkb_keysym_from_name("XF86ScreenSaver", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_XF86Screensaver;
```

Fixed by making the order of the keysyms names consistent in `src/ks_tables.h`:

1. Sort by the casefolded name: e.g. `kana_ya` < `kana_YO`.
2. If same casefolded name, then sort by cased name, i.e for ASCII: upper before lower: e.g `kana_YA` < `kana_ya`.

Thus we now have e.g. `kana_YA` < `kana_ya` < `kana_YO` < `kana_yo`. The lookup logic has also been simplified.

Added exhaustive test for ambiguous case-insensitive names.

Note: This a **breaking change** because we do not rely on `xkb_keysym_is_lower` anymore, but I consider that the function `xkb_keysym_from_name` is currently broken anyway, so I do not think we should try to propose backward compatibility via e.g. `xkb_keysym_from_name2` or `XKB_KEYSYM_CASE_INSENSITIVE2`.

Closes #463


src/keysym.c
```diff
@@ -270,16 +270,18 @@ xkb_keysym_from_name(const char *name, enum xkb_keysym_flags flags)
...
+    * case-insensitive comparison and returns the "best" case-insensitive match.
     *
-    * The "best" case-insensitive match is the lower-case keysym which we find
-    * with the help of xkb_keysym_is_lower(). The only keysyms that only differ
-    * by letter-case are keysyms that are available as lower-case and
-    * upper-case variant (like KEY_a and KEY_A). So returning the first
-    * lower-case match is enough in this case.
+    * The "best" case-insensitive match is the lower-case keysym name. The only
+    * keysyms names that only differ by letter-case are keysyms that are
+    * available as “small” and “big” variants. For example:
```
dottedmag, line 280: I have found one more example of ambiguous keycode names: `XF86ScreenSaver` and `XF86Screensaver`, not sure which one is the right one to return here (maybe it doesn't matter?)

wismill: Check [this test file](https://github.com/xkbcommon/libxkbcommon/blob/7194d5a19eb79ba2b29081b7fcd38737f1c12fd5/test/keysym.h) to find out. Doc can be improved, comment welcome!

dottedmag: I'd say "most keysym names that only differ...", but that's a minor nitpick.

bluetech:
```c
xkb_keysym_from_name("a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_A;
xkb_keysym_from_name("dead_o", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_O;
xkb_keysym_from_name("dead_a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_a;
xkb_keysym_from_name("KANA_tsu", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_TSU;
```

These look very strange and seems too buggy to have been like this forever. Just so the motivation is clear, are these really the current results?!

wismill:
@bluetech sorry the example is not correct. I am correcting the description and will fix the commit accordingly.

So this is what we have currently:

    xkb_keysym_from_name("a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_a;
    xkb_keysym_from_name("A", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_a;
    xkb_keysym_from_name("dead_a", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_A;
    xkb_keysym_from_name("dead_A", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_A;
    xkb_keysym_from_name("dead_o", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_o;
    xkb_keysym_from_name("dead_O", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_dead_o;
    xkb_keysym_from_name("KANA_tsu", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_tsu;
    xkb_keysym_from_name("KANA_TSU", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_tsu;
    xkb_keysym_from_name("KANA_ya", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_YA;
    xkb_keysym_from_name("KANA_YA", XKB_KEYSYM_CASE_INSENSITIVE) == XKB_KEY_kana_YA;

wismill:
@bluetech you can see it in [my test branch](https://github.com/wismill/libxkbcommon/tree/keysyms/test-case_insensitive_name_lookup), where I just added a modified test: look for "ERROR" in the test log.

wismill:
I added a test in `test_ambiguous_icase_names` to ensure that:
> the behavior for keysyms from bicameral scripts remains the same

scripts/makekeys
```diff
@@ -10,7 +10,16 @@ pattern = re.compile(r"^#define\s+XKB_KEY_(?P<name>\w+)\s+(?P<value>0x[0-9a-fA-F
...
 
-entries_isorted = sorted(entries, key=lambda e: e[0].lower())
+# Sort based on the keysym name:
+#   1. Sort by the casefolded name: e.g. kana_ya < kana_YO.
+#   2. If same casefolded name, then sort by cased name, i.e for
+#      ASCII: upper before lower: e.g kana_YA < kana_ya.
+# E.g. kana_YA < kana_ya < kana_YO < kana_yo
+# WARNING: this sort must not be changed, as some functions e.g.
+# xkb_keysym_from_name rely on upper case variant occuring first.
+entries_isorted = sorted(entries, key=lambda e: (e[0].casefold(), e[0]))
```
bluetech, line 20: Something I noticed (not intended as a comment on this PR specifically). We currently run this script (through `update-keysyms`) using `export LC_CTYPE=C`.

- I wonder if changing to `LC_CTYPE=C.UTF-8` has the same results, if so I think we should do it since `C` is legacy.

- Since we're sorting, I think we also need to set `LC_COLLATE`, otherwise we might end up using a different locale for the ordering? (E.g. if the developer is Turkish, might have the infamous Turkish-i issue). I'm assuming Python respects `LC_COLLATE` but I'm not sure.

src/keysym.c
```diff
@@ -296,26 +298,14 @@ xkb_keysym_from_name(const char *name, enum xkb_keysym_flags flags)
             }
         }
         if (entry) {
-            const struct name_keysym *iter, *last;
-
-            if (icase && xkb_keysym_is_lower(entry->keysym))
```
bluetech, line 301: Can we keep the `icase` check? I know that we already have a fast path for `!icase` above so `icase` is guaranteed here, but it's nice to keep respecting it in the slow-path for possible comparison/debugging purposes.

bluetech:
Thanks. So I would say this is not a *breaking* change; our doc promise is

> If you use the XKB_KEYSYM_CASE_INSENSITIVE flag and two keysym differ only by case, then the lower-case keysym is returned.

and we still hold this promise. We just broaden the promise, making the function more useful and more intuitive for non-lower-caseable keysyms. So :+1: on the idea here.

bluetech (approved):
Excellent work as always, the code LGTM.

src/keysym.c
wismill: I gave it a try, but I am really not convinced it helps: I know later I will wonder _why_ is this test there! I could add a comment, but then we already have `Find the correct keysym for case-insensitive match.` heading the slow-path comment block.

scripts/makekeys
wismill: Since keysym names are ASCII-only I think is does not matter much. I tried using `tr_TR.UTF-8` and I have seen no difference.

I do not mind to set these though, see the corresponding commit.

src/keysym.c
bluetech: OK

bluetech (approved):

scripts/makekeys
bluetech: Thanks!

src/keysym.c
wismill: TO be sure: Do you want to keep the new test (and squash) or should I drop the corresponding commit?

whot (approved):
bit handwavy but this lgtm

src/keysym.c
wismill: The fix is quite important, so I am going to merge _without_ the extra test, but I am open to add it in a further PR.

PR #466, opened 2024-03-07
doc: Use towncrier to handle release notes

wismill:
(from a [proposal of Peter](https://github.com/xkbcommon/libxkbcommon/pull/445#issuecomment-1933593824))

Towncrier is a utility to produce useful, summarized news files.

See:
- https://pypi.org/project/towncrier/
- https://towncrier.readthedocs.io

Custom configuration for xkbcommon:
- New fragments are located in the `changes` directory.
- 3 sections:
  - API: `changes/api`
  - Tools: `changes/tools`
  - Build System: `changes/build`
- 3 news fragments:
  - Breaking changes: `.breaking`
  - New: `.feature`
  - Fixes: `.bugfix`

`NEWS` is renamed to `NEWS.md` because the the tool requires `.md` extension to write in markdown format.


bluetech:
From my experience, it's very helpful to have a README.md file in the `changes/` directory explaining a few things:

- General blurb what the directory is
- What are the available fragment types/categories (like what you wrote in the description)
- Which tense/voice to use, e.g. "Fix a segfault" or "Fixed a segfault" or "A segfault has been fixed" etc.

> NEWS is renamed to NEWS.md because the the tool requires .md extension to write in markdown format.

Do the existing contents render well in markdown?

wismill:
> it's very helpful to have a README.md file in the `changes/`

@bluetech yes, good idea. I added such file.

> Do the existing contents render well in markdown?

I would say yes: https://github.com/wismill/libxkbcommon/blob/towncrier/NEWS.md

changes/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- Breaking changes: `.breaking`
+- New: `.feature`
+- Fixes: `.bugfix`
+
+[towncrier]: https://pypi.org/project/towncrier/
+
+## Adding a fragment
+
+Add a short description of the change in a file `changes/SECTION/ID.FRAGMENT`,
```
bluetech, line 25: Is it possible to have an `.md` suffix so it renders as markdown in github?

wismill: Yes, it is accepted. Good idea!

bluetech (approved):

changes/README.md
```diff
@@ -0,0 +1,62 @@
...
+Add a short description of the change in a file `changes/SECTION/ID.FRAGMENT.md`,
+where:
+
+- `SECTION` and `FRAGMENT` values are described in the [previous section](#introduction).
+- `ID` is the corresponding issue identifier on Github, if relevant. If there is
+  no such issue, then `ID` should start with `+` and some identifier that make
+  the file unique in the directory.
+
+Examples:
+- A _bug fix_ for the _issue #463_ is an _API_ change, so the corresponding file
```
whot, line 34: I would recommend using something like `changes/api/463-fix-blah-add-xkb_foo.bugfix.md` so it's easier to spot without knowing all the issue numbers by heart (or looking into each file). But I think that could also be something we figure out over time.

changes/README.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+Then build the changelog:
+
+```bash
+# Only check the result. Useful after adding a new fragment.
+towncrier build --draft --version 1.7.0
+# Write the changelog
+towncrier build --version 1.7.0
+```
```
whot, line 62: I think there's a step missing for "remove old files" :)

whot (approved):
LGTM

changes/README.md
wismill: Agreed. I’ll check this tomorrow.

changes/README.md
wismill: Done

changes/README.md
wismill: So, the part preceding `.bugfix.md` can be anything, but currently I use it to add a link to the corresponding ticket. Unfortunately, `towncrier` does not handle the file name convention you propose.

So we can either:
- Only use ticket ID, as proposed: `123.bugfix.md`. We get the link to the issue. 
- Prefix with `+`, then optionally set ticket, then some short description:  `+123-description.bugfix.md`. No link to the issue.

We did not have links to issues until now. On one hand the second option is conservative and is more readable. On the other hand, these fragments should be written in the MR corresponding to the change, so it is unlikely we need to review them afterwards.

whot: Honestly, this isn't a huge deal so whatever makes life simpler for us (which seems to be the `123.bugfix.md` approach). Thanks for checking though!

PR #471, opened 2024-03-13
meson: disable doc builds by default

whot:
The only docs we build is the doxygen API and I reckon there's a near-zero amount of users who actually want these built locally instead of just browsing the online doc.

Let's not build this by default, dropping the requirement on doxygen and speeding up the build by quite a fair bit.


whot:
Closing in favour of #472 - for some reason force-pushing to fix the CI failure does nothing in this MR. :shrug:

PR #475, opened 2024-03-23
Small fix: in RMLVO, use `variant_idx` to index variants instead of `layout_idx`

wysiwys:
This pull request modifies a line of code in `xkbcomp/rules.c` , in which the `mapping.layout_idx` was used to index the `rmlvo.variants`. 

Is it correct that instead, `mapping.variant_idx` should be used here?

Thanks


bluetech (approved):
Thanks! I think we never noticed this because the variant index virtually always is the same as the layout index. Now I'm curious how you discovered this bug :)

wysiwys:
No problem! Actually I didn't find the bug from using the API, just saw it while looking at the code, with the goal of understanding RMLVO.

PR #482, opened 2024-06-04
doc (manuals): clean and expand apropos results

concussious:
Previously on FreeBSD/mandoc (probably affecting others), everything except xkbcli was being listed twice and wrapping in apropos because mdoc(7) names should not have spaces or be different than the title.

Further, xkb is already parsed in search results by being in the name, so expand xkb to "X keyboard" so that they match for apropos keyboard. We did this already in xkbutils, which matches setxkbmap.

These don't need to be quoted literals, and - doesn't need to be escaped, but I left them because there may be environments which are not to spec, but I'm happy to ammend this commit if that cleanup seems desirable (works on Linux/groff).

### DIFF
--- a/tools/xkbcli-compile-compose.1
+++ b/tools/xkbcli-compile-compose.1
@@ -1,10 +1,10 @@
-.Dd November 7, 2023
+.Dd June 4, 2024
 .Dt XKBCLI\-COMPILE\-COMPOSE 1
 .Os
 .
 .Sh NAME
-.Nm "xkbcli compile-compose"
-.Nd compile a Compose file
+.Nm "xkbcli\-compile\-compose"
+.Nd compile an X Keyboard Compose file
 .
 .Sh SYNOPSIS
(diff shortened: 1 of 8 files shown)


tools/xkbcli-compile-compose.1
```diff
@@ -1,10 +1,10 @@
-.Dd November 7, 2023
+.Dd June 4, 2024
 .Dt XKBCLI\-COMPILE\-COMPOSE 1
 .Os
 .
 .Sh NAME
-.Nm "xkbcli compile-compose"
-.Nd compile a Compose file
+.Nm "xkbcli\-compile\-compose"
+.Nd compile a X Keyboard Compose file
```
bluetech, line 7:
```suggestion
.Nd compile an X Keyboard Compose file
```

bluetech (approved):
Thanks, sounds alright to me.

PR #483, opened 2024-07-04
state: Add support for group latch action

### LINKED ISSUE #455: LatchGroup action is not supported
I was surprised to discover that the `LatchGroup` action is not implemented: [`filter_action_funcs` is missing `ACTION_TYPE_GROUP_LATCH`](https://github.com/xkbcommon/libxkbcommon/blob/e7570bcb78a48c0e3fb48087991a85af95943e98/src/state.c#L529). I think this is an oversight, because this feature is not widely used.

I discovered this while trying to revive #314.

wismill:
Surprisingly, the latch group action was not yet implemented.

Added tests to ensure we get the tricky bits right.

Fixes #455


wismill:
So, mixing `enum` and `signed int` in bit fields turned out to not be portable on Windows, so I replaced the `enum` field with an explicit `unsigned int`.

wismill:
The commit `DELETE ME: …` is for debugging only and should be removed once the rest is validated.

PR #487, opened 2024-07-12
actions: Add support for multiple actions per level

### LINKED ISSUE #486: support multiple actions for a key per level
Good idea! I read your comment on `xkeyboard-config`: we should check multiple actions per level too. Probably another MR.

_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/pull/421#pullrequestreview-1788284256_
            
              Actually multiple actions per level is not currently supported by the parser: not in symbols nor in compatibility files.

_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/issues/421#issuecomment-1870240328_

Still, could we implement one?

Original issue:
https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/428

wismill:
This makes **1 keysym == 1 action**, which seems both useful and more intuitive when using multiple keysyms per level.

The motivation of this new feature are:
- Make multiple keysyms per level more intuitive.
- Explore how to fix the issue with shortcuts in multi-layout settings (see the [`xkeyboard-config` issue](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/416)).

  The idea is to use e.g.:

  ```c
  key <LCTL> {
        symbols[1] = [ {Control_L,                  ISO_First_Group    } ],
        actions[1] = [ {SetMods(modifiers=Control), SetGroup(group=-4) } ]
  };
  ```

  in order to switch temporarily to a reference layout in order to get the same shortcuts on every layout.

  See the `xkeyboard-config` MR “[Add option shortcuts:qwerty](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/717#note_2484076)”.

This is quite a big change, because it breaks the design “1 action per level”.

This is WIP because there might be lots of corner cases.

I will try to break this into smaller commits, although the nature of the change makes it difficult to have smaller commits that compiles.

Fixes #486

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -863,7 +863,13 @@ xkb_context_set_log_fn(struct xkb_context *context,
 /** Flags for keymap compilation. */
 enum xkb_keymap_compile_flags {
     /** Do not apply any flags. */
-    XKB_KEYMAP_COMPILE_NO_FLAGS = 0
+    XKB_KEYMAP_COMPILE_NO_FLAGS = 0,
+    XKB_KEYMAP_COMPILE_RANGE_REDIRECT_TO_0 = (1 << 0),
+    XKB_KEYMAP_COMPILE_RANGE_SATURATE = (1 << 1),
+    _XKB_KEYMAP_COMPILE_ALL =                    \
+        ( XKB_KEYMAP_COMPILE_NO_FLAGS            \
+        | XKB_KEYMAP_COMPILE_RANGE_REDIRECT_TO_0 \
+        | XKB_KEYMAP_COMPILE_RANGE_SATURATE )
 };
 
 /**
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -69,6 +69,16 @@ xkb_keymap_new(struct xkb_context *ctx,
     keymap->format = format;
     keymap->flags = flags;
 
+    if (flags & XKB_KEYMAP_COMPILE_RANGE_REDIRECT_TO_0) {
+        keymap->out_of_range_group_action = RANGE_REDIRECT;
+        keymap->out_of_range_group_number = 0;
+    } else if (flags & XKB_KEYMAP_COMPILE_RANGE_SATURATE) {
+        keymap->out_of_range_group_action = RANGE_SATURATE;
+    } else {
+        keymap->out_of_range_group_action = RANGE_WRAP;
+    }
+    log_err(ctx, 0, "xkb_keymap_new: flags: %d, out_of_range_group_action=%u\n", flags, keymap->out_of_range_group_action);
+
     update_builtin_keymap_fields(keymap);
 
     return keymap;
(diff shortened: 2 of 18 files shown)


wismill:
Rebased and split commits (a bit).

wismill:
This new feature can be of help for shortcuts-related issues, such as: “[Switching layout temporarily while modifier is held](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/416)”.

For example, using:

```c
    key <LCTL>               {
        symbols[1] = [       {Control_L,                             ISO_First_Group } ],
        actions[1] = [       {SetMods(modifiers=Control,clearLocks), SetGroup(group=-4)}]
    };
```

*and* setting keymap `RANGE_REDIRECT` to 0, every shortcut involving `<LCTL>` will use the first layout. The `RANGE_REDIRECT` and `-4` are necessary because of how effective group computation works: it’s an *addition* of the various components (base, latched, locked), so `SetGroup(group=1)` does not set the effective group to 1.

That said, modifying actions in symbols is not really practical. But `interpret` statements only support one action and it is probably better so: we want to keep 1 keysym = 1 action.

We could, however, introduce “composite” actions, e.g. `RunSequence(action1(…), actions2(…), …)` (probably better with an array syntax). It would still count as a single action while parsing.

But in the end, I think we need a dedicated feature that:
- Add additional action to modifiers automatically;
- Handle the effective group computation properly;
- Maybe act differently if the non-modifier keysym of the shortcut is a (Latin) letter or not.

This would only internal with no special syntax, but a specialized API.

wismill:
Rebased and added exhaustive tests, that enabled to discover some memory issues. We now disallow mixing multiple modifiers or group actions, as it requires to update state handling. I may add it in the PR if it is easy enough, else will do in a further MR/release.

macOS CI is becoming very slow and fails. Looks like some issues with packaging. I do not think we had a full xorg server installed there, but now it does.

wismill:
Rebased to fix macOS CI.

wismill:
Rebased after merging #526.

wismill:
Note to self: we may want to merge #518 and #519 before this one. Or the contrary, whatever is easier.

wismill:
Rebased after merging #519.

wismill:
Rebased after merging #518.

mahkoh:
There is an issue with this. Xwayland and older xkbcommon will fail to parse keymaps with multiple actions, making this feature impossible to use in practice. I've worked around this in my implementation by not formatting such actions by default: https://github.com/mahkoh/kbvm/pull/11

wismill:
@mahkoh This is a known issue. Proper handling requires fallback and format version management, see #484. Contrary to kbvm, we have always the same number of keysyms and actions, so we are out of luck for now. #528 may help a bit.

I can see how having different number of keysyms and actions can be useful. The case 1 keysym/multiple actions would then be possible to handle as you have done. But *it is still not clear to me what the merge behavior should be*. Could you point me to your test cases?

Anyway keep in mind this is all very recent; multiple keysyms per level is not even very well supported.

I preliminary added multiple actions per level as a possible solution to the shortcut hell when using multiple layouts.

I aim to work on stability in [the next release](https://github.com/xkbcommon/libxkbcommon/milestone/4).

mahkoh:
>Contrary to kbvm, we have always the same number of keysyms and actions, so we are out of luck for now.

In that case a solution could be to send NO explicit actions for that key. The only good reason I know to have them in client-side keymaps in the first place is to allow clients to figure out how to type keysyms.

>But it is still not clear to me what the merge behavior should be. Could you point me to your test cases?

It's difficult since I wrote the testcases via mutation testing. First I modify a line in the source code in an incompatible way, then I run the tests. If no test fails, I add a new test for that line. But I don't keep track of which lines are tested by which testcases.

But here is the logic that merges the actions from the same level:

```rust
                        if new_level.actions.is_not_empty() {
                            if augment && existing_group.levels[idx].actions.is_not_empty() {
                                // emit diagnostic message
                            } else {
                                existing_group.levels[idx].actions = new_level.actions;
                                existing_group.has_explicit_actions |= new_group.has_explicit_actions;
                            }
                        }
```

wismill:
> There is an issue with this. Xwayland and older xkbcommon will fail to parse keymaps with multiple actions, making this feature impossible to use in practice. I've worked around this in my implementation by not formatting such actions by default: [mahkoh/kbvm#11](https://github.com/mahkoh/kbvm/pull/11)

See [xkbcomp MR](https://gitlab.freedesktop.org/xorg/app/xkbcomp/-/merge_requests/29). I have to check what other parts of the X11 stack require updating.

PR #488, opened 2024-07-12
Enable out-of-range group action configuration

wismill:
In #487 I added keymap compile flags to configure the out-of-range group action, but it should be separate PR. Now I am also not sure the flags are the best API for this, so I added the following:

- `xkb_keymap_set_out_of_range_layout_redirect`
- `xkb_keymap_set_out_of_range_layout_clamp`

I have not added a function for the wrap action because it is the default, but we may need it for completion.

Another question is if it’s better to add only one function and expose `xkb_range_exceed_type`?

TODO:
- [ ] Changelog entry

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -993,6 +993,33 @@ char *
 xkb_keymap_get_as_string(struct xkb_keymap *keymap,
                          enum xkb_keymap_format format);
 
+/**
+ * Set the out-of-range layout action to “redirect into range”:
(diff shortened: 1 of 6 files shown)


bluetech:
I haven't looked into the actual details, but just a general comment -- up to now `xkb_keymap` is an immutable type (well, at least except for the refcnt), which I think is a nice and useful property. Therefore I think it would be preferable if these settings were specified on keymap creation rather than after the fact.

wismill:
Closing in favor of #507

PR #489, opened 2024-07-12
keysyms: Require only 5 bytes for UTF-8 encoding

### LINKED ISSUE #418: `xkb_keysym_to_utf8()` requires output buffer size to be at least 7, but it should be 5
As I understand it, a size of 7 was needed when UTF-8 sequences could be up to 6 bytes long. Now that they're limited to 4, `xkb_keysym_to_utf8()`'s requirement should be reduced accordingly.

wismill:
Require only 5 bytes for the buffer of `xkb_keysym_to_utf8`, as UTF-8 encodes code points on up to 4 bytes + 1 byte for the NULL-terminating byte.

Previous standard [RFC 2279] (1998) required up to 6 bytes per code point, but has been superseded by [RFC 3629] (2003).

Note that the commit e0524296 that introduced this API is from 2012, which is 9 years after the new standard was introduced.

This PR modifies quite a lot of files (but with tiny diff), because we now enforce to use minimal buffer size internally.

Tests were already added in #414.

[RFC 2279]: https://datatracker.ietf.org/doc/html/rfc2279
[RFC 3629]: https://datatracker.ietf.org/doc/html/rfc3629

Fixes #418

### DIFF
--- a/changes/api/418.bugfix.md
+++ b/changes/api/418.bugfix.md
@@ -1,0 +1,5 @@
+`xkb_keysym_to_utf8`: Require only 5 bytes for the buffer, as UTF-8 encodes code
+points on up to 4 bytes + 1 byte for the NULL-terminating byte.
+Previous standard [RFC 2279](https://datatracker.ietf.org/doc/html/rfc2279)
+(1998) required up to 6 bytes per code point, but has been superseded by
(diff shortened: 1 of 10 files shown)


whot (approved):
seems correct :)

PR #491, opened 2024-07-12
build: Check for --undefined-version support

### LINKED ISSUE #481: meson.build script broke detection of --version-script support in GNU ld, as NetBSD uses
_Originally posted by @glechapelain in https://github.com/xkbcommon/libxkbcommon/issues/459#issuecomment-2134184602:_

> I'll verify here but this seems to be causing grieves to pkgsrc using GNU: https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58272

Relevant PR: #459

wismill:
Gate the use of `--undefined-version` in the linker because it breaks on older GNU `ld`: https://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=58272.

Fixes #481

@glechapelain @eli-schwartz

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,7 +7,7 @@ project(
         'warning_level=2',
         'b_lundef=true',
     ],
-    meson_version : '>= 0.52.0',
+    meson_version : '>= 0.58.0', # Released on May 2021
 )
 pkgconfig = import('pkgconfig')
 cc = meson.get_compiler('c')
@@ -270,9 +270,7 @@ dep_libxkbcommon = declare_dependency(
     link_with: libxkbcommon,
     include_directories: include_directories('include'),
 )
-if meson.version().version_compare('>= 0.54.0')
-    meson.override_dependency('xkbcommon', dep_libxkbcommon)
-endif
+meson.override_dependency('xkbcommon', dep_libxkbcommon)
 pkgconfig.generate(
     libxkbcommon,
     name: 'xkbcommon',
@@ -341,9 +339,7 @@ You can disable X11 support with -Denable-x11=false.''')
         link_with: libxkbcommon_x11,
         include_directories: include_directories('include'),
     )
-    if meson.version().version_compare('>= 0.54.0')
-        meson.override_dependency('xkbcommon-x11', dep_libxkbcommon_x11)
-    endif
+    meson.override_dependency('xkbcommon-x11', dep_libxkbcommon_x11)
     pkgconfig.generate(
         libxkbcommon_x11,
         name: 'xkbcommon-x11',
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -143,9 +143,14 @@ configh_data.set('_CRT_NONSTDC_NO_DEPRECATE', 1)
...
 # Supports -Wl,--version-script?
+if cc.has_link_argument('-Wl,--undefined-version')
+    extra_linker_args = ',--undefined-version'
+else
+    extra_linker_args = ''
+endif
 have_version_script = cc.links(
     'int main(){}',
-    args: '-Wl,--undefined-version,--version-script=' + meson.current_source_dir()/'xkbcommon.map',
+    args: '-Wl,--version-script=' + meson.current_source_dir()/'xkbcommon.map' + extra_linker_args,
```
whot, line 153: fwiw, meson has for quite a while supported f-strings so this may be easier as:
```suggestion
    args: f'-Wl,--version-script=@map@@extra_linker_args@',
```
with an extra
```
map = meson.current_source_dir()/'xkbcommon.map
```
above somewhere

whot (approved):

meson.build
```diff
@@ -143,9 +143,14 @@ configh_data.set('_CRT_NONSTDC_NO_DEPRECATE', 1)
 configh_data.set('WIN32_LEAN_AND_MEAN', 1)
 
 # Supports -Wl,--version-script?
+if cc.has_link_argument('-Wl,--undefined-version')
+    extra_linker_args = ',--undefined-version'
+else
+    extra_linker_args = ''
+endif
```
eli-schwartz, line 150: Another option would be to assume it's only needed if `cc.get_linker_id() == 'ld.lld'`, which can be done as a string comparison instead of compiling a test program to check `has_link_argument`

wismill: From a maintainer POV, it’s much more robust to use the builtin features of the build tool. And again, GNU ld does support this flag if recent enough.

meson.build
wismill: @whot Good idea. I raised minimal meson version to from `0.52` (October 2019) to `0.58` (May 2021) to enable f-strings.

meson.build
eli-schwartz: Whether GNU ld supports the flag is irrelevant -- what matters is whether GNU ld defaults to no-undefined-version, as it is the change in default that caused the original lld issue.

Which it currently doesn't. So handling this on GNU ld becomes a question of whether you:
- future-proof the code against a theoretical future change in the GNU binutils defaults (I personally am skeptical they will ever change, but that's a random personal opinion).
- simplify the meson.build support code and optimize the configure time by a small amount.

There are arguments to be made both ways. I did say it was another option, not that you must choose it.

meson.build
wismill: Given that #313 updated to a meson version of barely one year old at that time, I guess 3 years old is OK here.

PR #492, opened 2024-07-16
Scan-build and covscan fixes

whot:
This fixes a bunch of minor covscan/scan-build complaints, mostly in the test code.

I think there's only one legitimate bug in `test/registry.c` where we need `sizeof()` -> `ARRAY_SIZE()`. The rest is mostly to shut up the analyziers.

There are still a few left over but they're deep in the parsing code and I couldn't immediately tell if they were actual bugs (I think they are false positives).

### DIFF
--- a/test/keysym.c
+++ b/test/keysym.c
@@ -391,7 +391,7 @@ main(void)
         test_icu_case_mappings(ks);
 #endif
     }
-    iter = xkb_keysym_iterator_unref(iter);
+    xkb_keysym_iterator_unref(iter);
     assert(ks_prev == XKB_KEYSYM_MAX_ASSIGNED);
     assert(count == XKB_KEYSYM_UNICODE_MAX - XKB_KEYSYM_UNICODE_MIN + 1 + count_non_unicode);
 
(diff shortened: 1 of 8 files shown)


test/keysym.c
```diff
@@ -391,7 +391,7 @@ main(void)
         test_icu_case_mappings(ks);
 #endif
     }
-    iter = xkb_keysym_iterator_unref(iter);
+    xkb_keysym_iterator_unref(iter);
```
wismill, line 394: I remember you actually [asked me](https://github.com/xkbcommon/libxkbcommon/pull/414#discussion_r1421837962) 😉 to add this to avoid dangling pointers. I guess it does not matter here because we do not reuse the variable later.

wismill (approved):
LGTM. I guess there is still quite a lot of work to make the analyzers happy, especially with respect to implicit casts.

I am going to merge this soon. Should I squash the commits?

test/keysym.c
whot: yeah. I'd personally prefer resetting it because it's a safety operation but tools like scan-build are useful so keeping them noise-free unfortunately trumps my personal preference :)

I'd probably leave the commits as-is but if you want to squash them please do so, I have no preference on that.

PR #496, opened 2024-07-30
build: Set the ABI version properly

### LINKED ISSUE #494: Please increment minor DSO version
Basically stealing @smcv's well-written rationale in [wayland#175](https://gitlab.freedesktop.org/wayland/wayland/-/issues/175): can we please increment the libxkbcommon (and -x11) minor versions every time new ABI is added, and also ideally increment the patch version for every release?

I'm not going to try to paraphrase Simon's writeup here because he's expressed it far better than I ever could ...

Thanks!

wismill:
Rational (from: https://gitlab.freedesktop.org/wayland/wayland/-/issues/175):

The ABI of a shared library on Linux is given by a major version, which is part of the SONAME and is incremented (rarely) on incompatible changes, and a minor version, which is part of the basename of the regular file to which the SONAME provides a symlink. It's conventional to manage the ABI version in the style of semantic versioning (major.minor.micro where the major version goes up for incompatible changes, the minor version goes up for new ABI, or the micro version goes up for internal fixes), although this is not strictly required.

The minor version is used by `ldconfig(8)` and by some projects to choose which of potentially several copies of a library is the newest; keeping it the same means we can't tell, and will potentially choose an outdated version.

---

In xkbcommon we did not have an API break for a long time, so in order to avoid an unnecessary SONAME bump, let’s keep the major version version at 0, e.g. xkbcommon 1.x.y produces `libxkbcommon.so.0.x.y`.

The same goes for `libxkbcommon-x11` and `libxkbregistry`.

~~Note: each library has its own handling of the major version, although right now it is identical.~~

Fixes #494

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -9,6 +9,22 @@ project(
     ],
     meson_version : '>= 0.58.0', # Released on May 2021
 )
+
+xkbcommon_project_version = meson.project_version().split('.')
+if xkbcommon_project_version[0] != '1'
+    # The versioning used for the shared libraries assumes that the major
+    # version of xkbcommon as a whole will increase to 2 if and only if there
+    # is an ABI break, at which point we should probably bump the SONAME of
+    # all libraries to .so.2.
+    error('We probably need to bump the SONAME of libxkbcommon')
+else
+    # To avoid an unnecessary SONAME bump, xkbcommon 1.x.y produces
+    # libxkbcommon.so.0.x.y, libxkbcommon-x11.so.0.x.y, libxkbregistry.so.0.x.y.
+    soname_version = '.'.join(
+        ['0', xkbcommon_project_version[1], xkbcommon_project_version[2]]
+    )
+endif
+
 pkgconfig = import('pkgconfig')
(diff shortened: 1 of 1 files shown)


fooishbar (approved):
This looks great, thankyou!

meson.build
```diff
@@ -385,6 +412,19 @@ if get_option('enable-xkbregistry')
...
+        # is an ABI break, at which point we should probably bump the SONAME of
+        # all libraries to .so.2.
+        error('We probably need to bump the SONAME of libxkbregistry')
+    else
+        # To avoid an unnecessary SONAME bump, xkbcommon 1.x.y produces
+        # libxkbregistry.so.0.x.y.
+        libxkbregistry_version = '.'.join(
+            ['0', xkbcommon_project_version[1], xkbcommon_project_version[2]]
+        )
+    endif
```
whot, line 427: how about moving this somewhere near the top and set a more generic `soname_version` variable that is then used for all three .so files. I don't se a future where we need to have different sonames for the three libraries, so this could be done without duplicating.

PR #499, opened 2024-08-20
rules: Fix wild card handling

### LINKED ISSUE #497: rules: setxkbmap deals with empty component differently
Consider the following dummy rules file `~/.config/xkb/rules/dummy`:

```
! model		=	keycodes
  *		=	evdev

! model		=	geometry
  *		=       pc(pc104)

! layout	variant	=	symbols
  *		*	=	pc+%l%(v)

! layout[1]	variant[1]	=	symbols
  *		*		=	pc+%l[1]%(v[1])

! layout[2]	variant[2]	=	symbols
  *		*		=	+%l[2]%(v[2]):2

! layout[3]	variant[3]	=	symbols
  *		*		=	+%l[3]%(v[3]):3

! layout[4]	variant[3]	=	symbols
  *		*		=	+%l[4]%(v[4]):4

! model		layout		=	compat
  *		*		=	complete

! model		layout[1]	=	compat
  *		*		=	complete

! model		=	types
  *		=	complete
```

`setxkbmap` and `xkbcli compile-keymap` give different result with layout = `a,b`:

```bash
# Note: for setxkbmap you need a corresponding .lst file.
# e.g. `ln -s /usr/share/X11/xkb/rules/evdev.lst ~/.config/xkb/rules/dummy.lst`

###################
# Without variant #
###################

# No symbol match!
> setxkbmap -print -rules ~/.config/xkb/rules/dummy -layout a,b
xkb_keymap {
        xkb_keycodes  { include "evdev" };
        xkb_types     { include "complete"      };
        xkb_compat    { include "complete"      };
        xkb_geometry  { include "pc(pc104)"     };
};

# Not working, “same” as setxkbmap
> ./build/compile-keymap --kccgst --rules dummy --layout a,b

######################
# With empty variant #
######################

# No symbol match!
> setxkbmap -print -rules ~/.config/xkb/rules/dummy -layout a,b -variant ,
xkb_keymap {
        xkb_keycodes  { include "evdev" };
        xkb_types     { include "complete"      };
        xkb_compat    { include "complete"      };
        xkb_geometry  { include "pc(pc104)"     };
};

# Working, contrary to setxkbmap!
> ./build/compile-keymap --kccgst --rules dummy --layout a,b --variant ,
xkb_keymap {
  xkb_keycodes { include "evdev" };
  xkb_types { include "complete" };
  xkb_compat { include "complete" };
  xkb_symbols { include "pc+a+b:2" };
};

#################
# With variants #
#################

# Symbol match!
> setxkbmap -print -rules ~/.config/xkb/rules/dummy -layout a,b -variant x,y
xkb_keymap {
        xkb_keycodes  { include "evdev" };
        xkb_types     { include "complete"      };
        xkb_compat    { include "complete"      };
        xkb_symbols   { include "pc+a(x)+b(y):2"        };
        xkb_geometry  { include "pc(pc104)"     };
};

# Working, same as setxkbmap
> ./build/compile-keymap --kccgst --rules dummy --layout a,b --variant x,y
xkb_keymap {
  xkb_keycodes { include "evdev" };
  xkb_types { include "complete" };
  xkb_compat { include "complete" };
  xkb_symbols { include "pc+a(x)+b(y):2" };
};
```

So we somehow match on *empty* variant `--layout x,y --variant ,`, but not when variant is missing `--layout x,y` (or `--layout x,y --variant ""`).

This neither consistent nor behaves as `setxkbmap`, so this is a bug. Agreed, this does not trigger yet because the current rules of `xkeyboard-config` have cover missing variant, but I did encounter this bug while working on [`xkeyboard-config#478`](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/478#note_2513592).

Actions to take:
1. Add a test
2. Fix the bug
3. Update the rules doc page about the wildcard behavior. I initially thought it matched everything, even missing component, but it _does_ require a component value (here a variant) in order to match.

wismill:
The handling of wild card `*` is different in libxkbfile and X server: wild card matches empty strings for model and option but not for layout nor variant, while in libxkbcommon wild cards always match empty strings.

See:
- https://gitlab.freedesktop.org/xorg/lib/libxkbfile/-/blob/bf985c68acb1244f51ec91414532a2347fbc1c4c/src/maprules.c#L687
- https://gitlab.freedesktop.org/xorg/lib/libxkbfile/-/blob/bf985c68acb1244f51ec91414532a2347fbc1c4c/src/maprules.c#L712

The difference of handling between the components is unfortunately not documented, but we should follow the behavior of the original implementations for consistency.

- Fixed by implementing the same behavior than libxkbfile.
- Added tests and fixed failing tests.

Fixes #497

### DIFF
--- a/src/xkbcomp/rules.c
+++ b/src/xkbcomp/rules.c
@@ -234,6 +234,13 @@ enum mlvo_match_type {
     MLVO_MATCH_GROUP,
 };
 
+enum wildcard_match_type {
+    /* Match only non-empty strings */
+    WILDCARD_MATCH_NONEMPTY = 0,
+    /* Match all strings */
+    WILDCARD_MATCH_ALL,
+};
+
 struct rule {
     struct sval mlvo_value_at_pos[_MLVO_NUM_ENTRIES];
     enum mlvo_match_type match_type_at_pos[_MLVO_NUM_ENTRIES];
@@ -694,10 +701,12 @@ match_group(struct matcher *m, struct sval group_name, struct sval to)
 
 static bool
 match_value(struct matcher *m, struct sval val, struct sval to,
-            enum mlvo_match_type match_type)
+            enum mlvo_match_type match_type,
+            enum wildcard_match_type wildcard_type)
 {
     if (match_type == MLVO_MATCH_WILDCARD)
-        return true;
+        /* Wild card match empty values only if explicitly required */
+        return wildcard_type == WILDCARD_MATCH_ALL || !!to.len;
     if (match_type == MLVO_MATCH_GROUP)
         return match_group(m, val, to);
     return svaleq(val, to);
@@ -705,9 +714,10 @@ match_value(struct matcher *m, struct sval val, struct sval to,
 
 static bool
 match_value_and_mark(struct matcher *m, struct sval val,
-                     struct matched_sval *to, enum mlvo_match_type match_type)
+                     struct matched_sval *to, enum mlvo_match_type match_type,
+                     enum wildcard_match_type wildcard_type)
 {
-    bool matched = match_value(m, val, to->sval, match_type);
+    bool matched = match_value(m, val, to->sval, match_type, wildcard_type);
     if (matched)
         to->matched = true;
     return matched;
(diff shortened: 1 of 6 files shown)


src/xkbcomp/rules.c
```diff
@@ -694,20 +694,22 @@ match_group(struct matcher *m, struct sval group_name, struct sval to)
 
 static bool
 match_value(struct matcher *m, struct sval val, struct sval to,
-            enum mlvo_match_type match_type)
+            enum mlvo_match_type match_type, bool wildcard_matches_empty)
```
whot, line 697: how about something like this:
```c
enum wildcard_match_type {
   WILDCARD_MATCH_EMPTY,
   WILDCARD_MATCH_NONEMPTY,
};
```
and then the invocations would be more readable than a plain `true`/`false`.

test/data/rules/wildcard
```diff
@@ -0,0 +1,32 @@
...
+  *		*		=	complete
+
+! model		layout[1]	=	compat
+  *		*		=	complete
+
+! model		layout[2]	=	compat
+  *		*		=	complete
+
+! model		=	types
+  *		=	complete
```
whot, line 32: might be github's diff view but it looks like the indentation/alignment here is quite off which makes this file a bit hard to look at.

whot (approved):

whot:
let's cc @daniels and @bluetech explicitly, if this differs from the original implementation there may be a reason for it they remember?

src/xkbcomp/rules.c
wismill: No boolean blindness? Good idea. Although it shines much more in public API.

test/data/rules/wildcard
wismill: You need to set tab size to 8, as in `xkeyboard-config`. I changed it to spaces to avoid this issue.

whot: yeah, it rendered fine in the hunk above but this one looked weird. tab size is (as it should be!) 8 by default anyway, I didn't even know that was a setup in github :)

Anyway, all good, thanks!

wismill:
@fooishbar @bluetech kind reminder 🙂

bluetech (approved):
Looks good to me, I probably just didn't notice this difference in libxkbfile. Thanks!

PR #502, opened 2024-09-02
Add XKB_LED_NAME_COMPOSE and XKB_LED_NAME_KANA

C0rn3j:
The end goal is to be able to press `Num Lock`, `Caps Lock`, `Scroll Lock`, `Compose` or `Kana` and have either of the 5 LEDs on my QMK keyboard light up.

I could use a bit of help here.

This is a continuation of https://gitlab.freedesktop.org/libinput/libinput/-/merge_requests/990 and is directly related to the [current WIP for kwin](https://invent.kde.org/plasma/kwin/-/merge_requests/5627) for adding the same two LED keys.

[HID spec](https://usb.org/sites/default/files/hid1_11.pdf) defines five LEDs, but only the first 3 are currently supported by XKB and most of the Linux tooling:

image

I am not familiar with XKB, where else do I need to put the necessary definitions?

### DIFF
--- a/changes/api/+compose-and-kana-led-names.features.md
+++ b/changes/api/+compose-and-kana-led-names.features.md
@@ -1,0 +1,4 @@
+`xkbcommon-names.h`: Add `XKB_LED_NAME_COMPOSE` and `XKB_LED_NAME_KANA` definitions
+to cover all LEDs defined in [USB HID](https://usb.org/sites/default/files/hid1_11.pdf).
+
+Contributed by Martin Rys
--- a/include/xkbcommon/xkbcommon-names.h
+++ b/include/xkbcommon/xkbcommon-names.h
@@ -31,6 +31,7 @@
  * @brief Predefined names for common modifiers and LEDs.
  */
 
+/* Real modifiers names are hardcoded in libxkbcommon */
 #define XKB_MOD_NAME_SHIFT      "Shift"
 #define XKB_MOD_NAME_CAPS       "Lock"
 #define XKB_MOD_NAME_CTRL       "Control"
@@ -38,8 +39,13 @@
 #define XKB_MOD_NAME_NUM        "Mod2"
 #define XKB_MOD_NAME_LOGO       "Mod4"
 
-#define XKB_LED_NAME_CAPS       "Caps Lock"
+/* LEDs names are encoded in xkeyboard-config, in the keycodes and compat files.
+ * They have been stable since the beginning of the project and are unlikely to
+ * ever change. */
 #define XKB_LED_NAME_NUM        "Num Lock"
+#define XKB_LED_NAME_CAPS       "Caps Lock"
 #define XKB_LED_NAME_SCROLL     "Scroll Lock"
+#define XKB_LED_NAME_COMPOSE    "Compose"
+#define XKB_LED_NAME_KANA       "Kana"
 
 #endif


wismill:
Thanks for your contribution. LGTM, these names have been stable for 17 years in `xkeyboard-config`. But since this is an area of XKB in which I have no experience, I am going to put this on hold while I investigate it. Thanks for the links.

wismill:
@C0rn3j I converted to a draft and removed the “WIP” prefix because it’s clearer in Github’s interface; feel free to use the “ready for review” button when you find it relevant.

wismill:
I think I got it. For the current use case, the name is a just an indirect way to access the [LED index in the `keycodes` XKB file](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/17261b995a423643fad714db6c49dea45e53cb03/keycodes/evdev#L602). The [corresponding `compat` entry](https://xkbcommon.org/doc/current/keymap-text-format-v1.html#indicator-effect) is more complex but irrelevant here.

@C0rn3j could you add a [changelog entry](https://github.com/xkbcommon/libxkbcommon/tree/master/changes/README.md) for the API?

C0rn3j:
Oh, so I have by chance used the existing correct names? Neat!

I have not tested this (nor libinput) with my kwin PR yet, I need to first prepare my keyboard FW and figure out how to swap out kwin, but I suppose this PR works as is anyway, even if subsequent additions will be needed later.

I have added the changelog fragment.

---

```
	indicator 1  = "Caps Lock";
	indicator 2  = "Num Lock";
```

On a side note, I find it curious that Num and Caps is swapped in the X components, it's equally wrong here and in [evdev](https://gitlab.freedesktop.org/xorg/driver/xf86-input-evdev/-/merge_requests/8), so it ends up working. I see you have found my post there while I was writing my reply.

I also wonder what's causing Scroll Lock to break in both X and Wayland, but that's for another issue.

changes/api/502.compose-and-kana.feature.md
```diff
@@ -0,0 +1 @@
+Add `XKB_LED_NAME_COMPOSE` and `XKB_LED_NAME_KANA` definitions to conform to USB HID.
```
wismill, line 1:
```suggestion
`xkbcommon-names.h`: Add `XKB_LED_NAME_COMPOSE` and `XKB_LED_NAME_KANA` definitions
to cover all LEDs defined in [USB HID](https://usb.org/sites/default/files/hid1_11.pdf).

Contributed by Martin Rys
```

The *conformance* to HID is at driver level, not here. `libxkbcommon` just provides theses names as a convenience, but the actual indexes are defined in `xkeyboard-config`, as commented previously.

wismill: Also please change the file name to `+compose-and-kana-led-names.features.md`, as `towncrier` cannot handle both issue and PR ID and we chose issue ID.

wismill (changes requested):

wismill:
> On a side note, I find it curious that Num and Caps is swapped in the X components, it's equally wrong here and in [evdev](https://gitlab.freedesktop.org/xorg/driver/xf86-input-evdev/-/merge_requests/8), so it ends up working.

@C0rn3j I am actually wondering the same! Do the compositor or toolkit fixes it?

wismill:
I think I understand now: LED index in XKB is only an implementation detail at least on Wayland, where the XKB LED state is queried by the LED’s *name*. Maybe the index is used by old drivers though, as you mentioned in [`xf86-input-evdev`](https://gitlab.freedesktop.org/xorg/driver/xf86-input-evdev/-/merge_requests/8). So I guess we will never invert the indexes of `CapsLock` and `NumLock`, because it will require to sync all the drivers relying on the indexes and 1) nobody cares 2) nobody would take the risk for the satisfaction of having the same USB HID order. By the way the XKB stuff predates USB!

whot:
I *think* this is correct though as for the XKB state I'm a bit unclear myself.

wismill:
I investigated a bit some code bases:
- In `xf86-input-evdev`, the mapping X11 LED ↔ device indicator is [harcoded](https://gitlab.freedesktop.org/xorg/driver/xf86-input-evdev/-/blob/917c2ea204ab03250cd6e022a7de1e51652ac420/src/evdev.c?page=2#L1117).
- `xf86-input-libinput` also [hardcodes LEDs](https://gitlab.freedesktop.org/xorg/driver/xf86-input-libinput/-/blob/bb41cc730c38570f1f90240dcfffb7013895d110/src/xf86libinput.c?page=2#L1175). @C0rn3j you may want to send a MR there too.
- In Plasma, this mapping is not hardcoded and instead is [queried using the XKB indicator name](https://invent.kde.org/plasma/kwin/-/blob/a01b71f9226d2eeac1ae104598544d901013a3e5/src/xkb.cpp#L664).
- Gtk *Wayland* also [queries the LEDs by their name](https://gitlab.gnome.org/GNOME/gtk/-/blob/b18c6b5781ba79b1cefd21ea8e3349ac753ce682/gdk/wayland/gdkkeymap-wayland.c#L109), as well as [Mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/b20e0370aa804bef8e4d6b88630743da76d26b99/src/backends/native/meta-seat-impl.c?page=3#L2954). I do not know on X11 though. 
@C0rn3j some hero will still have to propose the changes to Gtk/Mutter to support the remaining LEDs.

So its seems that the correct mapping depends very much on the type of session and the driver used. At this point it looks easier to fix in `xkeyboard-config` rather than in the drivers.

We should really encourage the use of `XKB_LED_*` in xkbcommon, in order to query the LED state by its *name*, not the *index* defined in the keymap. So this MR is a good step.

Last point: `xkeyboard-config` should really have some comments about these mappings and names. I am going to send a MR there to inform indicators indexes and names must stay stable.

wismill:
Added comment about the origin of the names and squashed.

C0rn3j:
I knew about `mutter` and possibly `wlroots` needing similar treatment too, but would not have guessed `gtk` and `xf86-input-libinput`, I'll most likely get to them too, since I already started kicking the can.

Thanks a lot for your input and investigation!

C0rn3j:
I messed this up, will send a fix up PR soon.

PR #505, opened 2024-09-11
xkbcli how-to-type: Enhance arguments parsing & doc

wismill:
Currently the positional parameter of the CLI is either a Unicode code
point or a keysym. However their respective format is not documented.

It turns out that there are multiple issues due to the use of `strtol`:
- Code points can be parsed as octal, decimal and hexadecimal while
  keysyms can only be parsed as hexadecimal. Some programs outputs
  keysyms in their decimal form (e.g. `wev`) so it is worth to bring
  symmetry with code points.
- Octal format is unusual for both and is triggered by leading zeros,
  which is unintuitive in this context.
- `U+NNNN` format is the standard format for Unicode code points but is
  not supported.
- Plain characters are not supported, e.g.: a, é, ß, Æ, γ, 🦆, etc.
  Although this is probably the easiest format for most users.

Fixed the issues above:
- Allow the code point to be passed exactly in the following formats:
  - Literal character (requires UTF-8 character encoding of the terminal);
  - Decimal number;
  - Hexadecimal number: either `0xNNNN` or `U+NNNN`.
- Allow the keysym to be passed exactly in the following formats:
  - Decimal number;
  - Hexadecimal number: either `0xNNNN`;
  - Name.
- Improve both `--help` message and manual.

### DIFF
--- a/changes/tools/+how-to-type-format.feature.md
+++ b/changes/tools/+how-to-type-format.feature.md
@@ -1,0 +1,11 @@
+`xkbcli how-to-type`: added new input formats and their corresponding documentation.
+
+*Unicode code points* can be passed in the following formats:
+- Literal character (requires UTF-8 character encoding of the terminal);
+- Decimal number;
+- Hexadecimal number: either `0xNNNN` or `U+NNNN`.
+
+*Keysyms* can to be passed in the following formats:
+- Decimal number;
+- Hexadecimal number: `0xNNNN`;
+- Name.
--- a/meson.build
+++ b/meson.build
@@ -753,7 +753,13 @@ test(
 )
 test(
     'utf8',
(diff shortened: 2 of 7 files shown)


wismill:
Hmm the use of `strtol` with base `0` parses string starting with `0` as octal. We should probably strip leading zeros to parse only decimal and hexadecimal.

wismill:
A big part of the new code is about UTF-8 to UTF-32 decoding. It’s kept as an internal API for now. I added the related tests.

wismill:
Note that we could go further and try to parse a keysym *name* without the `--keysym` flag, if previous attempts to parse as a code point / plain character failed. Maybe a new flag `--codepoint` would be useful then for completeness.

PR #507, opened 2024-09-17
Enable out-of-range group action configuration using keymap flags (take 2)

wismill:
In #487 I added keymap compile flags to configure the out-of-range group action, but it should be separate PR. 

Note that I proposed #488 previously but @bluetech [advised to keep `xkb_keymap` “immutable”](https://github.com/xkbcommon/libxkbcommon/pull/488#issuecomment-2336825700).


wismill:
Missing: get and use the corresponding control value from X11 for libxbbcommon-x11.

wismill:
The issue with the current implementation is that we are using flag positions to encode the target layout number, but with #515 coming this will not scale. We could keep the flag, but add a new function to change the target layout (default: 0).

A use case for having target layout different of 0 is tweaking the shortcuts (Ctrl, Alt, etc.) to change the group, in order to make e.g. non Latin layout work without relying on the toolkit to look for Latin letter in other layouts. This would most likely require #487.

wismill:
Closing in favor of #516.

PR #508, opened 2024-09-20
meson: Enable linker garbage collection

wismill:
Alternative to #427 [suggested](https://github.com/xkbcommon/libxkbcommon/pull/427#issuecomment-2336828098) by @bluetech.

Adapted from [systemd](https://github.com/systemd/systemd/blob/978e7d166c362302ce2302ac36efe9717d642d90/meson.build#L482-L489).

I am not an expert on the topic but got curious. From [gcc doc](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffunction-sections):

> Only use these options when there are significant benefits from doing so. When you specify these options, the assembler and linker create larger object and executable files and are also slower. These options affect code generation. **They prevent optimizations by the compiler and assembler** using relative locations inside a translation unit since the locations are unknown until link time. An example of such an optimization is relaxing calls to short call instructions.

How can we tell accurately if this is interesting for us and not affecting perf?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -48,6 +48,15 @@ cflags = [
     '-Wwrite-strings',
     '-Wno-documentation-deprecated-sync',
 ]
+if get_option('buildtype') != 'debug'
+    cflags += [
+            '-ffunction-sections',
+            '-fdata-sections',
+    ]
+    base_link_args = ['-Wl,--gc-sections']
+else
+    base_link_args = []
+endif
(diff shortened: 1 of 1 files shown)


whot:
fwiw, I generally tend to say "leave this up to the distributions". The vast majority of our install base is distribution-based packages which can make decisions for compiler flags at the distribution level. Except for warnings aimed at developers I usually don't bother.

wismill:
Fair enough, closing.

PR #509, opened 2024-09-20
keysyms: Fix off-by-one XKB_KEYSYM_NAME_MAX_SIZE

wismill:
The constant did not account for the terminating `NULL` byte and this was sadly not caught by the tests.

Fixed the invalid value, the corresponding script and the tests.


wismill:
I am surprised it was not caught by at least `xkeyboard-config` tests.

PR #512, opened 2024-09-23
Support querying whether virtual modifiers are active (alternative 2)

### LINKED ISSUE #88: API for reading xkb_mod_index_t to keysym mapping
The documentation could be improved for xkb_keymap_mod_get_name() to describe the historical meaning of Mod1-Mod5 names. For someone who has never worked with linux, using these APIs can be difficult (for example a student wanting to write a native application on Linux, without using Qt or GTK). Only very few people truly understand the internals of Virtual vs Real modifiers (the X11 legacy). For example with 'us' keymap we see that Shift + Alt = Meta:

```
# alt key on level 2 (e.g. when holding Shift) becomes Meta key
key <LALT>               {	[           Alt_L,          Meta_L ] };
```

` xkb_state_mod_name_is_active(state, "Meta", XKB_STATE_MODS_EFFECTIVE) == false`

but without deep understanding of  XKB keymaps and X11 legacy, one would expect this to return 'true';
on my system Meta is mapped to "Mod1" (default on Ubuntu). X11 clients can get mod-to-keysym mapping from X server. The same info AFAICT is present in the keymap. XKB should prove API to extract / read this mapping.

The XKB spec also says "keysyms, when bound to modifiers, affect the rules [..]", but without this API we can't know which keysyms are bound to Mod1-Mod5.

wismill:
Allow to query also *virtual* modifiers with the following functions, so they work with *any* modifiers (real *and* virtual):
- `xkb_state_mod_index_is_active`
- `xkb_state_mod_indices_are_active`
- `xkb_state_mod_name_is_active`
- `xkb_state_mod_names_are_active`
- `xkb_state_mod_index_is_consumed`
- `xkb_state_mod_index_is_consumed2`

**Warning:** they may overmatch in case there are overlappings virtual-to-real
modifiers mappings.

Also added the following modifiers names definitions in `xkbcommon-names.h`:
- `XKB_MOD_NAME_MOD1`
- `XKB_MOD_NAME_MOD2`
- `XKB_MOD_NAME_MOD3`
- `XKB_MOD_NAME_MOD4`
- `XKB_MOD_NAME_MOD5`
- `XKB_VMOD_NAME_ALT`
- `XKB_VMOD_NAME_META`
- `XKB_VMOD_NAME_NUM`
- `XKB_VMOD_NAME_SUPER`
- `XKB_VMOD_NAME_HYPER`
- `XKB_VMOD_NAME_LEVEL3`
- `XKB_VMOD_NAME_LEVEL5`
- `XKB_VMOD_NAME_SCROLL`

---

This is a simplier version of #353, which is also an alternative to #36.

This is necessary to implement fully #450.

Fixes #88


wismill:
TODO: more tests, with all modifiers.

wismill:
We should probably deprecate the following in favor of the new names:
- `XKB_MOD_NAME_ALT` → `XKB_VMOD_NAME_ALT`
- `XKB_MOD_NAME_LOGO` → `XKB_VMOD_NAME_SUPER`
- `XKB_MOD_NAME_NUM` → `XKB_VMOD_NAME_NUM`

src/state.c
```diff
@@ -1523,7 +1557,11 @@ xkb_state_mod_index_is_consumed2(struct xkb_state *state, xkb_keycode_t kc,
     if (!key || idx >= xkb_keymap_num_mods(state->keymap))
         return -1;
 
-    return !!((1u << idx) & key_get_consumed(state, key, mode));
+    xkb_mod_mask_t mapping = get_mod_mapping(state->keymap, idx);
+    if (!mapping)
+        /* Modifier not mapped */
+        return 0;
```
whot, line 1563: same here

whot (approved):
I think this LGTM though admittedly I haven't managed to page everything back in in detail. Two nitpicks, otherwise :+1:

src/state.c
wismill: Fixed

src/state.c
```diff
@@ -1265,6 +1265,27 @@ mod_mask_get_effective(struct xkb_keymap *keymap, xkb_mod_mask_t mods)
     return mask;
 }
 
+/* [WARNING] Caller must ensure that the modifier index is valid! */
+static inline bool
+is_real_mod(struct xkb_keymap *keymap, xkb_mod_index_t mod) {
+    return !!(keymap->mods.mods[mod].type & MOD_REAL);
```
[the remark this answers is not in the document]
wismill: This check has to be done by the caller, as the comment states. My idea was to use a proper (inlined) function rather than a macro, for better typing/semantics. This is a very local function, but I could prefix it with `unsafe_` to highlight it.

test/state.c
```diff
@@ -36,6 +36,25 @@
...
 
+#define xkb_mod_index_t_true xkb_mod_index_t
+#define xkb_mod_index_t_1    xkb_mod_index_t
+#define xkb_mod_index_t_false
+#define xkb_mod_index_t_0
+
+#define make_mod_idx(declare, keymap, mod, name)                            \
+    xkb_mod_index_t_##declare mod = xkb_keymap_mod_get_index(keymap, name); \
+    assert(mod != XKB_MOD_INVALID)
+#define make_mod_idx2(declare, keymap, mod, name) \
```
bluetech, line 47: This seems unused unless I'm missing something

test/state.c
```diff
@@ -36,6 +36,25 @@
  * keycode set (where ESC is 9). */
 #define EVDEV_OFFSET 8
 
+#define xkb_mod_index_t_true xkb_mod_index_t
```
bluetech, line 39: Heh it took me (more than...) a minute to understand what you did here :smile: 

I think it's a bit too clever. IIUC the main purpose of `make_mod_idx` is to add the assert, if so then I'd a static inline wrapper around `xkb_keymap_mod_get_index` which adds the assert, but I'd leave the declaration and assignment in the code, not in a macro.

test/state.c
```diff
@@ -274,95 +358,140 @@ test_serialisation(struct xkb_keymap *keymap)
...
     xkb_state_update_mask(state, base_mods, latched_mods, locked_mods,
                           base_group, latched_group, locked_group);
 
-    assert(xkb_state_mod_index_is_active(state, ctrl, XKB_STATE_MODS_DEPRESSED) > 0);
-    assert(xkb_state_mod_index_is_active(state, ctrl, XKB_STATE_MODS_EFFECTIVE) > 0);
+    assert(xkb_state_mod_index_is_active(state, ctrlIdx, XKB_STATE_MODS_DEPRESSED) > 0);
+    assert(xkb_state_mod_index_is_active(state, ctrlIdx, XKB_STATE_MODS_EFFECTIVE) > 0);
+
+    const struct test_active_mods_entry test_data[] = {
+        { 0,            0                         },
```
bluetech, line 383: Spelling out the fields would make it read more clearly

```suggestion
        { .state = 0,            .active = 0                         },
```

bluetech (commented):
It will be very cool if we can make this work.

I only did a very shallow skim of the code. I too have mostly forgotten the details unfortunately. Some questions:

1. Is the overmatching possibility fundamental to the way virtual modifiers work, or is it possible to rework the internals such that they track particular vmods?

2. Does this work for xkbcommon-x11?

3. In practice, I think most users ignore the `is_active` functions and just use `xkb_state_serialize_mods`, despite it being documented as for serialization only. Unfortunate, and maybe it has changed by now, but probably not, since people always take the path of least resistance. The question is -- is there anything we can do about it?

wismill:
> Some questions:
> 
>     1. Is the overmatching possibility fundamental to the way virtual modifiers work, or is it possible to rework the internals such that they track particular vmods?

@bluetech Unfortunately it’s fundamental, at least until #450 is merged. Here some examples: 
1. `Alt` and `Meta` both map to `Mod1` by default, so activating any of the 3 modifiers activates the two others. 
2. If a virtual modifier is mapped to *2* real modifiers, activating these 2 modifiers independently will also activate this vmod.

In practice, 1) is already an issue so DE and apps handle it already. 2) is probably extremely rare and more due to an accident. The only use I know is to differentiate left and right variant of a modifier.

>     2. Does this work for xkbcommon-x11?

Yes, AFAICT: in the end we rely on the real modifiers. It will start to change when #450 is merged.

>     3. In practice, I think most users ignore the `is_active` functions and just use `xkb_state_serialize_mods`, despite it being documented as for serialization only. Unfortunate, and maybe it has changed by now, but probably not, since people always take the path of least resistance. The question is -- is there anything we can do about it?

That is a good question! State and its value are supposed to be opaque, as stated by Daniel in [this Wayland issue](https://gitlab.freedesktop.org/wayland/wayland/-/issues/371#note_1838596). This may change, but apps working on states values directly expose themselves to bugs. Do you have some examples in mind?

More probably they replicate the logic, like [KDE KWin](https://gitlab.freedesktop.org/wayland/wayland/-/issues/371#note_1838596). Anyway, if they work directly with the state mod masks they must already work with *real* modifiers and I did not change the state representation/serialization with the current MR. Again, what is a real modifier will change when #450 get merged: we will have to distinguish *X11* and *non-X11* real modifiers, because “pure” virtual modifiers as stated in #450 are in fact non-X11 real modifiers.

So, until [this Wayland issue](https://gitlab.freedesktop.org/wayland/wayland/-/issues/371) declare the contrary, the only thing we can do about it is to document that the state values should only be processed by the xkbcommon API.

wismill (commented):
Updated according to @bluetech review.

PR #515, opened 2024-09-27
Raise the layout limit to 32

### LINKED ISSUE #37: Remove the 4 layout limit on Wayland.

wismill:
Time has come!

32 is probably too much, but unless we are wasting memory somewhere, I see no reason to not allow the maximum.

Fixes #37

TODO:
- [ ] Add more tests (all components, states, X11, etc.)
- [ ] Publish working rules in `xkeyboard-config`

### DIFF
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -88,20 +88,11 @@
 #include "utils.h"
 #include "context.h"
 
-/* This limit is artificially enforced, we do not depend on it any where.
- * The reason it's still here is that the rules file format does not
- * support multiple groups very well, and the rules shipped with
- * xkeyboard-config (see rules/evdev) depend on this limit extensively.
- * So just lifting this limit would cause problems for people who will use
- * more than 4 layouts.
- * TODO: Fix the group index syntax in the rules format, preferably in a
- *       backwards compatible way.
- *       See e.g. https://bugs.freedesktop.org/show_bug.cgi?id=14372
- * Note: A limit on the number of groups we *do* depend on is imposed by
- * the size of the xkb_layout_mask_t type (32). This is more than enough
- * though.
- */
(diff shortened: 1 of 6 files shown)


zzag:
On the desktop environment side, would we need to do something? or things will "just work" after updating `xkbcommon` and `xkeyboard-config`?

wismill:
@zzag For *now* we do not expose the max number of layouts, so this limit is probably set using X11's `XkbNumKbdGroups` or simply hard-coded. We should provide our own upper bound.

If a DE supports both Wayland and X11, the tricky part is how to interface with XWayland. Indeed, under a pure X11 or Wayland session, it suffices to use the respective limits.

First solutions that come to my mind:
- Change nothing 😄
- Use a 4 *moving* subset of the layouts, which should include the first one[^1] and the current one, with some *buffer*. E.g. for layouts `a,b,c,d,e,f`, if `e` is active then compile the keymap with layouts `a,d,e,f` for XWayland.

[^1]: The first layout is the default layout and may have an impact on the overall keymap.

zzag:
> By the way, what DE do you deal with?

Plasma

wismill:
> > By the way, what DE do you deal with?
> 
> Plasma

I opened [a KWin issue](https://invent.kde.org/plasma/kwin/-/issues/246) a few weeks ago.

wismill:
I think we need to introduce a new version format to handle this. `V1` would simply drop the excessive layouts > 4.

bluetech:
@wismill Congratulations on this huge achievement! Finally lifting this 30 years old limit which is so entrenched is no small feat. Really happy to see this.

PR #521, opened 2024-10-02
 state: add API for updating latched and locked mods & layout in server state

wismill:
(Revival of #314)

Up to now, the “server state” `xkb_state` API only offered one entry point to update the server state - `xkb_state_update_key`, which reflects the direct keyboard keys state. But some updates come out-of-band from keyboard input events stream, for example, a GUI layout switcher.

The X11 XKB protocol has a request which allows for such updates, `XkbLatchLockState`[^1], but xkbcommon does not have similar functionality. So server applications ended up using `xkb_state_update_state` for this, but that’s a function intended for client applications, not servers.

Add support for updating the latched & locked state of the mods and layout. Note that the depressed states cannot be updated in this way -- XKB does not expect them to be updated out of band.

[^1]: https://www.x.org/releases/X11R7.7/doc/kbproto/xkbproto.html#Querying_and_Changing_Keyboard_State

Fixes: #310
Closes: #574

TODO:
- [x] Test (pure) virtual modifiers
- [x] Changelog entry


wismill:
Rebased and fix latch-to-lock clearance side-effects.

wismill:
Refactored tests, added lots of new tests and fix numerous edges cases. This is much more tricky than expected, so the tests are crucial. We should probably also try some random config à la [QuickCheck](https://en.wikipedia.org/wiki/QuickCheck).

PR #523, opened 2024-10-04
Shortcuts tweak

wismill:

Tweak keyboard shortcuts by adding the option to switch to a different group after pressing `Control`, `Alt` or `Super`. The target group can be set for each group, although the most common use case would probably be the same target for all layouts.

This also introduces the keymap compile options API.

This can be alternatively activated with the environment variable `XKB_TARGET_SHORTCUT_LAYOUT`[^1]. The target group is then the same for all groups. This allow to use the new feature even if the DE did not implement a UI to configure it.

TODO:
- [ ] Tests
- [ ] Support latch & lock?
- [ ] Doc
- [ ] Fix the new FIXMEs
- [ ] Move option API to other MR?
- [ ] Commit messages
- [ ] Remove environment variable support?
- [ ] Log entry

[^1]: Actually the env variable will also activate this processing in clients for now, but this is a no-op, as the layout is set by the compositor, which would have already re-target it for shortcuts, if necessary.


wismill:
Note that there is another way of handling this: set the shortcut keysyms directly in the keymap, by defining new key types with e.g. `map[Control] = levelX`, etc. Some issues:
1. We need entries for every modifier combination.
2. We need the corresponding `preserve` entries, e.g. `preserve[Control] = Control`.
3. It is not easy to select the correct target layout using rules.

We could alleviate this by modifying the keymap *on the fly* from a given simple source → target layout mapping. Here we could have also fine-grained control on whether the punctuation keysym are affected or not, contrary to the current PR where it’s all-in or all-out.

ezhan:
As I understood, it is a Qt issue?
I've opened an issue against Qt: https://bugreports.qt.io/browse/QTBUG-135336

Will see that that brings.

Jules-Bertholet:
I would appreciate it if `XKB_TARGET_SHORTCUT_LAYOUT` also accepted a comma-separated list of layouts, to configure each group separately.

Issue #529, opened 2024-10-12
Crash when parsing malformed XML after rxkb_context_parse() due to error handler not being reset


sbstnk:
We got [a report of gnome-shell crashing](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7974) when libgweather is trying to parse malformed XML after gnome-shell has previously called (via `gnome-desktop`) something like:

```
rxkb_context_new();
rxkb_context_parse();
rxkb_context_unref();
```

`rxkb_context_parse()` calls `xmlSetGenericErrorFunc()` via `parse()` to set the global libxml2 error handler to its `xml_error_func()`, which depends on a valid `rxkb_context`. The context however is no longer valid after `rxkb_context_unref()`, while the error handler is still pointing to `xml_error_func()` . This then results in a crash whenever something causes libxml2 to trigger an error afterwards, such as libgweather in the gnome-shell report.

`parse()` should restore the default error handler after it is done parsing.

wismill:
Thanks for the great debugging and the corresponding PR!

PR #530, opened 2024-10-12
registry: Restore default libxml2 error handler after parsing

sbstnk:
Leaving the custom error handler could have resulted in a crash after the context has been freed.

Closes: https://github.com/xkbcommon/libxkbcommon/issues/529

### DIFF
--- a/changes/api/529.fix-xml-error-handler.bugfix.md
+++ b/changes/api/529.fix-xml-error-handler.bugfix.md
@@ -1,0 +1,4 @@
+registry: Fixed `libxml2` global error handler not reset after parsing, which
+could trigger a crash if the corresponding `rxkb_context` has been freed.
+
+Contributed by Sebastian Keller.
--- a/meson.build
+++ b/meson.build
@@ -879,7 +879,7 @@ if get_option('enable-xkbregistry')
         'registry',
         executable('test-registry', 'test/registry.c',
                    include_directories: include_directories('src'),
-                   dependencies: [dep_libxkbregistry, test_dep]),
+                   dependencies: [dep_libxkbregistry, dep_libxml, test_dep]),
         env: test_env,
     )
 endif
--- a/src/registry.c
+++ b/src/registry.c
@@ -1250,20 +1250,32 @@ parse(struct rxkb_context *ctx, const char *path,
 
     doc = xmlParseFile(path);
     if (!doc)
-        return false;
+        goto parse_error;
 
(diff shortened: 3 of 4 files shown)


sbstnk:
To test this you can use the following C program:

`gcc -I/usr/include/libxml2 ./xkb_test.c -o ./xkb_test -lxml2 -lxkbregistry`

```C
#include <xkbcommon/xkbregistry.h>
#include <libxml/parser.h>
#include <libxml/tree.h>
#include <string.h>

int main (int argc, char *argv[]) {
  struct rxkb_context *ctx;
  xmlDocPtr doc;
  const char *invalid_xml = "<test";
  char *temp;

  ctx = rxkb_context_new (0);
  rxkb_context_parse (ctx, "evdev");
  rxkb_context_unref (ctx);

  LIBXML_TEST_VERSION
  doc = xmlParseMemory (invalid_xml, strlen(invalid_xml));
  xmlFreeDoc (doc);
  xmlCleanupParser();

  return 0;
}
```

Without the patch it should randomly segfault if you run it a couple of times with backtraces like:

```
#0  0x00000000c9734006 in ?? ()
#1  0x00007ffff7e326f2 in rxkb_log () from /lib64/libxkbregistry.so.0
#2  0x00007ffff7e32b3a in xml_error_func () from /lib64/libxkbregistry.so.0
#3  0x00007ffff7e6b61f in xmlReportError () from /lib64/libxml2.so.2
#4  0x00007ffff7e6d4b1 in __xmlRaiseError () from /lib64/libxml2.so.2
#5  0x00007ffff7e71041 in xmlFatalErrMsgStrIntStr.lto_priv.0 () from /lib64/libxml2.so.2
#6  0x00007ffff7e82227 in xmlParseElementStart () from /lib64/libxml2.so.2
#7  0x00007ffff7e83e45 in xmlParseElement () from /lib64/libxml2.so.2
#8  0x00007ffff7e84ab0 in xmlParseDocument () from /lib64/libxml2.so.2
#9  0x00007ffff7e8503c in xmlSAXParseMemoryWithData () from /lib64/libxml2.so.2
#10 0x00000000004011fe in main ()
```

wismill:
Thanks for the excellent work! I completed with comments, a test and a changelog entry.

sbstnk:
Thanks for the quick review, improvements and merge! Also glad that you noticed that I missed an early return, that I somehow kept reading as another goto :)

PR #531, opened 2024-10-13
Registry: use safe libxml2 API

wismill:
See: #529 and the related comment in the [original Gnome issue](https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/7974#note_2245926).

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -383,6 +383,13 @@ endif
 # libxkbregistry
 if get_option('enable-xkbregistry')
     dep_libxml = dependency('libxml-2.0')
+    if cc.has_header_symbol(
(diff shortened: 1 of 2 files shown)


wismill:
@nwellnhof I do not see how to get rid of `xmlSetGenericErrorFunc` when using `xmlIOParseDTD` and `xmlValidateDtd`.

nwellnhof:
Yes, we need additional function for parsing and validating against external DTDs that take a parser context. I just created a libxml2 issue: https://gitlab.gnome.org/GNOME/libxml2/-/issues/808

wismill:
Tested with `libxml2-master`.

PR #533, opened 2024-10-30
Add typing annotations to Python scripts

wismill:


scripts/update-unicode.py
```diff
@@ -1228,7 +1228,7 @@ def stats(self, int_size) -> Stats:
             offsets2_int_size=0,
         )
 
-    @classmethod
+    @staticmethod
     def test(cls):
```
whot, line 1232: huh, I thought `staticmethod` doesn't take a `cls` argument, does it? If there's any callers they should now fail with "too few arguments"?

scripts/perfect_hash.py
```diff
@@ -384,7 +391,12 @@ def __call__(self, data, quote=False):
         return "\n".join(l.rstrip() for l in aux.getvalue().split("\n"))
 
 
-def generate_code(keys, Hash=StrSaltHash, template=None, options=None):
+def generate_code(
+    keys: list[str],
+    Hash: type = StrSaltHash,
```
whot, line 396: sigh, github doesn't let me do suggestions here:
shoudl this be 

```
Hash: type[H] = StrSaltHash,
```
for consistency with `generate_hash()`

whot (commented):
belated :+1: but the `staticmethod` bit looks wrong

scripts/update-unicode.py
wismill: You are right. Fixed in #534

scripts/perfect_hash.py
wismill: Fixed in #534

PR #534, opened 2024-11-03
Fix types annotations of Python scripts

wismill:
Follow-up of #533


wismill:
Mypy checks (apart from `ctypes` functions), so I guess we’re good. Will merge this week, unless there are issues.

PR #536, opened 2024-11-14
Keymap compiler server

wismill:
Create a keymap compiler *server*:
- Cache keymap components right after parsing (`XkbFile`)
- Propose a thread-safe atom table variant
- Use Unix sockets to serve keymap compiler service

**This speeds up the xkeyboard-config test by about _5x_**.

Unfortunately, this adds a fair amount of code, in particular in `ast-build.c`. I am not sure we are willing to commit to such addition for only this particular use case. Given the include mechanism and the interaction between components, I do not see how to cache things in a better/easier way. Maybe mmaping all files beforehand could save us costly storage reads.

Anyway, this was a good experience for me to understand better our source code, experiment with thread-safe structures and Unix sockets in C.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -512,6 +512,19 @@ if build_tools
                c_args: ['-DENABLE_PRIVATE_APIS'],
                include_directories: [include_directories('src', 'include')],
                install: false)
+    thread_dep = dependency('threads', required: false)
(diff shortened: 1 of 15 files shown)


wismill:
@whot @bluetech any suggestion for profiling our code base?

PR #542, opened 2024-11-25
changelog: Fix release title format

wismill:
Towncrier doc states that the `underlines` setting is not used with Markdown files, so let’s apply the tag ourselves.


wismill:
> There was a problem generating the rebase commit

Let’s wait a bit 📖

PR #551, opened 2024-12-09
keysyms: Add sharp S upper case mapping exception

### LINKED ISSUE #550: ß should uppercase to ẞ
The German letter ß (`XKB_KEY_ssharp`) never appears in the middle of words; therefore, it is only capitalized in all-caps contexts. Traditionally, there was no widely accepted capital form of the letter, so SS was used instead. However, such a form was recently introduced (ẞ, `XKB_KEY_Ssharp`), and, as of this year, is now the primary recommended form (with SS still permitted as an alternative). Unicode’s stability guarantees mean that it cannot adopt the new mapping for its default casing rules. But `xkbcommon` should not be restricted by this, and should designate ß and ẞ as a case pair.

(Related discussion thread on Unicode’s mailing list: https://corp.unicode.org/pipermail/unicode/2024-November/011162.html)

wismill:
The case mapping `ssharp` ß ↔ `U1E9E` ẞ was added in 13b30f4f0dccc08dfea426d73570b913596ed602 but was broken:
- For the lower case mapping it returned the keysym `0x10000df`, which is an invalid Unicode keysym.
- For the upper case mapping it returned the upper Unicode code point rather than the corresponding keysym.

It did accidentally enable the detection of alphabetic key type for the pair (ß, ẞ) though. However this detection was accidentally removed in 5c7c79970a2800b6248e829464676e1f09c5f43d (v1.7) with an attempt to fix the wrong keysym case mapping. Finally both the *lower* case mapping and the key type detection were fixed for good when we implemented the complete Unicode simple case mappings and corresponding tests in e83d08ddbc9851944c662c18e86d4eb0eff23e68.

However, the *upper* case mapping `ssharp` → `U1E9E` remained disabled. Indeed, ẞ is a relatively recent addition to Unicode (2008) and had no official recommendation, until recently. So while the lower mapping ẞ→ß exists in Unicode, its converse upper mapping does not. Yet since 2017 the Council for German Orthography (Rat für deutsche Rechtschreibung) recommends[^1] ẞ as the capitalization of ß.

Due to its stability policies, the Unicode Character Database (UCD) that we use to generate our keysym case mapping (via ICU) cannot update the simple case mapping of ß. Discussions are currently ongoing in the Unicode mailing list[^2] and CLDR[^3] about how to deal with the new recommended case mapping. However, the discussions are oriented on text-processing and compatibility mappings, while libxkbcommon is on a rather lower level.

It seems that the slow adoption of ẞ is partly due to the difficulty to type it. Since ẞ is used only for ALL CAPS casing, the expectation is to type it using CapsLock. While our detection of alphabetic key types works well[^4] for the pair (ß,ẞ), the *internal capitalization* currently does not work and is fixed by this commit.

Added the ß → ẞ upper mapping:
- Added an exception in the generation script
- Fixed tests
- Added documentation of the exceptions in `xkbcommon.h`
- Added/updated log entries

[^1]: https://www.rechtschreibrat.com/regeln-und-woerterverzeichnis/
[^2]: https://corp.unicode.org/pipermail/unicode/2024-November/011162.html
[^3]: https://unicode-org.atlassian.net/browse/CLDR-17624
[^4]: Except libxkbcommon 1.7, see the second paragraph.

Fixes #550


wismill:
It took me a while to properly remember and assess the consequences of 13b30f4f0dccc08dfea426d73570b913596ed602 and 5c7c79970a2800b6248e829464676e1f09c5f43d. I hope the commit message provides now sufficiently information.

wismill:
FYI: I filed [a fix for libX11](https://gitlab.freedesktop.org/xorg/lib/libx11/-/merge_requests/274) too.

PR #567, opened 2024-12-19
symbols: Parse empty key

### LINKED ISSUE #565: xkbcomp: Some lists are not allowed to be empty
(Follow-up of [this comment](https://github.com/xkbcommon/libxkbcommon/issues/209#issuecomment-2553080832)

The following syntax does not parse in xkbcommon, but it does in `xkbcomp`:

```c
xkb_symbols "x" {
    key <AC12> { };
};
```

While the usefulness of such statement is debatable, the fact that it does parse in `xkbcomp` and that tools may generate such keymap entry make it relevant to handle.

There are other places we could allow empty lists as well.

CC @mahkoh who detected this issue

wismill:
The following syntax does not parse in xkbcommon, but it does in xkbcomp:

```
xkb_symbols "x" {
    key <AD01> { };
};
```

While the usefulness of such statement is debatable, the fact that it does parse in xkbcomp and that tools may generate such keymap entry make it relevant to handle.

See: #565


wismill:
Test will be added in #568.

PR #568, opened 2024-12-19
xkbcomp: Fix key merge mode (keysyms, actions)

### LINKED ISSUE #564: Abort when augmenting key definition
The following code triggers an assertion:

```
    xkb_symbols {
        key <a> { [ A ] };
        augment key <a> { [ { A, A } ], [ { SetMods(), SetMods() } ] };
    };
```

compile-keymap: ../src/xkbcomp/symbols.c:361: MergeGroups: Assertion `intoLevel->num_syms == fromLevel->num_syms' failed.

wismill:
This also adds exhaustive tests for merge modes.

Fixes #564

TODO
- [x] Remove first commit once #567 is merged

TODO in follow-up MRs:
- Merge mode propagation
- Other fields/objects

@mahkoh


wismill:
Windows [Compiler Error C2026](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2026?view=msvc-170). Wonderful.

Fixed by using `assert_printf` instead of `assert`. We could also use a temp var.

wismill:
Note to self: in order to check other key properties, we should probably just compare resulting keymap strings with a ref.

Also let's add the same tests for `xkbcomp` (at least where the syntax is supported: e.g. no multi keysym per level).

bam80:
May be related: #573

wismill:
Trying to fix the failing macOS pipeline. I suspect an out-of-bound in xkbcomp (see [this MR](https://gitlab.freedesktop.org/xorg/app/xkbcomp/-/merge_requests/26)) to be the origin of:

```
Warning:          Multiple actions for level 3/group 1 on key <AD03>
                  Using Private, ignoring SetGroup
```

The `Private` action bit above is suspicious.

---

The Linux pipeline is fragile too: it works without Valgrind, so I suspect race conditions with the various threads.

wismill:
I finally managed to install the [patched xkbcomp](https://gitlab.freedesktop.org/xorg/app/xkbcomp/-/merge_requests/26) on the macOS CI. Linux CI seems robust now.

wismill:
@mahkoh With this MR we will have a first round of tests for the common uses of merge modes multiple fixes.

TODO:
- Merge mode propagation
- Other fields/objects

wismill:
Added tests for local merge mode on the keys, in addition to the previous global merge mode tests. Both local/global merge modes should lead to the same keymap, but there is some slight discrepancy on a couple of cases.

mahkoh:
I've compared the test cases and the only significant differences between this and my implementation are as follows:

- The symbols in a level are either all applied or none are applied. `{ a, NoSymbol }` and `{ NoSymbol, b }` do not merge to `{ a, b }`.
- Symbols and actions are merged independently with no requirement that they have the same cardinality.

wismill:
>     * The symbols in a level are either all applied or none are applied. `{ a, NoSymbol }` and `{ NoSymbol, b }` do not merge to `{ a, b }`.

@mahkoh  It means that with your implementation one cannot *partially* update a keysym list for one level, correct? Why not allowing it?

>     * Symbols and actions are merged independently with no requirement that they have the same cardinality.

Usually the actions mirror the keysyms; is it not counter-intuitive to handle them completely separately? Multiple actions per levels are pretty new, but the idea was to have keysyms count = actions count, to make it more intuitive. Do you have a use case for different keysyms and actions count?

mahkoh:
>@mahkoh It means that with your implementation one cannot partially update a keysym list for one level, correct? Why not allowing it?

I'm already a bit concerned about allowing symbols and actions of the same level to be updated independently. For example, consider

```
key <a> { [ a ] };
augment key <a> { [ Shift_L ], [ SetMods(mods = Shift) ] };
```

This produces

```
key <a> { [ a ], [ SetMods(mods = Shift) ] };
```

which is a bit confusing. Ideally both the symbols and actions of a single level would remain consistent. But since xkbcomp updates them separately, I went with that.

Since xkbcomp doesn't handle multiple objects per level, I decided to be a bit more conservative here.

>Usually the actions mirror the keysyms; is it not counter-intuitive to handle them completely separately? Multiple actions per levels are pretty new, but the idea was to have keysyms count = actions count, to make it more intuitive. Do you have a use case for different keysyms and actions count?

I assume that this notion comes from the `interpret` declarations which connect keysyms and actions. As long as there is this close relationship, it makes sense for things to be symmetric.

However, I believe that this whole `interpret` stuff is unnecessary complexity. When reading a keymap, I very much prefer the key declarations to contain their actions directly instead of going via the `interpret` indirection. This makes it much easier to se what is happening when a key is pressed. And once you sever this connection, I don't believe that there is any reason for there to be any symmetry between symbols and actions.

In particular, when formatting keymaps, I always list all actions explicitly and only emit a single `interpret` statement to work around https://gitlab.freedesktop.org/xorg/lib/libxkbfile/-/issues/12.

PR #585, opened 2025-01-10
Use test map for checking --version-script, instead of full map

DimitryAndric:
Various problems can occur when using the full `xkbcommon.map` file when checking whether the `--version-script` linker option works in `meson.build`. See https://lists.freebsd.org/archives/freebsd-current/2025-January/006888.html for an example.

For instance, newer linkers typically complain about non-existing symbols, so in commit ebe4157 `--undefined-version` was added to work around that. But then in commit 1d8a25d6 another workaround needed to be added for older linkers.

It is better to use a minimalistic linker script for testing instead. The other workarounds can then be removed.

### DIFF
--- a/changes/build/481.bugfix.md
+++ b/changes/build/481.bugfix.md
@@ -1,2 +1,1 @@
-Gated the use of `--undefined-version` in the linker because it breaks on older
-GNU `ld`.
+Make the test of `-Wl,--version-script` more robust.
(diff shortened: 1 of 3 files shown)


DimitryAndric:
Btw, if anybody knows a way to have `meson.build` write the `test.map` file by itself, that would probably be slightly neater.

DimitryAndric:
Thanks!

wismill:
Thank you for you contribution! 👏

PR #597, opened 2025-01-16
Fix integer literal suffixes

wismill:
Fix a few of them and make the rest portable using `UINTn_C` macros.

@whot @bluetech

### DIFF
--- a/src/atom.c
+++ b/src/atom.c
@@ -85,7 +85,7 @@
 static inline uint32_t
 hash_buf(const char *string, size_t len)
 {
-    uint32_t hash = 2166136261u;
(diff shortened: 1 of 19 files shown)


bluetech:
First time I'm seeing these stdint.h `UINT32_C` macros. Just for my curiosity can you explain their purpose? Wasn't able to figure it out quickly.

wismill:
@bluetech  They create literals with the proper suffix, e.g. `u` or `ul`, depending of the platform, to get a >=32bit int literal. We could use `1ul` everywhere and forget about it, but it is more correct with these macros. I did forgot a `l` in some places.

See: https://en.cppreference.com/w/c/types/integer.

PR #599, opened 2025-01-20
ci: Enable sanitizers

wismill:


wismill:
@whot @bluetech It seems using ELF sections is not compatible with `asan` (at least not with our current code). Any idea?

src/xkbcomp/keymap-dump.c
```diff
@@ -88,7 +88,8 @@ check_write_buf(struct buf *buf, const char *fmt, ...)
 
     available = buf->alloc - buf->size;
     va_start(args, fmt);
-    printed = vsnprintf(buf->buf + buf->size, available, fmt, args);
+    printed = vsnprintf(buf->buf ? buf->buf + buf->size : buf->buf, available,
```
bluetech, line 91: Is `buf->buf` really meant to be NULL here?

bluetech:
> It seems using ELF sections is not compatible with asan (at least not with our current code). Any idea?

I think using ELF sections is a too clever for a portable library like xkbcommon. I think we can manage without them.

> And I cannot use asan on our Linux pipeline because it does not play well with Valgrind.

I think asan and Valgrind mostly overlap, but if you want both, we can run separate jobs for them. The jobs are quick so I'm not worried about that. That said, if asan is running in macos, maybe that's sufficient coverage.

src/xkbcomp/keymap-dump.c
wismill: It can if `available` is 0. We try to write or fail with the required size used later to reallocate. I added a comment.

wismill:
@bluetech I would prefer to leave the ELF machinery untouched if the only harm is on macOS with asan. Anyway we have more issues with macOS (see #591 that require a patched `xkbcomp` due to a [bug](https://gitlab.freedesktop.org/xorg/app/xkbcomp/-/merge_requests/26)).

As for the Valrind/asan overlap, Valgrind did not caught the issue in the second commit, which fixes an *undefined behaviour*. So it may be worth to use both separately.

src/xkbcomp/keymap-dump.c
```diff
@@ -87,16 +87,24 @@ check_write_buf(struct buf *buf, const char *fmt, ...)
...
+    if (printed < 0) {
         goto err;
-
-    if ((size_t) printed >= available)
-        if (!do_realloc(buf, printed))
+    } else if ((size_t) printed >= available) {
+        if (!do_realloc(buf, printed + 1))
             goto err;
+    } else {
+        buf->size += printed;
```
bluetech, line 105: Good catch, we were calling vsnprintf again even if it wasn't needed.

bluetech (approved):
The code changes LGTM

bluetech:
> As for the Valrind/asan overlap, Valgrind did not caught the issue in the second commit, which fixes an undefined behaviour. So it may be worth to use both separately.

Ah, that's ubsan, not asan. But probably they are both incompatible with Valgrind.

src/xkbcomp/keymap-dump.c
wismill: It’s your previous comment that trigger my investigation. Love team work!

wismill:
I could not get the sanitizers run on Linux with clang ([linking errors](https://github.com/xkbcommon/libxkbcommon/actions/runs/12868960840/job/35876709055?pr=599) nor on macOS ([error due to the use of `__DATA` section](https://github.com/xkbcommon/libxkbcommon/actions/runs/12869364920/job/35877994007) in the X11 tests. Any idea?

bluetech:
> I could not get the sanitizers run on Linux with clang (linking errors nor on macOS (error due to the use of __DATA section in the X11 tests. Any idea?

Looks like it's incompatible with `b_lundef=true`. Not sure if it's a meson bug or intended.

bluetech:
(I meant for linux clang, I don't know about macos)

eli-schwartz:
> Looks like it's incompatible with `b_lundef=true`. Not sure if it's a meson bug or intended.

Meson prints a warning: https://github.com/xkbcommon/libxkbcommon/actions/runs/12868960840/job/35876709055?pr=599#step:6:88

```
meson.build:1132: WARNING: Trying to use address,undefined sanitizer on Clang with b_lundef.
This will probably not work.
Try setting b_lundef to false instead.
```

It is a clang design issue. Works fine with GCC.

PR #600, opened 2025-01-21
Fix misc compiler warnings

bluetech:
Small changes fixing a UBSAN error and some compiler warnings.

Sorry for lazily putting it all in a single PR, I can split it if you prefer.

### DIFF
--- a/src/xkbcomp/parser.y
+++ b/src/xkbcomp/parser.y
@@ -255,7 +255,7 @@ resolve_keysym(struct parser_param *param, const char *name, xkb_keysym_t *sym_r
  */
 
 XkbFile         :       XkbCompositeMap
-                        { $$ = param->rtrn = $1; param->more_maps = !!param->rtrn; }
+                        { $$ = param->rtrn = $1; param->more_maps = !!param->rtrn; (void) yynerrs; }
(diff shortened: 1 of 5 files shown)


bluetech:
I might go over the msvc warnings and try to fix them later (in a separate PR).

src/registry.c
wismill, line 1: This does not look right. I think we should remove the `destroy` field and generate `*_unref` that call the proper `*_destroy` function directly.

src/xkbcomp/parser.y
wismill, line 1: FYI: An upstream fix [is pending](https://github.com/akimd/bison/issues/103), but [Bison is marked as needing a comaintainer](https://www.gnu.org/server/takeaction.html#unmaint).

wismill (changes requested):
LGTM, but the registry seems to require a proper refactor. I can have a look at it in another PR.

src/registry.c
wismill: See #601

wismill: #601 has been merge, you may drop the corresponding commit.

bluetech: Rebased

PR #601, opened 2025-01-21
registry: Refactor to avoid hazardous casts

wismill:
Inline `rxkb_object_unref` for each type and remove the `destroy` field.


bluetech (approved):
This works, thanks!

PR #602, opened 2025-01-21
Fix warnings about empty variadic macros

wismill:
Avoid using the GNU extension for `, ##`.

I got a flood of warnings with `-Wpedantic`. 

Note that if we do not merge this, we should remove the following comments in the code:

> The format is not part of the argument list in order to avoid the "ISO C99 requires rest arguments to be used" warning when only the format is supplied without arguments.


bluetech:
I don't think we should do this, it makes the code worse and I don't think it improves portability in any meaningful way.

Issue #131, opened 2019-12-28
MSVC compiler warnings


bluetech:
These are warnings the Windows CI displays. Some of these should be looked into. Mostly overflow and truncation potential due to casts. The enum casts can possibly lead to unaligned access.

```
../src/xkbcomp/parser.y(484): warning C4244: 'function': conversion from 'int64_t' to 'unsigned int', possible loss of data
../src/xkbcomp/parser.y(496): warning C4244: 'function': conversion from 'int64_t' to 'unsigned int', possible loss of data
../src/xkbcomp/parser.y(498): warning C4244: 'function': conversion from 'int64_t' to 'unsigned int', possible loss of data
../src/xkbcomp/parser.y(702): warning C4244: 'function': conversion from 'int64_t' to 'int', possible loss of data
d:\a\1\s\src\compose\parser.c(428): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
d:\a\1\s\src\xkbcomp\action.c(175): warning C4133: 'function': incompatible types - from 'xkb_action_type *' to 'unsigned int *'
d:\a\1\s\src\xkbcomp\action.c(181): warning C4133: 'function': incompatible types - from 'action_field *' to 'unsigned int *'
d:\a\1\s\src\xkbcomp\action.c(302): warning C4133: 'function': incompatible types - from 'xkb_action_flags *' to 'unsigned int *'
d:\a\1\s\src\xkbcomp\action.c(632): warning C4133: 'function': incompatible types - from 'xkb_action_controls *' to 'unsigned int *'
d:\a\1\s\src\xkbcomp\compat.c(269): warning C4133: 'function': incompatible types - from 'xkb_match_operation *' to 'unsigned int *'
d:\a\1\s\src\xkbcomp\rules.c(312): warning C4267: '=': conversion from 'size_t' to 'unsigned int', possible loss of data
d:\a\1\s\src\xkbcomp\scanner.c(55): warning C4244: '=': conversion from 'double' to 'int64_t', possible loss of data
d:\a\1\s\test\log.c(64): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
d:\a\1\s\test\log.c(66): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
src/xkbcomp/keywords.gperf(95): warning C4267: 'initializing': conversion from 'size_t' to 'unsigned int', possible loss of data
```

PR #604, opened 2025-01-21
Fix some issues with implicit int casts

wismill:
Most notably, the cast for enum *pointers* mentioned in #131.

### DIFF
--- a/src/xkbcomp/action.c
+++ b/src/xkbcomp/action.c
@@ -170,15 +170,21 @@ static const LookupEntry fieldStrings[] = {
 };
 
 static bool
-stringToAction(const char *str, enum xkb_action_type *type_rtrn)
+stringToActionType(const char *str, enum xkb_action_type *type_rtrn)
 {
-    return LookupString(actionTypeNames, str, type_rtrn);
+    unsigned int type = 0;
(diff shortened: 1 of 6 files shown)


src/xkbcomp/action.c
```diff
@@ -169,16 +171,22 @@ static const LookupEntry fieldStrings[] = {
     { NULL,               0                          }
 };
 
-static bool
```
bluetech, line 172: The previous signature is less likely to be misused, so I prefer it. But no hard objection.

bluetech (approved):
Thanks!

src/xkbcomp/action.c
wismill: Agreed, but when I tried to rewrite it to fix the enum pointer cast it somehow felt off and inefficient. Since these 2 functions are only called twice, maybe we could just inline them?

bluetech: Type-safe wrappers seem like a good thing to me. Personally I'd add the bit of boilerplate but up to you :)

wismill: Let’s do your way then.

bluetech (approved):

PR #605, opened 2025-01-21
Fix numbers types

### LINKED ISSUE #594: Virtual modifier explicit mapping is limited to platform-dependent int size
_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/issues/450#issuecomment-2576088237_
            
              @mahkoh Good point, setting `virtual_modifiers` using numeric values is also possible, although this is quite an obscure setting IMHO.

There is a limitation though: the numeric values `ExprInteger::ival` are `int` and depending on the target compiler/platform it may not be big enough to hold the mask. Agreed, nowadays on most cases we should have `sizeof(int) * CHAR_BIT >= 32`, but the C standard guarantee only 16 bits. Such overflow would be completely silent and difficult to debug. Note that overflow is still possible and silent with the current implementation with 32-bits ints: e.g. `virtual_modifiers PureVirtual = 0xffffffff + 1`;

So we may want to switch `ExprInteger::ival` from `int` to the explicit `int32_t`, which should be plenty enough.

wismill:
WIP, made public only to make you aware I am working on this.

Follow-up of #604 that is mostly a refactor of the parser for integers. Will probably split the minor refactoring in another PR.

Fixes #594

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -246,13 +246,12 @@ log_level(const char *level) {
 
 static int
 log_verbosity(const char *verbosity) {
-    char *endptr;
-    int v;
-
     errno = 0;
-    v = strtol(verbosity, &endptr, 10);
-    if (errno == 0)
-        return v;
+    const long int v = strtol(verbosity, NULL, 10);
+    if (errno == 0) {
+        /* Truncate */
+        return (int) v;
+    }
 
     return 0;
 }
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -8,6 +8,9 @@
 
 #include "config.h"
 
+#include <assert.h>
+#include <stdint.h>
+
 #include "xkbcommon/xkbcommon-names.h"
 #include "keymap.h"
 
@@ -174,10 +177,12 @@ XkbWrapGroupIntoRange(int32_t group,
          * C99 says a negative dividend in a modulo operation always
          * gives a negative result.
          */
-        if (group < 0)
-            return ((int) num_groups + (group % (int) num_groups));
-        else
+        if (group < 0) {
+            static_assert(XKB_MAX_GROUPS < INT32_MAX, "Max groups don't fit");
+            return ((int32_t) num_groups + (group % (int32_t) num_groups));
+        } else {
             return group % num_groups;
+        }
     }
 }
 
--- a/src/keymap.c
+++ b/src/keymap.c
@@ -13,6 +13,9 @@
 
 #include "config.h"
 
+#include <assert.h>
+#include <stdint.h>
+
 #include "keymap.h"
 #include "text.h"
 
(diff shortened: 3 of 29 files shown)


bluetech:
Some possible followups:
- In xkbcomp/scanner we should change the `strtoul` to `strtol` since we're using `int64_t`. While integer literal cannot be negative, the cast of `unsigned long` to `int64_t` is lossy. (We could change `ival` to unsigned but then it just defers the cast so I think we shouldn't).
- We should probably check the add/sub/mul in expr.c for overflows/underflows and error if it happens. C23 has stdckdint.h, maybe we can use some backport of it.

I will try to do this unless you feel like it :)

wismill:
> * In xkbcomp/scanner we should change the `strtoul` to `strtol` since we're using `int64_t`. While integer literal cannot be negative, the cast of `unsigned long` to `int64_t` is lossy. (We could change `ival` to unsigned but then it just defers the cast so I think we shouldn't).

@bluetech I agree, it showed multiple times already with `clang-tidy`. But it’s not ideal, as `strtol` returns a `long` but we want an `int64_t`.

> * We should probably check the add/sub/mul in expr.c for overflows/underflows and error if it happens. C23 has stdckdint.h, maybe we can use some backport of it.

While I agree that would make our implementation more robust:
1. I think I have never seen those in real use.
2. We already use 64 bits ints, although all our ints are max 32 bits. So we already have some great margin.

Do you have a realistic use case in mind?

bluetech:
> @bluetech I agree, it showed multiple times already with clang-tidy. But it’s not ideal, as strtol returns a long but we want an int64_t.

We can use strtoll then I suppose.

> I think I have never seen those in real use.

It's new in C23 so probably no one uses it yet, instead using gcc builtins or custom code.

> We already use 64 bits ints, although all our ints are max 32 bits. So we already have some great margin.

Do you mean before or after this PR? I think that before, we are using int for both operands and result, so overflow is possible, and after, we are using int64_t for both operands and result, so overflow is still possible. Do I have it wrong?

bluetech (approved):
Looks great! This was very much needed.

wismill:
> > I think I have never seen those in real use.
> 
> It's new in C23 so probably no one uses it yet, instead using gcc builtins or custom code.

@bluetech I meant the int operations in XKB files, not the C23 functions.

> Do you mean before or after this PR? I think that before, we are using int for both operands and result, so overflow is possible, and after, we are using int64_t for both operands and result, so overflow is still possible. Do I have it wrong?

You are right, but what I mean is that we do not expect the user to *input* numbers > 32 bits. So after we will handle 32 bits correctly, with some margin to handle overflow and report it. But if the `int64_t` overflows, then indeed we would not catch it. But I find it very unlikely, see my comment about the use of operations before.

bluetech:
Oh yes, there is definitely no real use case for doing such things. And the arithmetic isn't even allowed in many places. But I just feel bad about allowing user input to trigger UB (signed int overflow). We could use `-fwrapv` but it's not portable.

wismill:
@bluetech I tackle the `strtoul` issue and fixed 2 more silent casts for compat group and LED index. Added some tests too.

bluetech (approved):
The extra changes LGTM

wismill:
Latch batch for this PR. There are still cast warnings, but I leave it for a further PR.

wismill:
Note to self: check the Windows CI logs!

PR #606, opened 2025-01-21
actions: Fix SwitchScreen screen value range

### LINKED ISSUE #571: Range of accepted screen values does not match
In `src/keymap.h`, the type of the `screen` member in the `xkb_switch_screen_action` struct is `int8_t`. This also matches the type specified in the X Keyboard Extension Protocol Specification for *SA_SwitchScreen*. However, during the validation step in `src/xkbcomp/action.c`, in the function `HandleSwitchScreen()`, all values in the range [-255, +255] are accepted.

Should the validation step instead check that the value is in [-128, +127], in order to fit inside the range of `int8_t`?

wismill:
Currently we accept values of -255..255 while parsing the screen value of `SwitchScreen` actions`, but then we silently cast it against `int8_t`, i.e. in range -128..127.

We actually do as xkbcomp, but this seems a bug because the target storage is a `char`. Let’s simply raise a parse error if the value does not fit in our type.

Fixes #571

CC @wysiwys

### DIFF
--- a/src/xkbcomp/action.c
+++ b/src/xkbcomp/action.c
@@ -600,16 +600,18 @@ HandleSwitchScreen(struct xkb_context *ctx, const struct xkb_mod_set *mods,
 
         if (!ExprResolveInteger(ctx, scrn, &val))
             return ReportMismatch(ctx, XKB_ERROR_WRONG_FIELD_TYPE, action->type,
-                                  field, "integer (0..255)");
+                                  field, "integer (-128..127)");
 
-        if (val < 0 || val > 255) {
+        val = (value->expr.op == EXPR_NEGATE ? -val : val);
+        if (val < INT8_MIN || val > INT8_MAX) {
             log_err(ctx, XKB_LOG_MESSAGE_NO_ID,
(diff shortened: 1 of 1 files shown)


src/xkbcomp/action.c
```diff
@@ -600,16 +600,18 @@ HandleSwitchScreen(struct xkb_context *ctx, const struct xkb_mod_set *mods,
...
-                    "Screen index must be in the range 1..255; "
-                    "Illegal screen value %d ignored\n", val);
+                    "Screen index must be in the range %d..%d; "
+                    "Illegal screen value %d ignored\n",
+                    INT8_MIN, INT8_MAX, val);
             return false;
         }
 
-        act->screen = (value->expr.op == EXPR_NEGATE ? -val : val);
+        act->screen = val;
```
bluetech, line 614: Would add an explicit cast to make it clear that we checked the int -> int8_t conversion. 

```suggestion
        act->screen = (int8_t) val;
```

bluetech (approved):

PR #607, opened 2025-01-21
meson: Make building benchmarks & fuzzers optional

wismill:
I do not think there is a reason to make everybody build those *by default*.


bluetech:
This doesn't save a lot of time, while on the other end introduces a small risk of hiding problems, and adds a bit of complexity. This is why I never added these options.

wismill:
Fair enough

PR #608, opened 2025-01-22
symbols: Fix 2 possible null pointer deferences

wismill:
Detected using `clang-tidy`. We probably need to activate some automatic fuzzer to ensure there are no other hazardous code paths!

### DIFF
--- a/test/buffercomp.c
+++ b/test/buffercomp.c
@@ -285,6 +285,20 @@ test_invalid_symbols_fields(struct xkb_context *ctx)
         "    xkb_types { };\n"
         "    xkb_compat { };\n"
         "    xkb_symbols { key <> { groupsredirect = [] }; };\n"
+        "};",
+        /* Used to parse without error because the “repeats” key field is valid,
+         * but we should fail in the following 2 keymaps */
+        "xkb_keymap {\n"
+        "    xkb_keycodes { <> = 9; };\n"
+        "    xkb_types { };\n"
+        "    xkb_compat { };\n"
+        "    xkb_symbols { key <> { vmods=[], repeats=false }; };\n"
+        "};",
+        "xkb_keymap {\n"
+        "    xkb_keycodes { <> = 9; };\n"
+        "    xkb_types { };\n"
+        "    xkb_compat { };\n"
+        "    xkb_symbols { key <> { repeats=false, vmods=[] }; };\n"
         "};"
     };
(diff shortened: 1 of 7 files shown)


test/buffercomp.c
```diff
@@ -250,6 +250,50 @@ test_multi_keysyms_actions(struct xkb_context *ctx)
...
+test_invalid_symbols_fields(struct xkb_context *ctx)
+{
+    /* Any of the following is invalid syntax, but also use to trigger
+     * a NULL pointer deference, thus this regression test */
+    const char * const keymaps[] = {
+        "xkb_keymap {\n"
+        "    xkb_keycodes { <> = 9; };\n"
+        "    xkb_types { };\n"
+        "    xkb_compat { };\n"
+        "    xkb_symbols { key <> { vmods = [] }; };\n"
```
bluetech, line 263: Is `[]` the only way this NULL can happen? The grammar rule is

```y
ArrayInit       :       OBRACKET MultiKeySymList CBRACKET
                        { $$ = $2; }
                |       OBRACKET MultiActionList CBRACKET
                        { $$ = $2; }
                |       OBRACKET CBRACKET
                        { $$ = NULL; }
```

The `OBRACKET CBRACKET` (i.e. `[]`) ends up as NULL, but what if we turn it to an empty `ExprKeysymList` or `ExprActionList` so we don't have to deal with yucky NULL? From my reading of `SetSymbolsField` it will result in a clean "type error" in all cases. And perhaps will allow to remove the `value == NULL` case from `AddSymbolsToKey`.

wismill: I introduced `EXPR_EMPTY_LIST`, so we have clear semantics: we do not end to check for empty keysym list when processing actions. The first 2 commit requires to be squashed.

I also detected another issue with error handling, see the corresponding commit.

wismill:
Thanks to a typo, I uncovered another possible memory leak. There was also an inconsistent error handling in symbols.

test/buffercomp.c
bluetech: > we do not end to check for empty keysym list when processing actions

Right, bummer.

Having a special expression type for empty array is pretty funny but the keysym/action ambiguity forces it.

bluetech (approved):

wismill:
Added a comment and rework the commit message to explain the motivation of `EXPR_EMPTY_LIST`.

PR #609, opened 2025-01-22
Clang tidy

wismill:


src/compose/parser.c
```diff
@@ -734,10 +734,10 @@ lhs_mod_list_tok: {
                          "right-hand side string is too long: "
                          "expected max: %d, got: %d; skipping line",
                          (int)sizeof(production.string) - 1,
-                         (int)val.string.len);
+                         (int)val.string.len - 1);
             goto skip;
         }
-        strcpy(production.string, val.string.str);
+        strncpy(production.string, val.string.str, val.string.len);
```
bluetech, line 740: strncpy does useless padding of the destination string. And if it's not error checked then it's still not good, unless silent truncation is really the desired outcome. IMO it's a bad function so I'm not a huge fan of this commit.

strcpy is safe here since we check the sizes, but if want to avoid it then memcpy would be better.

src/text.c
```diff
@@ -273,7 +273,9 @@ ModMaskText(struct xkb_context *ctx, const struct xkb_mod_set *mods,
     if (unlikely(mask & ~((UINT64_C(1) << mods->num_mods) - 1))) {
         /* If we get a mask that cannot be expressed with the known modifiers,
          * print it as hexadecimal */
-        snprintf(buf + pos, sizeof(buf) - pos, "0x%"PRIx32, mask);
+        const int ret = snprintf(buf, sizeof(buf), "0x%"PRIx32, mask);
+        if (ret >= 0 && (size_t) ret < sizeof(buf))
```
bluetech, line 277: Would change to an assert since we are not handling failure here. If the `else` ever happens, it's no good.

src/x11/keymap.c
```diff
@@ -832,6 +832,8 @@ get_sym_interprets(struct xkb_keymap *keymap, xcb_connection_t *conn,
         case XCB_XKB_SYM_INTERPRET_MATCH_EXACTLY:
             sym_interpret->match = MATCH_EXACTLY;
             break;
+        default:
+            break;
```
bluetech, line 838: Hmm, leaving out the `default` was sort of an "this can't happen", but if we add it, it looks legit, but we can't leave `sym_interpret->match` uninitialized though. It needs to either error, assert or fallback.

tools/interactive-x11.c
```diff
@@ -253,6 +253,8 @@ process_xkb_event(xcb_generic_event_t *gevent, struct keyboard *kbd)
                               event->state_notify.latchedGroup,
                               event->state_notify.lockedGroup);
         break;
+    default:
```
bluetech, line 256: Add empty line above to keep the style.

src/keymap.h
```diff
@@ -141,6 +141,7 @@ enum xkb_action_type {
 };
 
 enum xkb_action_flags {
+    ACTION_FLAG_NONE = 0,
```
bluetech, line 144: To me, `0` for flag enums seems clearer than `NONE`. What is the tidy rule which wants this?

bluetech (commented):
Nice changes! I left comments on a few things, but everything else LGTM.

src/text.c
wismill: Fix with assertions

src/keymap.h
wismill: > What is the tidy rule which wants this?

It’s:

> warning: switching on non-enum value without default case may not cover all cases [bugprone-switch-missing-default-case]

Let’s add a default case and call it a day.

src/x11/keymap.c
```diff
@@ -338,9 +338,9 @@ translate_action(union xkb_action *action, const xcb_xkb_action_t *wire)
 
         /* Treat high unknown actions as Private actions. */
         action->priv.type = wire->noaction.type;
-        STATIC_ASSERT(sizeof(action->priv.data) == 7 &&
-                      sizeof(wire->noaction.pad0) == 7,
-                      "The private action data must be 7 bytes long!");
+        _Static_assert(sizeof(action->priv.data) == 7 &&
```
bluetech, line 341: Can we use `static_assert` instead of the uglier `_Static_assert`? It should be defined in `<assert.h>` according to [here](https://en.cppreference.com/w/c/error/static_assert).

bluetech (approved):

src/x11/keymap.c
wismill: Fixed. I missed that although not a keyword in C11, we do have this convenience macro.

PR #614, opened 2025-01-24
xkbcomp: rework ActionList and KeysymList AST representation

bluetech:
The AST is heavily based on intrusive lists for representing lists, but actions and keysyms are an exception, instead using darray. I don't see any reason for this; it ends up allocating more, and we don't especially need a flat array for this.

Change it to use the familiar linked-list style.

bench/rulescomp allocations, before:

```
total heap usage: 11,391,536 allocs, 11,391,536 frees, 511,335,213 bytes allocated
```

after:

```
total heap usage: 10,965,536 allocs, 10,965,536 frees, 496,327,213 bytes allocated
```

PR #615, opened 2025-01-25
build: link tests and benches against shared library, not static library

bluetech:
This makes the tests, and especially benches, more realistic, since xkbcommon is almost always used as a shared library.

Also significantly reduces the build time with LTO enabled (for me, from 90s to 30s).


bluetech:
With some effort I got windows to work, unfortunately it required some additional commits.

wismill:
Nice work!

Could you add a milestone to your PRs, if relevant? We are close to 1.8 release. Unfortunately, I also found myself keeping adding MRs after discovering bugs or limitations, but we ought to do the release soon, else we may miss some distro packages freeze that will not ship important bug fixes.

bluetech:
Sure -- I added this and #614 to the 1.9.0 milestone.

bluetech:
Looks like `S_ISFIFO` is not available on Windows. There's also a new warning on `execv`. Not sure how it compiled before. Will look into it.

wismill (approved):
LGTM, I have trust you on this one.

PR #616, opened 2025-01-25
[POC] xkbcomp: allocate AST nodes using bump allocator

bluetech:
Note: only last 2 commits are relevant.

This is a proof of concept I quickly hacked together to switch the AST allocation to use a bump allocator. This means we just allocate AST nodes in big chunks and then free it all together once done.

This is just a proof of concept, will need more work if we want to merge it. But enough to benchmark:

bench/rulescomp
before:
total heap usage: 10,980,536 allocs, 10,980,536 frees, 490,563,213 bytes allocated
compiled 1000 keymaps in 2.300165s

after:
total heap usage: 4,647,536 allocs, 4,647,536 frees, 538,021,213 bytes allocated
compiled 1000 keymaps in 1.949309s

The reduction in allocs is very nice, but the speedup is less than I hoped.

### DIFF
--- a/include/xkbcommon/xkbcommon-compose.h
+++ b/include/xkbcommon/xkbcommon-compose.h
@@ -232,7 +232,7 @@ enum xkb_compose_format {
  *
  * @memberof xkb_compose_table
  */
-struct xkb_compose_table *
+XKB_EXPORT struct xkb_compose_table *
 xkb_compose_table_new_from_locale(struct xkb_context *context,
                                   const char *locale,
                                   enum xkb_compose_compile_flags flags);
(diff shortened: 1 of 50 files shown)


wismill:
Nice! This could also help #539: it would be easier to cache the AST by just cloning the corresponding arena.

bluetech:
Also converting the scanner strdups bring the allocs down to:

total heap usage: 1,483,536 allocs, 1,483,536 frees, 512,020,213 bytes allocated

but the runtime isn't reduced by much. At this point memory allocation is not a significant component of the runtime, it's mostly the parser+scanner, compilation logic, atom interning, keysym lookup, etc.

PR #620, opened 2025-01-27
symbols: Allow levels with different keysyms and actions counts

### LINKED ISSUE #619: Drop NoSymbol in multi-keysym lists and NoAction() for multi-action lists
Consider:

```c
key <A> { [ { a, NoSymbol, c } ] };
augment key <A> { [ { NoSymbol, b, NoSymbol } ] };
key <B> { [ { 1, NoSymbol } ] };
```

Currently we keep the `NoSymbol` in the list. It is however useless, because:
1. Until #568 (not released) it could not be used to augment a previous entry: e.g. the second entry above will have no effect.
2. `NoSymbol` triggers no interpret, can be converted to no character and is usually associated to an empty level or an error.

I changed 1) in #568, but it complicates the code for no good reason. Indeed, with perspective I see no use case that would not be very niche.

So I propose that we compile the above to the *canonical*:

```c
key <A> { [ { a, c } ] };
key <B> { [ 1 ] };
```

The same should be done for *actions* with `NoAction()`.

wismill:
Contrary to groups, there is no reason for levels to restrict the same count of keysyms and actions. See the [related discussion](https://github.com/xkbcommon/libxkbcommon/pull/568#issuecomment-2592839603) with @mahkoh. 

This will allow better support for the use case Modifier + Group change, since we only want to communicate the modifier keysym to app, not the group keysym.

Also always normalize the multi-keysyms/actions lists by dropping `NoSymbol` and `NoAction`.

Fixes #619
 
TODO:
- [ ] Modify xkbcomp so that it ignores multi-actions per level, as it already ignores multi-keysyms.


Jules-Bertholet:
Do we really want to drop `NoAction()`? It’s really somewhat of a misnomer, as it does perform an action—breaking latches.

wismill:
@Jules-Bertholet I though about introducing `VoidAction()`, to mirror `NoSymbol/VoidSymbol`. Because currently we also do not have an easy way to erase previous explicit actions when using RMLVO:
- `NoAction()` does not overwrite the previous action.
- `replace key` works only in the same file or after a child file.
- `SetControls()` kind of work if you want to break latches. But it’s ugly.

The no/void qualifiers are not very clear, but at least we would be consistent with the keysyms.

PR #623, opened 2025-01-28
test/compose: reduce quickcheck iterations

bluetech:
They're a *bit* too slow for interactive test runs when running in debug+sanitizers.

### DIFF
--- a/test/compose.c
+++ b/test/compose.c
@@ -1094,7 +1094,7 @@ main(int argc, char *argv[])
     srand(seed);
 
     /* Determine number of loops for quickchecks */
-    size_t quickcheck_loops = 100; /* Default */
+    size_t quickcheck_loops = 50; /* Default */
     if (argc > 2) {
         /* From command-line */
         quickcheck_loops = (size_t)atoi(argv[2]);


test/compose.c
```diff
@@ -1130,12 +1131,17 @@ main(int argc, char *argv[])
...
     test_override(ctx);
-    test_traverse(ctx, quickcheck_loops);
     test_string_length(ctx);
     test_decode_escape_sequences(ctx);
     test_encode_escape_sequences(ctx);
     test_roundtrip(ctx);
 
+    /* Turn off noisy deprecated keysym warnings from here. */
+    xkb_context_set_log_level(ctx, XKB_LOG_LEVEL_INFO);
+    xkb_context_set_log_verbosity(ctx, 0);
```
wismill, line 1141: We may also consider to actually fix them at some point 😄

wismill (approved):

test/compose.c
wismill: See #625

bluetech: I removed this part since it's no longer needed.

PR #625, opened 2025-01-29
tests: Fix deprecated keysyms in Compose data

wismill:
Context: [see my comment](https://github.com/xkbcommon/libxkbcommon/pull/623#discussion_r1933226393) in #623.

The data set is big enough so we can just drop most sequences with deprecated keysyms and fix most keysyms an alternative non-deprecated name.

We keep a handful of them for testing purpose.


bluetech (approved):
Thanks, that's a better fix. I'll merge this and rebase.

PR #626, opened 2025-01-29
xkbcomp: Add stricter bounds for keycodes and levels

wismill:
Our current implementation uses continuous arrays indexed by keycodes and levels. This is simple and good enough for realistic keymaps. However, they are allowed to have big values that will lead to either memory exhaustion or a waste of memory (sparse arrays).

Added the much stricter upper bound `0xfff` that should still be plenty enough and provides stronger memory security.

### DIFF
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -387,6 +387,36 @@ struct xkb_mod_set {
     xkb_mod_mask_t explicit_vmods;
 };
 
+/*
+ * Our current implementation with continuous arrays does not allow efficient
+ * mapping of keycodes. Allowing the API max valid keycode XKB_KEYCODE_MAX could
+ * result in memory exhaustion or memory waste (sparse arrays) with huge enough
+ * valid values. Let’s be more conservative for now, based on the existing Linux
+ * keycodes.
+ */
+#define XKB_KEYCODE_MAX_IMPL 0xfff
+static_assert(XKB_KEYCODE_MAX_IMPL < XKB_KEYCODE_MAX,
+              "Valid keycodes");
+static_assert(XKB_KEYCODE_MAX_IMPL < darray_max_alloc(sizeof(xkb_atom_t)),
+              "Max keycodes names");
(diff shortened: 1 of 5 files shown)


src/keymap.h
```diff
@@ -387,6 +387,30 @@ struct xkb_mod_set {
     xkb_mod_mask_t explicit_vmods;
 };
 
+/*
+ * Our current implementation with continuous arrays does not allow efficient
+ * mapping of keycodes, let’s be further conservative.
+ */
+#define XKB_KEYCODE_MAX_IMPL 0xfff
```
bluetech, line 394: It should be possible to use a typed const instead of define, always seems better to me. The other consts in this file are defines but we can start with this one.

I also suggest using a decimal here, it's easier to understand the limit than in hex.

```suggestion
const xkb_keycode_t XKB_KEYCODE_MAX_IMPL = 4095;
```

src/keymap.h
```diff
@@ -387,6 +387,30 @@ struct xkb_mod_set {
...
+static_assert(XKB_KEYCODE_MAX_IMPL < darray_max_alloc(sizeof(xkb_atom_t)),
+              "Max keycodes names");
+static_assert(XKB_KEYCODE_MAX_IMPL < darray_max_alloc(sizeof(struct xkb_key)),
+              "Max keymap keys");
+
+/*
+ * Our current implementation with continuous arrays does not allow efficient
+ * mapping of levels, let’s be further conservative.
+ */
+#define XKB_LEVEL_MAX_IMPL 0xfff
```
bluetech, line 406:
```suggestion
const xkb_level_index_t XKB_LEVEL_MAX_IMPL = 4095;
```

bluetech (approved):
LGTM, we should definitely do this.

I left a couple of optional suggestions.

src/keymap.h
wismill: But it will not work with `static_assert`? We would need C23’s `constexpr` or using `extern` in the header + usual `assert`.

As for the hex format: no strong opinion here, I used it because it’s the format of Linux (upper) keycodes.

PR #630, opened 2025-01-30
test: Add check for keymap and compose compilation logs

wismill:
Let’s check a bit more our logs, especially the token positions.


bluetech (approved):
Great as always, thanks! After this is merged, I'll rebase the lazy location PR to make sure these tests pass.

PR #632, opened 2025-01-30
tools: Enable Compose file positional argument and piping

wismill:
Follow-up of #621.

### DIFF
--- a/tools/compile-compose.c
+++ b/tools/compile-compose.c
@@ -32,12 +32,13 @@
 #include "xkbcommon/xkbcommon-compose.h"
 #include "src/compose/dump.h"
 #include "src/keysym.h"
+#include "tools/tools-common.h"
 
 static void
 usage(FILE *fp, char *progname)
 {
     fprintf(fp,
-            "Usage: %s [--help] [--file FILE] [--locale LOCALE] [--test]\n",
+            "Usage: %s [--help] [--locale LOCALE] [--test] [FILE]\n",
             progname);
     fprintf(fp,
             "\n"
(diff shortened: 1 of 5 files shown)


tools/compile-compose.c
```diff
@@ -32,12 +32,13 @@
...
 #include "src/compose/dump.h"
 #include "src/keysym.h"
+#include "tools/tools-common.h"
 
 static void
 usage(FILE *fp, char *progname)
 {
     fprintf(fp,
-            "Usage: %s [--help] [--file FILE] [--locale LOCALE] [--test]\n",
+            "Usage: %s [--help] [--file FILE] [--locale LOCALE] [--test] [FILE]\n",
```
bluetech, line 41: Maybe hide `--file` from the usage?

tools/compile-compose.c
```diff
@@ -118,24 +116,55 @@ main(int argc, char *argv[])
...
+    if (optind < argc && !isempty(argv[optind])) {
+        /* Some positional arguments left: use a file */
+        if (path) {
+            fprintf(stderr, "ERROR: Path already provided\n");
+            usage(stderr, argv[0]);
+            exit(EXIT_INVALID_USAGE);
+        }
+        path = argv[optind++];
+        if (optind < argc) {
+            fprintf(stderr, "ERROR: Too much positional arguments\n");
```
bluetech, line 128:
```suggestion
            fprintf(stderr, "ERROR: Too many positional arguments\n");
```

bluetech (approved):

tools/compile-compose.c
wismill: Fixed here and in compile-keymap

tools/compile-compose.c
wismill: Fixed. Also added warning if used.

PR #633, opened 2025-01-30
Update license notices to SDPX short identifiers + update LICENSE

### LINKED ISSUE #628: Switch to SPDX License IDs
The huge license texts on many files are annoying. Lets switch them to SPDX comments. If it's good enough for Linux, it should be good enough for us. The copyright lines should be retained.

xkbcommon inherits some old X11 licenses, we need to be careful not to remove license text which doesn't have SPDX identifiers, if any.

bluetech:
Fix #628.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,30 +1,38 @@
 The following is a list of all copyright notices and license statements which
-appear in the xkbcommon source tree.
+appear in the xkbcommon source tree, along with the SPDX short license
+identifiers which source files use to refer to them. To see which parts are
+covered by which license and copyright, refer to the source files.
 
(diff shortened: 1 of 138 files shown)


bluetech:
There's a mistake with the sell variant - fixing.

bluetech:
OK, went over it again myself and I don't think there are any more mistakes.

I intend to merge this tomorrow.

sertonix:
Thanks for using SPDX identifiers! Do I understand this correctly that the license of the build binaries is `MIT AND MIT-open-group AND HPND AND HPND-sell-variant AND LicenseRef-digital-equipment-corporation` (excluding `X11`)?

bluetech:
Yes, that's correct.

PR #639, opened 2025-02-02
symbols: Enable writing keysyms list as UTF-8 strings

wismill:
Each Unicode code point of the string will be translated to their respective keysym, if possible. When such conversion is not possible, this will raise a syntax error.

This introduces the following syntax:

```c
// Single code point = single keysym
key <1> {["é"]}; // eacute
// String = translate each code point to their respective keysym
key <2> {["sßξك🎺"]}; // {s, ssharp, Greek_xi, Arabic_kaf, U1F3BA}
// Mix string and keysyms
key <3> {[{"ξ", Greek_kappa, "β"}]}; // { Greek_xi, Greek_kappa, Greek_beta}
```

Mixing string and keysyms would allow to mix keysyms with and without Unicode translation[^1].

Implements #636

TODO:
- [x] Tests
- [x] Doc
- [x] Changelog
- [x] Rebase on #717
- [ ] ~~enable also in Compose?~~

[^1]: Although right now I have no use case in mind. It’s more about the principle of least astonishment: since `"é"` translates to `eacute`, then `{"é", e}` and `{eacute, e}` should be both valid and result in the same keysym sequence. Then users may get creative and mix e.g. modifier or system keysyms with characters.

### DIFF
--- a/doc/keymap-format-text-v1.md
+++ b/doc/keymap-format-text-v1.md
@@ -104,7 +104,8 @@ Some additional resources are:
 
   The complete list of keysyms is defined in `xkbcommon/xkbcommon-keysyms.h`.
 
-  See [xkb_symbols] for further details on binding keysyms to keycodes.
+  @sa [xkb_symbols] for further details on binding keysyms to keycodes.
+  @sa [Keysym syntax][@ref keysym-syntax]
   </dd>
   <dt><a name="modifier-def">Modifier</a></dt>
   <dd>
@@ -569,6 +570,104 @@ Comments are introduced following either `//` or `#` until the end of the line.
   - _decimal floating-point number:_ `1.23`, etc.
   - _hexadecimal integer:_ prefixed with `0x`: `0x123`, `0xff`, `0xAB`, etc.
   </dd>
+</dl>
+
+### Keysyms {#keysym-syntax}
+
+[Keysyms][keysym] may be written in multiple ways:
+
+<dl>
+<dt>Name</dt>
+<dd>
+
+Keysym names are defined in `xkbcommon/xkbcommon-keysyms.h`; remove the
+`XKB_KEY_` prefix to get the name.
+
+Example: the keysym `0xffbe = XKB_KEY_F1` has the name `F1`.
+</dd>
+<dt>Unicode</dt>
+<dd>
+
+The *Unicode* syntax `Unnnn` denotes a keysym whose corresponding character is
+the Unicode code point `U+nnnn`, where `nnnn` is an hexadecimal number in the
+range `0x100 .. 0x10ffff`. The resulting keysym value is `0x01000000 + nnnn`.
+
+Example: `U1F3BA` has value `0x0101F3BA` and corresponds to the code point
+`U+1F3BA`: ‘🎺’ (TRUMPET).
+</dd>
+<dt>String literal</dt>
+<dd>
+
+A keysym or list of keysyms can be written as a *string literal*, with the
+following semantics:
+
+1. The string must be encoded in UTF-8.
+2. If the encoding is invalid, it raises a syntax error.
+3. If the string expands to 0 Unicode code point, the resulting keysym is
+   `NoSymbol`.
+4. If the string expands to 1 Unicode code point, the resulting keysym is the
+   output of `xkb_utf32_to_keysym`.
+5. Otherwise the string expands to a list `{ ... }` with each Unicode code point
+   converted via `xkb_utf32_to_keysym`.
+
+   @note This is only possible where multiple keysyms is valid, e.g. in the
+   @ref key-multiple-symbols-per-level "key symbols".
+
+
+Examples:
+
+<table>
+  <thead>
+    <tr>
+      <th>Keysym string</th>
+      <th>Keysym name</th>
+      <th>Keysym value</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>`""`</td>
+      <td>`NoSymbol`</td>
+      <td>`0x0000`</td>
+    </tr>
+    <tr>
+      <td>`"a"`</td>
+      <td>`a`</td>
+      <td>`0x0061`</td>
+    </tr>
+    <tr>
+      <td>`"ü"`</td>
+      <td>`udiaeresis`</td>
+      <td>`0x00dc`</td>
+    </tr>
+    <tr>
+      <td>`"🎺"`</td>
+      <td>`U1F3BA`</td>
+      <td>`0x0101F3BA`</td>
+    </tr>
+    <tr>
+      <td>`"g̃"`</td>
+      <td>`{g, combining_tilde}`</td>
+      <td>`{0x0101F3BA, 0x01000303}`</td>
+    </tr>
+  </tbody>
+</table>
+
+@sa @ref keysyms-string "Keysym strings"
+
+@since 1.9.0
+</dd>
+<dt>Numeric value</dt>
+<dd>
+
+A keysym can be written directly with its *numeric* value: e.g. `0x61` is `a`.
+
+@note Digits `0 .. 9` have a special treatment because they are interpreted as
+names, not values. E.g. `1` is the keysym with name `1` and value `0x31`.
+
+@warning This syntax should be avoided for its poor readability, except if it
+is not possible to write the keysym with the previous syntaxes.
+</dd>
 </dl>
 
 ### Keywords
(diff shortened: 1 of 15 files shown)


mahkoh:
https://github.com/mahkoh/kbvm/pull/19

whot (commented):
quite handwavy but this lgtm on a quick review

wismill:
Kbvm enables keysyms strings as well in `interpret`s and `modmap`s, but I am not sure this is a road I want to take. I get that if we allow strings to translate to single-keysyms, then we may want to do so everywhere else for consistency. 

But if we consider instead that they translate to keysym *lists* of length 1 then we would somehow break the semantics: `interpret`s accept only single keysyms and `modmap`s accept only mixed lists of single keys and single keysyms. Moreover, it is really unusual to associate characters keysyms to actions. I have no use case in mind; do you?

wismill:
Note: while this MR is only for *input* files and keep serialization with proper keysyms for compatibility, it would be great to have an option for pretty-printing with these strings.

mahkoh:
>But if we consider instead that they translate to keysym lists of length 1

I treat `a` and `{ a }` the same.

>I have no use case in mind; do you?

I don't want people to be surprised. If a user starts out by writing `"ア"` in a `key`, then they would be surprised if they were expected to look up the keysym name in other positions.

wismill:
@mahkoh While working on the documentation of this feature, I introduced the following example taken from the standard Compose file:

```c
key <AC05> { [ ⁧"لا"⁩  , ⁧"لآ"⁩   ] }; // ⁧لا⁩ ⁧لآ⁩ Arabic Lam-Alef ligatures decomposed
```

But in order to display it correctly, I needed to insert [bidi Unicode controls](https://www.w3.org/International/questions/qa-bidi-unicode-controls). Consider the following example without these controls:

```c
key <AC05> { [ "لا"  , "لآ"   ] };
```

While the latter will parse correctly although its display is wrong, the former example is currently a syntax error in libxkbcommon (and a possible panic in mahkoh/kbvm#30).

While we currently only require a file encoding compatible with ASCII, it may be worth considering parsing and discarding such BiDi controls with their UTF-8 encoding.

mahkoh:
Ignoring unicode characters with the whitespace property between tokens sounds reasonable to me.

wismill:
BiDi controls do not have the whitespace property. And if we allow them, we need to take care of the [Trojan Source](https://trojansource.codes/) attacks.

Jules-Bertholet:
@wismill There are two kinds of bidi control characters. The directional override controls are dangerous and shouldn’t be allowed. But the danger of LRM and RLM is much lesser, as any strongly directional character (including Latin letters) overrides their effect.

The whitespace property you want to use is `Pattern_White_Space`; it includes LRM and RLM. https://www.unicode.org/reports/tr31/#Whitespace

doc/keymap-format-text-v1.md
```diff
@@ -1602,13 +1692,34 @@ level 1 and `XKB_KEY_Q` for level 2. These levels are configured by the
...
+key <AD08> { [ {i, j}        , U0132         ] }; // IJ Dutch digraph
+key <AC05> { [ {g, U0303}    , {G, U0303}    ] }; // G̃ Guarani letter
+key <AB05> { [ {U0644, U0627}, {U0644, U0622}] }; // ⁧لا⁩ ⁧لآ⁩ Arabic Lam-Alef ligatures decomposed
+key <AD01> { [ {c, U2019, h} , {C, U2019, h} ] }; // C’H Breton trigraph
+```
 
-    key <AD01> { [ {a, b}, Q ] };
+In this example, the keycode `<AD08>` produces two symbols on level 1
+(`XKB_KEY_i` and `XKB_KEY_j`) and one symbol (Unicode keysym `U0132` for “Ĳ”)
+on level 2. `<AD08>` and `<AD01>` produces letters that have no *precomposed*
```
bluetech, line 1708: produces -> produce

src/xkbcomp/ast-build.c
```diff
@@ -200,6 +206,73 @@ ExprAppendKeysymList(ExprDef *expr, xkb_keysym_t sym)
...
+ExprKeysymListAppendString(struct scanner *scanner,
+                           ExprDef *expr, const char *string)
+{
+    /* TODO: use strnlen with max len = 4 * MAX_KEYSYMS_LIST_LENGTH */
+    size_t len = strlen(string);
+    while (len) {
+        size_t count = 0;
+        uint32_t cp = utf8_next_code_point(string, len, &count);
+        if (cp == INVALID_UTF8_CODE_POINT) {
+            scanner_err(scanner, XKB_ERROR_INVALID_FILE_ENCODING,
```
bluetech, line 219: Using `scanner_err` from the parser is a little problematic because the location might be off (if the parse rule requires further tokens before the action is called). But I think it's not a blocker (we probably already have a couple of these).

meson.build
```diff
@@ -47,6 +47,7 @@ cflags = [
     '-Wwrite-strings',
     '-Wno-documentation-deprecated-sync',
     '-Wno-pedantic',
+    '/utf-8',
```
bluetech, line 50: Feel free to cherry-pick this to master if you'd like.

bluetech (commented):
Not a proper review, but some comments.

src/xkbcomp/scanner.c
```diff
@@ -61,6 +61,13 @@ _xkbcommon_lex(YYSTYPE *yylval, struct scanner *s)
     /* Skip spaces. */
     while (is_space(scanner_peek(s))) scanner_next(s);
 
+    /*
+     * Skip U+200E LEFT-TO-RIGHT MARK and U+200F RIGHT-TO-LEFT MARK,
+     * assuming UTF-8 encoding
+     */
+    if (scanner_lit(s, u8"\u200E") || scanner_lit(s, u8"\u200F"))
```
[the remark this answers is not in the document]
wismill: We do not expect consecutive BiDi controls, but this case would be handled just fine, wouldn’t it?

src/xkbcomp/scanner.c
```diff
@@ -61,6 +61,13 @@ _xkbcommon_lex(YYSTYPE *yylval, struct scanner *s)
     /* Skip spaces. */
     while (is_space(scanner_peek(s))) scanner_next(s);
 
+    /*
+     * Skip U+200E LEFT-TO-RIGHT MARK and U+200F RIGHT-TO-LEFT MARK,
```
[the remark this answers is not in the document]
wismill: Done

doc/keymap-format-text-v1.md
```diff
@@ -561,6 +562,94 @@ Comments are introduced following either `//` or `#` until the end of the line.
...
+
+Keysym names are defined in `xkbcommon/xkbcommon-keysyms.h`; remove the
+`XKB_KEY_` prefix to get the name.
+
+Example: the keysym `0xffbe = XKB_KEY_F1` has the name `F1`.
+</dd>
+<dt>Unicode</dt>
+<dd>
+
+The *Unicode* syntax `Unnnn` denotes a keysym which corresponding character is
```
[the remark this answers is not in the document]
wismill: Never got this one right it seems 😅. Fixed

doc/keymap-format-text-v1.md
```diff
@@ -561,6 +562,94 @@ Comments are introduced following either `//` or `#` until the end of the line.
...
+
+<table>
+  <thead>
+    <tr>
+      <th>Keysym string</th>
+      <th>Keysym name</th>
+      <th>Keysym value</th>
+    </tr>
+  </thead>
+  <tbody>
```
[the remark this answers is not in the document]
wismill: Good point! I added this in the changelog as well.

doc/keymap-format-text-v1.md
wismill: Fixed

src/xkbcomp/ast-build.h
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "ast.h"
+#include "src/scanner-utils.h"
```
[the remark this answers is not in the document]
wismill: Done

meson.build
wismill: There is nothing that requires it atm, let’s wait to merge the MR.

src/xkbcomp/ast-build.c
wismill: Interesting. But `STRING` is currently always the last token when it is called, isn’t it? I added indication of the failing string byte position + code point position, plus the corresponding log test.

wismill:
Waiting for #717 to land.

PR #642, opened 2025-02-03
xkbcomp/scanner: avoid unneeded strdup of IDENT tokens

bluetech:
The allocation is immediately discarded, either turned into a keysym or
an atom. So use an sval slice into the input string instead strdup'ing.

memusage ./release/bench-compile-keymap --iter=1000 --layout us,de --variant ,neo

Before:

    Memory usage summary: heap total: 534063576, heap peak: 581022, stack peak: 18848
            total calls   total memory   failed calls
     malloc|   11240525      291897104              0
    realloc|    1447657      192307328              0  (nomove:37629, dec:0, free:0)
     calloc|     430573       49859144              0
       free|   13993903      534063576

After:

    Memory usage summary: heap total: 506839909, heap peak: 581022, stack peak: 18960
            total calls   total memory   failed calls
     malloc|    8016419      264673437              0
    realloc|    1447657      192307328              0  (nomove:37278, dec:0, free:0)
     calloc|     430573       49859144              0
       free|    0769797      506839909

### DIFF
--- a/src/xkbcomp/parser.y
+++ b/src/xkbcomp/parser.y
@@ -15,7 +15,6 @@
 #include "xkbcomp/xkbcomp-priv.h"
 #include "xkbcomp/ast-build.h"
 #include "xkbcomp/parser-priv.h"
-#include "scanner-utils.h"
 #include "keysym.h"
 
 struct parser_param {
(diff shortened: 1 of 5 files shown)


src/xkbcomp/parser.y
```diff
@@ -59,25 +58,32 @@ _xkbcommon_error(struct parser_param *param, const char *msg)
...
     }
 
-    if (istreq(name, "none") || istreq(name, "voidsymbol")) {
+    if (isvaleq(name, SVAL_LIT("none")) || isvaleq(name, SVAL_LIT("voidsymbol"))) {
         *sym_rtrn = XKB_KEY_VoidSymbol;
         return true;
     }
 
-    sym = xkb_keysym_from_name(name, XKB_KEYSYM_NO_FLAGS);
+    char buf[256];
```
bluetech, line 75: This is a bit dubious, but I think we can rely on a max keysym name size, @wismill WDYT?

Ideally we'd have a variant of xkb_keysym_from_name which takes a buf+len instead of C string, then we could avoid this copy entirely (and would also be nice for FFI bindings in languages which don't use C strings), but it's not immediately straightforward so I went with this copy for now.

PR #643, opened 2025-02-03
meson: suppress MSVC C4100 unreferenced formal parameter warnings

bluetech:
Equivalent to `-Wno-unused-parameter` which we use.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -33,6 +33,7 @@ dir_libexec = get_option('prefix')/get_option('libexecdir')/'xkbcommon'
 cflags = [
     '-fno-strict-aliasing',
     '-Wno-unused-parameter',
+    '/wd4100', # MSVC C4100: unreferenced formal parameter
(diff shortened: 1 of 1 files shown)


meson.build
```diff
@@ -33,6 +33,7 @@ dir_libexec = get_option('prefix')/get_option('libexecdir')/'xkbcommon'
 cflags = [
     '-fno-strict-aliasing',
     '-Wno-unused-parameter',
+    '/wd4100', # MSVC C4100: unreferenced formal parameter
```
whot, line 36: can we move this up or down? It looks a bit jarring to me, moving it a line up would separate the warnings from the strict aliasing flag, or moving it to the bottom would group the `-W` and the "not `-W`" better.

bluetech: My intention was to keep it next to its gcc companion. But I can move it down.

PR #644, opened 2025-02-04
xkbcomp/parser: enable Bison detailed syntax error

bluetech:
It's not much, but instead of

xkbcommon: ERROR: [XKB-769] (unknown file):5:25: syntax error

we get

xkbcommon: ERROR: [XKB-769] (unknown file):5:25: syntax error, unexpected +, expecting INTEGER

which is at least a little helpful.

### DIFF
--- a/src/xkbcomp/parser.y
+++ b/src/xkbcomp/parser.y
@@ -88,70 +88,72 @@ resolve_keysym(struct parser_param *param, struct sval name, xkb_keysym_t *sym_r
 %lex-param      { struct scanner *param_scanner }
 %parse-param    { struct parser_param *param }
 
+%define parse.error detailed
+
 %token
-        END_OF_FILE     0
-        ERROR_TOK       255
-        XKB_KEYMAP      1
-        XKB_KEYCODES    2
-        XKB_TYPES       3
-        XKB_SYMBOLS     4
-        XKB_COMPATMAP   5
-        XKB_GEOMETRY    6
-        XKB_SEMANTICS   7
-        XKB_LAYOUT      8
-        INCLUDE         10
-        OVERRIDE        11
-        AUGMENT         12
-        REPLACE         13
-        ALTERNATE       14
-        VIRTUAL_MODS    20
-        TYPE            21
-        INTERPRET       22
-        ACTION_TOK      23
-        KEY             24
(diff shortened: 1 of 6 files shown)


wismill:
I did not find a byacc equivalent for this feature. Maybe we need to guard it with a `bison` check?

bluetech:
> I did not find a byacc equivalent for this feature. Maybe we need to guard it with a bison check?

Yeah but it'd be a bit of a hassle, I think we'd need to template parser.y :( Then we'd also need to make the test/log snapshot tests conditional.

Alternatives are to drop this change, drop byacc, or ask byacc to add this feature (unlikely, but Thomas Dickey did add a feature the byacc in the past when I asked nicely).

bluetech:
I decided to drop byacc. I don't think it's really used so it shouldn't hold us back. But if someone complains we can probably bring it back with some ifdefs or similar.

src/xkbcomp/parser.y
```diff
@@ -4,9 +4,7 @@
  */
 
 /*
- * The parser should work with reasonably recent versions of either
- * bison or byacc.  So if you make changes, try to make sure it works
- * in both!
+ * The parser should work with reasonably recent versions of bison.
```
wismill, line 7: It would be great to have a version range here. But since it is the role of the meson file, maybe remove the comment entirely, so we only handle the version in one place.

Maybe instead we could document here why we drop `byacc`:

```suggestion
 * The parser supports only bison, because byacc does not support the following features: …
```

wismill (changes requested):
That’s a great improvement! 👏

firasuke:
> @firasuke Could you provide more details to motivate this work in a _new issue_?

Sure, a good number of smaller distributions are using `byacc` as the default YACC implementation, and providing `bison` as a symlink to `byacc`. You would argue that these distributions should carry the burden of patching `libxkbcommon` to support `byacc` but then again if it already did support `byacc`, why remove it in the first place?

wismill:
@firasuke as written before, please *create an issue*, because we will not tackle this in a closed MR.

In the new issue, please write a *self-contained* description that:
- links to this MR
- provides examples of distributions with such configuration (with links)
- links to at least one relevant *bug report* in these distributions

PR #653, opened 2025-02-07
tools: Fix compile-keymap default KcCGST values

wismill:
Values may be read from the environment!

While working on a xkbcomp MR I was very confused by `<BKSP>` being mapped differently with and without an option that does *not* affect it. It turns out it is just `compile-keymap` that replaces null options with my env variable `$XKB_DEFAULT_OPTIONS = terminate:ctrl_alt_bksp`, as expected.

But debugging with `--kccgst` did not help, because it does not take the environment into account.

Now I am wondering how we could document this better in the tool and if how to read the env to output `usage` with correct default values.

### DIFF
--- a/tools/compile-keymap.c
+++ b/tools/compile-keymap.c
@@ -270,10 +270,13 @@ print_rmlvo(struct xkb_context *ctx, const struct xkb_rule_names *rmlvo)
 }
 
 static int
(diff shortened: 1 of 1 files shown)


bluetech (approved):
> Now I am wondering how we could document this better in the tool

Maybe an `ENVIRONMENT VARIABLES` section in the manpage?

wismill:
Will address this in a follow-up PR.

PR #654, opened 2025-02-07
xkbcomp/expr: remove unused ExprResolveKeyCode

bluetech:
This function was added in commit 4e22851141d89436c0659e68da57e91bbf971461. But that commit also changed the grammar:

```diff
-KeyNameDecl    :       KeyName EQUALS Expr SEMI
+KeyNameDecl    :       KeyName EQUALS KeyCode SEMI
```

i.e. while before you could write

    <AE01> = 9+1;

now this is a syntax error, an integer literal is expected. I'm not sure if it was intended to remove this ability. In any case, this rendered `ExprResolveKeyCode` useless since there's no longer an expression to evaluate, and after some refactoring it went unused.

Even if we choose to restore Expr here, I don't see a reason for the specialized function over `ExprResolveInteger` except the type (which we should probably widen from int to int64_t...). So remove it.

+ Commit which removes an outdated comment.


wismill:
Agreed, I did this too in #605.

> which we should probably widen from int to int64_t...

See #605

wismill (approved):

bluetech:
Ah missed it sorry! The first two commits from #605 look good to me, if you'd like to cherry pick them.

PR #661, opened 2025-02-13
xkbcomp: Fix incorrect memory comparison

wismill:
Spotted by `clang-tidy`:

```
../src/keymap-priv.c:146:16: warning: comparing object representation of
                             type 'union xkb_action' which does not have
                             a unique object representation; consider
                             comparing the members of the object manually
                             [bugprone-suspicious-memory-comparison]
````

Indeed, from “EXP42-C. Do not compare padding data”[^1]:

> unnamed members of objects of structure and union type do not participate
> in initialization. Unnamed members of structure objects have indeterminate
> representation even after initialization.

But since we [zero the action](https://github.com/xkbcommon/libxkbcommon/blob/350931add1552013e3ac77dda6752e1c4cad44e7/src/xkbcomp/symbols.c#L912), I do not think the current behavior can cause a bug, can it?

Fixed by using a dedicated comparison function for `union xkb_action`.

[^1]: https://wiki.sei.cmu.edu/confluence/display/c/EXP42-C.+Do+not+compare+padding+data

### DIFF
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -137,14 +137,72 @@ XkbLevelsSameSyms(const struct xkb_level *a, const struct xkb_level *b)
     return memcmp(a->s.syms, b->s.syms, sizeof(*a->s.syms) * a->num_syms) == 0;
 }
 
+static bool
+action_equal(const union xkb_action *a, const union xkb_action *b)
+{
+    if (a->type != b->type)
+        return false;
+
+    /* Ensure we support all action types */
+    static_assert(ACTION_TYPE_PRIVATE == 15 &&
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/keymap-priv.c [hunk 1]
```diff
@@ -137,14 +137,72 @@ XkbLevelsSameSyms(const struct xkb_level *a, const struct xkb_level *b)
     return memcmp(a->s.syms, b->s.syms, sizeof(*a->s.syms) * a->num_syms) == 0;
 }
 
+static bool
+action_equal(const union xkb_action *a, const union xkb_action *b)
+{
+    if (a->type != b->type)
+        return false;
+
+    /* Ensure we support all action types */
```
bluetech, line 146: Heh that's clever :)

bluetech (approved):
LGTM

PR #662, opened 2025-02-13
Clang tidy: round 2

wismill:
~~CI will pass once #661 is merged.~~

EDIT: removed clang-tidy CI.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -16,6 +16,7 @@ Checks:
   - performance-*
   - google-*
   - -google-readability-braces-around-statements
+  - -google-readability-todo
   # - misc-*
   # - -misc-include-cleaner
 # ExtraArgs:
@@ -47,5 +48,6 @@ CheckOptions:
   llvm-else-after-return.WarnOnUnfixable: 'false'
   llvm-qualified-auto.AddConstToQualified: 'false'
   bugprone-suspicious-enum-usage.StrictMode: 'true'
+  bugprone-signed-char-misuse.CharTypdefsToIgnore: 'int8_t'
 SystemHeaders:   false
 ...
--- a/bench/bench.c
+++ b/bench/bench.c
@@ -127,7 +127,8 @@ predict(long long t1, long long t2, struct estimate *est)
 {
     const long long t = fit(t1, t2);
     est->elapsed = t;
-    est->stdev = truncl(sqrtl((long double)sqr(t1 - t) + (long double)sqr(t2 - 2 * t)));
+    est->stdev =
+        llroundl(sqrtl((long double)sqr(t1 - t) + (long double)sqr(t2 - 2 * t)));
 }
 
 #define high(t, prec) ((t) + (prec))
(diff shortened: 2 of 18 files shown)


wismill:
Well, no chance here. The `clang-tidy` on the runner seems too old for the current config. Maybe there is an action we can use or a runner with the expected clang-tidy version?

wismill:
The issue is that my `.clang-tidy` config works only with the latest clang-tidy, 19.1 ([see the release notes](https://releases.llvm.org/19.1.0/tools/clang/tools/extra/docs/ReleaseNotes.html#improvements-to-clang-tidy)). It is not yet available in the Github runners (latest is 18.1 on Ubuntu 24.0.4).

Not sure I want to dig how to make it compatible with earlier versions, or maintain two config versions.

wismill:
Removed the clang-tidy CI, we’ll deal with it in a further PR.

src/x11/keymap.c
```diff
@@ -197,7 +197,7 @@ translate_action(union xkb_action *action, const xcb_xkb_action_t *wire)
     case XCB_XKB_SA_TYPE_SET_GROUP:
         action->type = ACTION_TYPE_GROUP_SET;
 
-        action->group.group = wire->setgroup.group;
+        action->group.group = (int32_t) wire->setgroup.group;
```
bluetech, line 200: IIRC setgroup->group is an int8_t, so why does clang-tidy wants a cast here?

bluetech (approved):
LGTM

src/x11/keymap.c
wismill: The actual warning:
```
../src/x11/keymap.c:200:31: warning: 'signed char' to 'int32_t' (aka 'int') conversion; consider casting to 'unsigned char' first. [bugprone-signed-char-misuse]
  200 |         action->group.group = wire->setgroup.group;
      |                               ^
../src/x11/keymap.c:213:31: warning: 'signed char' to 'int32_t' (aka 'int') conversion; consider casting to 'unsigned char' first. [bugprone-signed-char-misuse]
  213 |         action->group.group = wire->latchgroup.group;
      |                               ^
../src/x11/keymap.c:226:31: warning: 'signed char' to 'int32_t' (aka 'int') conversion; consider casting to 'unsigned char' first. [bugprone-signed-char-misuse]
  226 |         action->group.group = wire->lockgroup.group;
      |                               ^
```

From the doc about [`bugprone-signed-char-misuse`](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/signed-char-misuse.html):

> When the code contains an explicit `signed char` -> integer conversion, the human programmer probably expects that the converted value matches with the character code (a value from `[0..255]`), however, the actual value is in `[-128..127]` interval. To avoid this kind of misinterpretation, the desired way of converting from a `signed char` to an integer value is converting to `unsigned char` first, which stores all the characters in the positive `[0..255]` interval which matches the known character codes.

The issue is that `int8_t` is just a typedef, so we get this unrelated warning.

wismill: I added `int8_t` to the option `CharTypdefsToIgnore` in `.clang-tidy` to avoid the explicit cast.

PR #664, opened 2025-02-14
tools: Do not load names from the environment by default

wismill:
Our tools are debugging tools and as such we need to have complete control to be able to reproduce setups. This is not currently the case, as we do not use `XKB_CONTEXT_NO_ENVIRONMENT_NAMES` by default nor can we set it.

So it is very easy to forget about the various `XKB_DEFAULT_*` environement variables for the default RMLVO values, then to get puzzled by unexpected results.

Added to that, these environment variables do not work correctly in `xkbcli-compile-xeymap`: calling the tool without RMLVO values will use these variables only if the RMLVO values are set explicitly empty or if the various *constants* `DEFAULT_XKB_*` are empty. This is unexpected, as the environment variables should *always* be used unless:

- `XKB_CONTEXT_NO_ENVIRONMENT_NAMES` is used (not the case here);
- the variable is empty; in this case the constants `DEFAULT_XKB_*` are used.

Fixed by the following *breaking change*: make the tools use `XKB_CONTEXT_NO_ENVIRONMENT_NAMES` *by default*, unless the new `--enable-environment-names` option is used.

### DIFF
--- a/changes/tools/+enable-environment-names.breaking.md
+++ b/changes/tools/+enable-environment-names.breaking.md
@@ -1,0 +1,3 @@
+The tools do not load the *default* RMLVO (rules, model, layout, variant, options)
+values from the environment anymore. The previous behavior may be restored by using
+the new `--enable-environment-names` option.
--- a/test/tool-option-parsing.py
+++ b/test/tool-option-parsing.py
@@ -33,19 +33,6 @@ except KeyError:
         file=sys.stderr,
     )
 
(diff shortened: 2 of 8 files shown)


bluetech (approved):

wismill:
Removed the incompatibility of `--rmlvo` and `--enable-environment-names` when building the internal tool, and comment the requirement of the internal API.

PR #671, opened 2025-03-07
keymap: fix modular arithmetic

mahkoh:
Consider num_groups = 1.

Fixes: 67b03cea ("state: correctly wrap state->locked_group and ->group")

-----

This should be backported.

### DIFF
--- a/changes/api/+group-wrap.bugfix.md
+++ b/changes/api/+group-wrap.bugfix.md
@@ -1,0 +1,7 @@
+Fixed segfault due to invalid arithmetic to bring *negative* layout indexes
+into range. It triggers with the following settings:
+
+- layouts count (per key or total) N: `N > 0`, and
+- layout index n: `n = - k * N` (`k > 0`)
+
+Note that these settings are unlikely in usual use cases.
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -205,6 +205,7 @@ XkbLevelsSameActions(const struct xkb_level *a, const struct xkb_level *b)
     return true;
 }
 
+/* See: XkbAdjustGroup in Xorg xserver */
 xkb_layout_index_t
 XkbWrapGroupIntoRange(int32_t group,
                       xkb_layout_index_t num_groups,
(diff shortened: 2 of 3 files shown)


wismill:
Good catch! I fixed the compilation error, added a test and further comments.

wismill:
Note that this is unlikely to trigger:
- when there is only one group, no group switch keysym is expected; 
- when there are multiple groups, it triggers only for negative multiples of the group count, e.g. -2 and -4 for 2 groups. This is difficult to achieve with the current xkeyboard-config.

So this explains it was undetected for so long.

mahkoh:
This causes a segfault in SDL because it initializes its group to -1 and then calls `xkb_keymap_num_levels_for_key` before the group gets set to something sensible.

wismill:
What is SDL? Do you have a link to the corresponding bug report?

mahkoh:
https://github.com/libsdl-org/SDL

There is no bug report. I encountered this when running https://gitlab.com/amini-allight/radv-bug-demo/. It only triggers if the first group does not have a name.

wismill:
While we will try to release soon with the fix, it seems still quite unlikely to trigger the bug in usual situations, e.g. when taking the state from the compositor (in the context of [my previous comment](https://github.com/xkbcommon/libxkbcommon/pull/671#issuecomment-2708354817)).

wismill:
Added changelog entry and more tests, especially every function in the API taking a `xkb_layout_index_t` parameter.

changes/api/+group-wrap.bugfix.md
```diff
@@ -0,0 +1,7 @@
+Fixed segfault due to invalid arithmetic to brought *negative* layout indexes
```
bluetech, line 1:
```suggestion
Fixed segfault due to invalid arithmetic to bring *negative* layout indexes
```

bluetech (approved):
Thanks @mahkoh and @wismill. LGTM.

changes/api/+group-wrap.bugfix.md
wismill: Fixed

PR #673, opened 2025-03-08
doc: Warn about the options order

### LINKED ISSUE #672: Options provided in a specific order by user seem to be sorted
I'm trying to define a custom option for the keypad to use **with** `keypad:oss`, named `custom:a5d`. Here it is:
```
hidden partial keypad_keys
xkb_symbols "a5d" {
    key.type[Group1]="FOUR_LEVEL_MIXED_KEYPAD" ;

    key <KP5> { [ KP_Down, NoSymbol, NoSymbol, NoSymbol ] };
};
```
However, when xkbcli compiles this option with `keypad:oss`, it puts it before `keypad:oss`, which overrides it.
```
$ xkbcli compile-keymap --verbose --layout 'us' --options 'keypad:oss,kpdl:commaoss,custom:a5d'
…
xkbcommon: DEBUG: Compiling from KcCGST: keycodes 'evdev+aliases(qwerty)', types 'complete', compat 'complete', symbols 'pc+us+custom(a5d)+inet(evdev)+keypad(oss)+kpdl(commaoss)'
```
It's not quite possible to use these two options together if the order is changed.

wismill:
Options are always applied depending of their order in the relevant rules files, independently of their order in the relevant RMLVO config.

Since this is counter-intuitive (e.g. when using `xkbcli`), let’s highlight this in the documentation.

Fixes #672

### DIFF
--- a/doc/rules-format.md
+++ b/doc/rules-format.md
@@ -337,6 +337,11 @@ is resolved and the file is parsed to get the [rule sets].
 Then *each rule set* is checked against the provided [MLVO] configuration,
 following their *order* in the rules file.
 
+@important @anchor irrelevant-options-order Contrary to layouts and variants,
+the *options order* in a [MLVO] configuration (e.g. via `xkbcli`) is irrelevant
+for its resolution: only the order of the rules matters. See
+“@ref rules-options-example ""” for an illustration.
+
 If a [rule] matches in a @ref rule-set-def "rule set", then:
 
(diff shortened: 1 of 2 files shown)


0n-s:
Yes, this should clear up any potential confusion about this behaviour, especially considering rules files are probably somewhat deeper in XKB than most users delve. Thank you!

PR #676, opened 2025-03-13
Rules: KcCGST public API

### LINKED ISSUE #669: improve tools for rules debugging
Continuation of https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/793#note_2797505

> Q: Any tool showing resulting symbols string, to debug rules?

> A: There is unfortunately no public API for this. Either:
>   - use the --verbose options or compile libxkbcommon
>   - compile [libxkbcommon](https://github.com/xkbcommon/libxkbcommon/) locally, then use the internal tool `build/compile-keymap --kccsgt --include … --layout … --variant …`.

So it's buried really deep for now, and is not documented.
Even if someone is lucky enough to deduce it, he has to compose complex pipes currently:
```
$ xkbcli compile-keymap --verbose --test --model applealu_iso --layout us  2>&1 | grep KcCGST
```
Could we improve the situation e.g. by adding `xkbcli compile-keymap --kccsgt` option, or expanding existing `--rmlvo` option there?

wismill:
- rules: Added `xkb_components_names_from_rules()`

  This is mainly for debugging purposes and to enable displaying KcCGST values from RMLVO resolution in `xkbcli compile-keymap --kccgst`.
- tools: Make `--kccgst` `compile-keymap` option public

  The new public option `--kccgst` enables to display the result of RMLVO resolution to KcCGST components.

  This option has the same function than `setxkbmap -print`. This is particularly useful for debugging issues with the rules.

  Before this commit it was a private API. This commit enables us to remove the *internal* version of `xkbcli-compile-keymap`.

Fixes #669

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -20,6 +20,7 @@
 #ifndef _XKBCOMMON_H_
 #define _XKBCOMMON_H_
 
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdarg.h>
@@ -281,13 +282,26 @@ typedef uint32_t xkb_led_mask_t;
 #define xkb_keycode_is_legal_x11(key) ((key) >= 8 && (key) <= 255)
 
 /**
- * Names to compile a keymap with, also known as RMLVO.
+ * @defgroup rules-api Rules
+ * Utility functions related to *rules*, whose purpose is introduced in:
+ * @ref xkb-the-config "".
+ *
+ * @{
+ */
+
+/**
+ * Names to compile a keymap with, also known as [RMLVO].
  *
  * The names are the common configuration values by which a user picks
  * a keymap.
  *
  * If the entire struct is NULL, then each field is taken to be NULL.
  * You should prefer passing NULL instead of choosing your own defaults.
+ *
+ * @see [Introduction to RMLVO][RMLVO]
+ * @see @ref rules-api ""
+ *
+ * [RMLVO]: @ref RMLVO-intro
  */
 struct xkb_rule_names {
     /**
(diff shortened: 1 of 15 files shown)


test/rules-file-includes.c
```diff
@@ -49,24 +45,43 @@ test_rules(struct xkb_context *ctx, struct test_data *data)
         fprintf(stderr, "Expecting: %s\t%s\t%s\t%s\n",
                 data->keycodes, data->types, data->compat, data->symbols);
 
-    if (!xkb_components_from_rules(ctx, &rmlvo, &kccgst, NULL)) {
-        fprintf(stderr, "Received : FAILURE\n");
-        return data->should_fail;
+    bool passed = true;
+    for (int k = 0; k < 2; k++) {
```
bluetech, line 49: Any particular reason to test the private API?

test/tool-option-parsing.py
```diff
@@ -223,9 +225,14 @@ def test_compile_keymap_args(self):
                 self.xkbcli_compile_keymap.run_command_success(args)
 
     def test_compile_keymap_rmlvo(self):
-        for rmlvo in rmlvos:
+        def run(rmlvo):
             with self.subTest(rmlvo=rmlvo):
```
bluetech: `subTest` is not thread safe, so it's probably better to move the `subTest` to the `future.result()`

test/tool-option-parsing.py
```diff
@@ -223,9 +225,14 @@ def test_compile_keymap_args(self):
...
-                self.xkbcli_compile_keymap.run_command_success(rmlvo)
+                return self.xkbcli_compile_keymap.run_command_success(rmlvo)
+
+        with concurrent.futures.ThreadPoolExecutor() as executor:
```
bluetech, line 439: By default the number of threads is `os.cpu_count() * 5` ([ref](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)). Just guessing, but I think just `max_workers=os.cpu_count()` would be more suitable in this case.

wismill: The default is `min(32, os.cpu_count() + 4)` (it’s an addition, not a multiplication), so in the range `[4, 32]`.

include/xkbcommon/xkbcommon.h
```diff
@@ -281,13 +282,26 @@ typedef uint32_t xkb_led_mask_t;
 #define xkb_keycode_is_legal_x11(key) ((key) >= 8 && (key) <= 255)
 
 /**
- * Names to compile a keymap with, also known as RMLVO.
+ * @defgroup rules-api Rules
+ * Utility functions related to *rules*, which purpose is introduced in:
```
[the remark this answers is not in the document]
wismill: Fixed

test/rules-file-includes.c
wismill: Not that I remember now 😅. It may have been about the RMLVO sanitizing (the private does not do it), but I do not remember… Oh well

wismill:
@bluetech I think we are all set, else I’ll address the issue in a follow-up MR. There are just too many MRs pending to hold.

bam80:
Thank you!
Probably [docs](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/master/README.md?ref_type=heads#building) (4. Test) should be updated too?

wismill:
@bam80 Feel free to propose a MR there. But we need to wait for the xkbcommon release. Keep in mind that while related those are 2 independent projects.

PR #680, opened 2025-03-21
xkbcomp: Fix parser failure on floats

wismill:
Before this commit we used `strtold`, which depends on the locale. But the XKB syntax is fixed and use a period as decimal separator. So ensure the syntax is correct without relying on `strtold` and truncate the result.

I encountered this issue parsing a keymap with a geometry which contains floats.

See #681 for an alternative fix.

TODO:
- [ ] Changelog entry

### DIFF
--- a/src/xkbcomp/scanner.c
+++ b/src/xkbcomp/scanner.c
@@ -5,8 +5,12 @@
 
 #include "config.h"
 
+#include <assert.h>
+
 #include "xkbcomp-priv.h"
 #include "parser-priv.h"
+
+const char DECIMAL_SEPARATOR = '.';
 
 static bool
 number(struct scanner *s, int64_t *out, int *out_tok)
@@ -21,7 +25,7 @@ number(struct scanner *s, int64_t *out, int *out_tok)
     }
     else {
         while (is_digit(scanner_peek(s))) scanner_next(s);
-        is_float = scanner_chr(s, '.');
+        is_float = scanner_chr(s, DECIMAL_SEPARATOR);
         while (is_digit(scanner_peek(s))) scanner_next(s);
     }
     if (s->s + s->pos == start)
@@ -33,16 +37,18 @@ number(struct scanner *s, int64_t *out, int *out_tok)
     if (is_hex)
         x = strtoll(start, &end, 16);
     else if (is_float) {
-        /* The parser currently just ignores floats, so the cast is
-         * fine - the value doesn't matter. */
-        #ifdef __GNUC__
-        #pragma GCC diagnostic push
-        #pragma GCC diagnostic ignored "-Wbad-function-cast"
-        #endif
-        x = (long long int) strtold(start, &end);
-        #ifdef __GNUC__
-        #pragma GCC diagnostic pop
-        #endif
+        /* NOTE: strtold does not work reliably, because it depends on the
+         * locale for e.g. the decimal separator (e.g. a period or a comma).
+         * Since the parser does not use floats, we do not care of the actual
+         * value: we just need to ensure that the syntax is correct. So just
+         * truncate it. */
+        x = strtoll(start, &end, 10);
+        assert(*end == DECIMAL_SEPARATOR);
+        /* Drop the optional decimal part */
+        if (end + 1 < s->s + s->pos)
+            strtoll(end + 1, &end, 10);
+        else
+            end++;
     }
     else
         x = strtoll(start, &end, 10);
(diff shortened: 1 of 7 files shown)


src/xkbcomp/scanner.c
```diff
@@ -5,9 +5,13 @@
 
 #include "config.h"
 
+#include <assert.h>
+
 #include "xkbcomp-priv.h"
 #include "parser-priv.h"
 
+#define DECIMAL_SEPARATOR '.'
```
bluetech, line 13:
```suggestion
const char DECIMAL_SEPARATOR = '.';
```

src/xkbcomp/scanner.c
```diff
@@ -33,16 +37,18 @@ number(struct scanner *s, int64_t *out, int *out_tok)
...
-        x = (long long int) strtold(start, &end);
-        #ifdef __GNUC__
-        #pragma GCC diagnostic pop
-        #endif
+        /* NOTE: strtold does not work reliably, because it depends on the
+         * locale for e.g. the decimal separator (e.g. a period or a comma).
+         * Since the parser does not use floats, we do not care of the actual
+         * value: we just need to ensure that the syntax is correct. So just
+         * truncate it. */
+        x = strtoll(start, &end, 10);
```
bluetech, line 45: I'm not sure why we bothered with the `strtold` at all. Can just set `x = 0` and be done with it...

bluetech (commented):
Good catch. I left a suggestion, let me know what you think.

src/xkbcomp/scanner.c
wismill: We still need to ensure the syntax is correct, even if the value is dropped.

bluetech: Right.

The lexer already ensured the input is `[0-9]*\.[0-9]*`.

Does `strtold` allow `.123` and/or `123.` syntax? In any case I think we can get by with a simple validation, basically len > 1 and (depending on answers to questions above) that first/last chars are not `'.'`.

There's also `00053.53` syntax but I think we should allow it -- probably strtold does too?

wismill: > Does strtold allow .123 and/or 123.

It allows both, but Xorg xkbcomp allow only the second one. While I do not think the first one is conflicting with other XKB syntax, I do think it is bad syntax. So I would avoid supporting if we need a special case.

> There's also 00053.53 syntax but I think we should allow it -- probably strtold does too?

`strtold` does allow it too. This case seems quite legit to me, e.g. for formatting purposes.

wismill: .123 cannot be parsed simply for now, as `.` is the `DOT` token. So even currently,  we cannot parse it using `strtold` because we simply never get there!

src/xkbcomp/scanner.c
wismill: I *really* have to stop using macros for every global var! Fixed

src/xkbcomp/scanner.c
wismill: I update the MR to accept `123.` syntax (no decimal part).

wismill:
I realized that `strtoll` & co still need a NULL-terminated string. For example, we get a stack-buffer-overflow error in Asan if we run the following test:

```c
    const char boom[115] =
                "xkb_keymap {\n"
                "  xkb_keycodes { };\n"
                "  xkb_types { };\n"
                "  xkb_compat { };\n"
                "  xkb_symbols { };\n"
                "  xkb_geometry {\n"
                "    width=1";
    struct xkb_keymap *keymap =
            test_compile_buffer(ctx, boom, 115);
```

I manually truncated `boom` to not include the NULL character. I really wonder if we should use our own functions to safely parse numbers in a buffer.

wismill:
Added tests

wismill:
@bluetech Tests are now run with the locale `fr_FR.UTF-8` in the Linux runner.

wismill:
Oh my! I realized that `xkbcomp` allows using floating-point numbers *everywhere* an int could be used! It also supports the exponent syntax, so such wonder is allowed:

```c
key <AE01> { symbols[0.3e1]= [1, exclam] };
```

I think it is safe to assume that no floating-point number is used in for such use case in real world. At least not in xkeyboard-config.

xkeyboard-config also does not use the exponent syntax, so I think we can just avoid to implement its support.

wismill:
Rebased & added doc about differences with Xorg tools.

PR #681, opened 2025-03-23
xkbcomp: Parse floats independently of the locale

wismill:
Before this commit we used `strtold`, which depends on the locale. But the XKB syntax is fixed and use a period as decimal separator. So ensure the syntax is correct without relying on `strtold`, using C++ `std::from_chars`.

This is an alternative to #680.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,6 +1,6 @@
 project(
     'libxkbcommon',
-    'c',
+    'c', 'cpp',
     version: '1.8.1',
     default_options: [
         'c_std=c11',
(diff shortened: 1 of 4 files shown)


bluetech (commented):
Hmm bringing C++ stdlib into the mix is a big hammer for this. So I prefer #680 :)

wismill:
I know, but I am quite jealous we do not have such basic function in the C land. I wonder if we have other bits that depend on the locale. It may be a good idea to run the test with a locale different than `en_US.UTF-8` just to check it.

bluetech:
To be fair, most libcs (even windows) do have `strtold_l` but it's not standard.

bluetech:
> I wonder if we have other bits that depend on the locale. It may be a good idea to run the test with a locale different than en_US.UTF-8 just to check it.

We've had some issues with it in the past, when we were using `ctype.h`. But then we had an issue with Turkish `i` and changed to own ascii variants and stopped using `ctype.h`. Other than that I can't think of any locale dependence (except the explicit ones in Compose). But the idea of testing with a different locale is a good one. The main difficulty is that usually AFAIK only the used locale is generated so to use other locales, need to enable it and run `locale-gen`, so the test will need to be conditional.

PR #682, opened 2025-03-25
xkbcomp: Fix whichGroupState serialization

wismill:
This indicator field was previously looked up in the wrong table, resulting the erroneous serialization `(null)`.

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -282,7 +282,8 @@ ModMaskText(struct xkb_context *ctx, enum mod_type type,
 }
 
 const char *
-LedStateMaskText(struct xkb_context *ctx, enum xkb_state_component mask)
+LedStateMaskText(struct xkb_context *ctx, const LookupEntry *lookup,
+                 enum xkb_state_component mask)
 {
     char buf[1024];
     size_t pos = 0;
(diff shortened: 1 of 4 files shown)


wismill:
@whot @bluetech these kind of errors are easy to fix but difficult to trigger, so I wonder how we could test this more systematically.

I added an `assert` as a very basic check.

src/text.c
```diff
@@ -298,9 +299,10 @@ LedStateMaskText(struct xkb_context *ctx, enum xkb_state_component mask)
 
         mask &= ~(1u << i);
 
+        assert(LookupValue(lookup, 1u << i) != NULL);
```
bluetech, line 302: Maybe store the result in a var to avoid doing the lookup twice?

bluetech, line 302: Is the assert safe? Is there any way we can get an unknown value from user input here (maybe xkbcommon-x11)?

bluetech (approved):
Good catch. If I'm reading correctly we inherited this bug from libxkbfile: https://gitlab.freedesktop.org/xorg/lib/libxkbfile/-/blob/3ba601efc462d0c39598aab8216580cdcf025c6a/src/xkbout.c#L270-286

src/text.c
wismill: > maybe xkbcommon-x11

That would be it, because this does not use the same lookup table.

Well, there is another one: the field can be legitimately `0`, so the caller has to ensure to use the `*MaskText` functions with a non-`0` value present in the lookup table.

It would make sense to change the definition of `LookupEntry` like so:

```diff
typedef struct {
     const char *name;
-    uint32_t value;
+    uint32_t xkbcommon;
+    uint32_t x11;
} LookupEntry;
```

then use the *same lookup table* in both our parser and in the X11 API. Then we can have a check for `NULL` in the dump function, so that we exit with a useful error and never print garbage. What do you think?

wismill:
> If I'm reading correctly we inherited this bug from libxkbfile: https://gitlab.freedesktop.org/xorg/lib/libxkbfile/-/blob/3ba601efc462d0c39598aab8216580cdcf025c6a/src/xkbout.c#L270-286

@bluetech They use the same table for both fields contrary to use, so it’s not exposed. But I much prefer our lookup tables that explicitly link text and numeric values than their implicit mapping.

src/text.c
wismill: Me too tired. Fixed

Issue #500, opened 2024-08-22
RMLVO: support for applying option on specific layout index


wismill:
See the corresponding [`xkeyboard-config` issue](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/56).

The idea would be to be able to have a suffix per option, such as it applies only to some layout indexes.

Some preliminary ideas:
- `,` cannot be used as a suffix delimeter, as it is the option separator.
- `:` should not be used as a suffix delimeter, e.g. `eurosign:e:2`, because it is ambiguous.
- `@` seems an interesting candidate. E.g.: `eurosign:e@2`
- We should embrace #362 and introduce `@all`, `@first`, etc. as well.
- We need DE devs to be implied, if we want a UI to be developed.
- There should be only very limited breakage, as it is the “compiled” keymap that is shared by the compositor with the apps, not the RMLVO values.

whot:
Or make this a proper builder API so we don't have to worry about magic syntax. Something like:

```c
rmlvo = xkb_rmlvo_new();

layout = xkb_rmlvo_layout_new("us");
xkb_rmlvo_layout_set_variant(layout, "dvorak");
xkb_rmlvo_layout_add_option("banana");
xkb_rmlvo_layout_add_option("orange");
xkb_rmlvo_add_layout(layout);

// repeat for layout 2 if need be

keymap = xkb_keymamp_new_from_rmlvo(context, rmlvo, flags);
```
or, for those who really like varargs:
```c
layout = xkb_rmlvo_layout_new("us",
        XBK_RMLVO_VARIANT, "dvorak",
        XKB_RMLVO_OPTION, "banana",
        XKB_RMLVO_OPTION, "banana",
        XKB_RMLVO_END);
```
sd-bus uses the latter extensively but in your case I think the first approach is better.


No backwards compat issue either since it's a new API :)

wismill:
I like this idea (without varargs)! But the new suffix syntax would still be useful for CLI, wouldn't it?

We need to figure out how to handle this in `xkeyboard-config`. The standard way is to duplicate for each layout index. #362 may come handy. Or we may implement an expection in rules resolving for the new API.

whot:
> I like this idea (without varargs)! But the new suffix syntax would still be useful for CLI, wouldn't it?

Might be better to dream up a new syntax. after all the xkbcommon cli isn't quite as wide-spread through config files everywhere as e.g. setxkbmap's.

[...]

PR #683, opened 2025-03-25
Enable merge mode “replace” in include statements

wismill:
Previously only the merge modes “override” and “augment” were available in include statements, using the prefix `+` and `|` respectively. While one can use the `replace` include statement in keymap files, *rules* files have no way to express the *replace* mode.

This commit enables the merge mode “replace” using the prefix `^`. This prefix was chosen because it due to its similarity with the `XOR` bit operator, which convey mutual exclusion.

[Related discussion](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/525) in `xkeyboard-config` project.

EDIT: changed from `!` to `^` prefix

## Discussion

Other ASCII symbols candidates:
- `!` conveys some kind of higher precedence, akin to CSS `!important`. But it conflicts with the section header `!`, which is a token in the current parser. It would require special handling, not worth it. It also convey the meaning of negation, which is confusing.
- `&`, which has the advantage of not corresponding to a token in the rules parser. `^` seems however to stand out more and it is less likely to trigger erroneous comparison with `|` and `&` bit operators.

Discarded ASCII symbols:
- `=`: another interesting candidate to convey the replace/unchanged info, but confusing as it is also used for assignment.
- `#`: probably not a good idea for its usual role as “number sign”.
- `*`: already used as wild card; not a good idea to overload it.
- `~`: do not convey much.
- `<`, `>`: interesting for the direction they point, but would be better paired with another character, else they are difficult to interpret.
- `@`: might have a use in #500.

While I like `!` better, we could go for `&` or another ASCII character if we foresee a use for the (`!`, `?`) pair in the include statements or the qualifier (the suffix starting with `:`, currently used for group index). Note that `!`, `?` and `+` are new wild cards for rules, proposed in #685.

@bluetech @whot @mahkoh @fooishbar

### DIFF
--- a/changes/api/+replace-include.feature.md
+++ b/changes/api/+replace-include.feature.md
@@ -1,0 +1,3 @@
+Enable using the merge mode *replace* in include statements using the prefix `^`,
+such as: `include "a^b"`. The main motivation is to enable this new syntax in
+*rules*, which previously could not handle this merge mode.
--- a/doc/compatibility.md
+++ b/doc/compatibility.md
@@ -44,6 +44,9 @@ unused in the standard dataset.
 - xkbcommon has stronger type checks, so floating-point numbers cannot
   be used where an integer is expected.
 - exponent syntax for floating-point numbers is not supported.
+- support the merge mode *replace* in include statements via the
+  prefix `^`, in addition to the standard *merge* `|` and *override*
+  `+` modes.
 
 On the other hand, some features and extensions were added.
 
@@ -61,6 +64,8 @@ On the other hand, some features and extensions were added.
 ### Additions
 
 - `! include` statement.
+- Support the merge mode *replace* via the prefix `^`, in addition to
+  the standard *merge* `|` and *override* `+` modes.
 - Additional wild cards:
   - `<none>` matches *empty* value;
   - `<some>` matches *non-empty* value in *every* context.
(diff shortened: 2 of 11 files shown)


wismill:
Now that #685 is pending, I wonder if `&` as *replace* merge mode is a good option after all, to avoid confusion with the `!` wild card and make a pair with the `|` *augment* merge mode.

mahkoh:
I'd prefer not to use ! since it is separated from the following token in my lexer.

wismill:
@mahkoh Same for xkbcommon, but I found a trick. Could you elaborate what character you would use?

mahkoh:
I don't think there is a strong need for a `replace` character in xkb since any RMLVO expansion can already be written as

```xkb
include "a"
replace "b"
```

instead of 

```xkb
include "a!b"
```

So the question is what syntax to use in RMLVO files. If I had to choose a character it would be ^ but maybe it would be better to use a completely new syntax similar to [verbose regex] (https://docs.rs/regex/1.11.1/regex/index.html#example-verbose-mode).

wismill:
> RMLVO files

@mahkoh I assume that you mean *rules* files?

> I don't think there is a strong need for a `replace` character in xkb

> So the question is what syntax to use in RMLVO files. 

Yes, the main motivation is to make it available in the rules files.

> If I had to choose a character it would be ^ 

Good suggestion, after analyzing it again I like the similarity with the mutual exclusive `XOR`. I went ahead and updated the MR.

> but maybe it would be better to use a completely new syntax similar to [verbose regex](https://docs.rs/regex/1.11.1/regex/index.html#example-verbose-mode).

I think we want to keep it simple, as we discussed to use a *builder* in #546 for more complex cases. Could you elaborate with examples?

PR #685, opened 2025-03-26
rules: Added extended wild cards <none>, <some> and <any>

wismill:
Added the following wild cards to the rules file syntax, in addition to the current `*` legacy wild card:

EDIT: *current* proposal:

- `<none>`: Match *empty* value.
- `<some>`: Match *non-empty* value.
- `<any>`: Match *any* (optionally empty) value. Its behavior does not depend on the context, contrary to the legacy wild card `*`.

The verbose wild cards are preferred to single characters:
- More intuitive: self-explanatory.
- Does not steal syntax from other token.
- Extensible syntax, should we need it.

This will enable writing much simpler rules, see [!764] for an example of tricky rules in the `xkeyboard-config` project, that would benefit from the new wild cards. The extended rules support is tracked in [this issue](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/524).

[!764]: https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/764

---

EDIT: *previous* proposal:
- `!`: Match *empty* value.
- `+`: Match *non-empty* value.
- `?`: Match *any* (optionally empty) value. Its behavior does not depend on the context, contrary to the legacy wild card `*`.

The characters were chosen for their similarity with the corresponding syntax of regular expressions (negative assertion & quantifiers), in line with `*`.

Note that the current parser disallows using `!` in the first MLVO component of a rule, because it would match a new rule set, an include statement or a group definition. This is unfortunate, but:
- rules are usually written respecting the MLVO components order;
- model and layout are normally non-empty if proper defaults are set;
- variant is normally used paired with layout, in the LV order;
- option is usually not tested for empty values;
- one can always use model as a first component and use the wild card `*` or `?` to always match, enabling `!` in the further MLVO components. So this limitation has a very low probability to cause issues in real use cases.

@bluetech @whot @mahkoh @fooishbar

### DIFF
--- a/changes/api/+extended-rules-wild-cards.feature.md
+++ b/changes/api/+extended-rules-wild-cards.feature.md
@@ -1,0 +1,6 @@
+Added the following *wild cards* to the **rules** file syntax, in addition to
+the current `*` legacy wild card:
+- `<none>`: Match *empty* value.
+- `<some>`: Match *non-empty* value.
+- `<any>`: Match *any* (optionally empty) value. Its behavior does not depend on
+  the context, contrary to the legacy wild card `*`.
--- a/doc/compatibility.md
+++ b/doc/compatibility.md
@@ -52,7 +52,18 @@ On the other hand, some features and extensions were added.
 - multiple keysyms per level
   + such levels are ignored by x11/xkbcomp.
 - key names (e.g. `<AE11>`) can be longer than 4 characters.
-- `! include` statement for rules files.
+
+## Rules support {#rules-support}
+
+### Additions
+
+- `! include` statement.
+- Additional wild cards:
+  - `<none>` matches *empty* value;
+  - `<some>` matches *non-empty* value in *every* context.
+  - `<any>` matches *optionally empty* value in *every* context, contrary to the
+    legacy `*` wild card which does not match empty values for layout and
+    variant;
 
 ## Compose support {#compose-support}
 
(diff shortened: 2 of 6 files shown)


mahkoh:
>Note that the current parser disallows using ! in the first MLVO component of a rule, because it would match a new rule set, an include statement or a group definition.

Why not just use another character?

wismill:
> Why not just use another character?

@mahkoh Please make a proposal 😄. The only other candidate I have in mind is `-`, but apart from the issue with the leading `!` I much prefer the triple (`!`, `+`, `?`) for its similarity with regex.

mahkoh:
How about

- `<any>`
- `<none>`
- `<anyOrNone>`

wismill:
@mahkoh I like your suggestion: we are not limited to single character indeed, and this syntax can be extended further, should we have the need.

However, I find `<any>` [confusing](https://www.merriam-webster.com/dictionary/any) in your proposal. What about:
- `<none>`
- `<some>`
- `<any>` : some or none

wismill:
Maintainers of xkeyboard-config are happy with this feature, so let’s merge. Actual deployment of this feature is discussed [this xkeyboard-config issue](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/524#note_2839162).

PR #687, opened 2025-03-26
Keysyms: parsing improvements

wismill:


src/xkbcomp/ast.h
```diff
@@ -140,6 +141,11 @@ typedef struct {
     xkb_atom_t key_name;
 } ExprKeyName;
 
+typedef struct {
+    ParseCommon common;
+    xkb_keysym_t keysym;
```
bluetech, line 146: Re. the "error already printed" comment in the code below, might be nice to also mention it in a comment here, e.g.

If the keysym is invalid, NoSymbol is stored in error recovery. An error was already logged.

bluetech (commented):
The `ExprResolveKeySym` removal is great!

src/xkbcomp/ast.h
wismill: I gave it a try, see #733.

PR #692, opened 2025-03-29
registry: Exclude the trailing NUL byte from embedded DTD

xry111:
Since libxml2-2.14.0, the trailing NUL byte is considered an invalid encoding error.

Link: https://gitlab.gnome.org/GNOME/libxml2/-/issues/886


wismill:
Thanks for reporting! This must be a recent change, because I did not have this issue in #675.

Fixed in #693

wismill:
libxml2 2.14 was released 2 days ago, so I think we can skip a minor release and wait a bit for pending MRs to land.

PR #699, opened 2025-03-31
Support empty symbols and actions lists

### LINKED ISSUE #565: xkbcomp: Some lists are not allowed to be empty

wismill:
Enable empty lists in more places:
- (Empty `interpret`)
- (Empty key `type`)
- (Empty `indicator`)
- Empty list of keysyms and actions: `{}`

Motivations:
- Follow the principle of least astonishment;
- Ensure consistency;
- Enhance the use of custom defaults;
- Facilitate the tests.

Fixes #565

NOTE: Using `{}` for both actions and keysyms introduces an ambiguity, which is resolved when reaching the first non `{}` in the list. I managed to modify the parser the handle it, but I wonder if we can do better.

TODO:
- [x] Rebase on #701
- [x] Rebase on #704
- [x] Moaar tests
- [x] Changelog

### DIFF
--- a/changes/api/+empty-keysym-action-lists.feature.md
+++ b/changes/api/+empty-keysym-action-lists.feature.md
@@ -1,0 +1,2 @@
+Added support for actions and keysyms level list of length 0 and 1: respectively
+`{}` and `{a}`. Example: `key <A> { [{}, {a}, {a, A}] };`.
--- a/doc/keymap-format-text-v1.md
+++ b/doc/keymap-format-text-v1.md
@@ -1656,14 +1656,38 @@ There are some *limitations* with this extension:
   - `SetMods` + `LockMods`: *error*
   - `SetMods` + `LockGroup`: ok
 - Multiple actions are only supported from version 1.8.0.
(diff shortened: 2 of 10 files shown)


wismill:
Pending of #701.

mahkoh:
> Now that we can parse e.g. `indicator "x" {};`, I wonder if we should also allow `indicator "x";`? I am fine _without_ this simpler syntax because it would be really niche use cases, just thinking out loud about the strange `xxx {};` syntax.

I've been thinking about the opposite and allowing omitting the `;` after most `}`.

wismill:
> I've been thinking about the opposite and allowing omitting the `;` after most `}`.

Haha those too!

> most `}`

Which `};` cannot be converted to `}`?

PR #714, opened 2025-04-07
[Experiment] Test the Unicode mess

wismill:
Added tests to check the Unicode keysyms support and compare with X11.

This is meant to highlight differences and shortcomings.

I created a MR to make it more visible.

**Do not merge!**


wismill:
Fixed in #717.

PR #731, opened 2025-04-27
doc/keymap-format-text-v1: fix inaccurate statement in Keysym glossary entry

bluetech:
Just a little thing I noticed -- not all keysyms have a name.


wismill (approved):
Good point!

PR #736, opened 2025-04-28
test: Fix `main()` function prototype

bgilbert:
Fix compilation error when building for Windows with clang-cl:

    ../subprojects/libxkbcommon-xkbcommon-1.9.0/test/keysym-unicode.c:11:1: error: return type defaults to 'int' [-Wimplicit-int]
    ../subprojects/libxkbcommon-xkbcommon-1.9.0/test/keysym-unicode.c:11:1: warning: function declaration isn't a prototype [-Wstrict-prototypes]

### DIFF
--- a/test/keysym-unicode.c
+++ b/test/keysym-unicode.c
@@ -8,7 +8,8 @@
 
 #ifdef _WIN32
 
-main()
+int
+main(void)
 {
     test_init();
     return SKIP_TEST;


wismill:
Thanks! Strange that our own CI did not catch this though. What am I missing?

bgilbert:
C89 allowed omitting the return type from function declarations.  Apparently MSVC warns about it but doesn't fail the build by default:

```
D:\a\libxkbcommon\libxkbcommon\test\keysym-unicode.c(11,1): warning C4431: missing type specifier - int assumed. Note: C no longer supports default-int [D:\a\libxkbcommon\libxkbcommon\build\test-keysym-unicode@exe.vcxproj]
```

Your Windows CI builds with MSVC but not clang-cl, and on other platforms this code is ifdef'd out.

wismill:
I am not much familiar with MSVC, are there flags that we could turn on to make these warnings actual errors?

bgilbert:
For this specific problem, it looks like you could set `/we4431`.  More broadly, you might consider failing CI on all warnings with `meson setup --werror`, ideally on all platforms.  There are currently a number of other compiler warnings, though, which might take some effort to clean up.

wismill:
Thanks!

>  There are currently a number of other compiler warnings, though, which might take some effort to clean up.

Yes, we have #131 for tackling those. I doubt that they may trigger any bug, but we will fix them eventually (already fixed the most likely ones). Wrong integer casts are nasty.

PR #739, opened 2025-05-02
Shortcut tweak v2

wismill:

Tweak keyboard shortcuts by adding the option to switch to a different group when some modifiers are active (typically `Control`, `Alt` or `Super`). The target group can be set for each group, although the most common use case would probably be the same target for all layouts.

This can be alternatively activated with the environment variables `XKB_UNSTABLE_EXPERIMENTAL_SHORTCUT_MASK` and `XKB_UNSTABLE_EXPERIMENTAL_SHORTCUT_TARGET_LAYOUTS` [^1]. This enables to use the new feature even if the DE does not implement a UI to configure it. While the names suggests that this setting method is temporary, I am unsure we should ship it: I do not want to see this setting method broadly used so that we cannot change/remove it later. It is more a mean for ~~early adopters~~ hackers to test the feature by installing their own local version of libxkbcommon.

I need help to rephrase the doc: this is more a hack than a tweak, as we are breaking the XKB protocol by not following its computation rule of the effective group.

TODO:
- [ ] Rebase on #730
- [ ] Improve Doc
- [ ] Remove environment variable support?
- [ ] Log entry

[^1]: Actually the env variable will also activate this processing in clients for now, but this is a no-op, as the layout is set by the compositor, which would have already re-target it for shortcuts, if necessary.

---

This is a refactor of #523, which got out of sync and used a more complex (broken?) approach.


wismill:
@bluetech @whot @mahkoh While this could be implemented in the compositors, it seems that such hack would be better handled in the lib. Also, this is a bit too hacky for my taste. Do you have a better idea how to handle this?

@bam80 and @Jules-Bertholet may want to try this up-to-date PR.

saeedmfarahani:
any updates?
could you please specify the value of following environment variables what should be?
XKB_UNSTABLE_EXPERIMENTAL_SHORTCUT_MASK
XKB_UNSTABLE_EXPERIMENTAL_SHORTCUT_TARGET_LAYOUTS
thank you!

wismill:
@saeedmfarahani No update. It is unclear if the best route is to implement this in the lib or to let each compositor implements its own strategy. I asked to some but did not get an answer.

This approach feels too hacky anyway, so I will have to rethink a better solution.

PR #735, opened 2025-04-28
doc: Introduce a FAQ page with how to get the vmod → rmod mapping

wismill:
Apart from the traditional questions:
- Document how to get the vmod → rmod mapping

  Some applications may need it to interface with legacy code.

Closes #732

Issue #640, opened 2025-02-03
Optional pretty-printing


wismill:
Currently we always pretty-print the keymap when serializing it, but it means more work to generate (computing paddings, human-friendly mask, etc.) and then more work to parse in the client.

We could make this optional:
- Use a minified[^1] keymap for serialization by default in the API, with optional pretty-print using a flag.
- Use a pretty-printed keymap for all tools by default and a tool flag to turn it off.

[^1]: Keep the newlines, we are no monsters

Issue #578, opened 2025-01-05
Incorrect handling of unbound virtual modifiers in key types


mahkoh:
Consider the following map:

```xkb
xkb_keymap {
    xkb_keycodes {
        <a> = 38;
        <leftshift> = 50;
    };
    xkb_types {
        virtual_modifiers Bound = Shift, Unbound;

        type "X" {
            modifiers = Bound+Unbound;
            map[Bound+Unbound] = Level1;
            map[Bound] = Level2;
        };
    };
    xkb_compat { };
    xkb_symbols {
        key <a> {
            type = "X",
            [ a, A ]
        };
        key <leftshift> {
            [ SetMods(mods = Shift) ]
        };
    };
};
```

The specification requires that 

>Map entries which specify unbound virtual modifiers (see [Inactive Modifier Definitions](https://www.x.org/releases/X11R7.7/doc/kbproto/xkbproto.html#Inactive_Modifier_Definitions)) are not considered
>[...]
>When it is essential to distinguish between OneThing and Another, XKB considers only those modifier definitions for which **all** virtual modifiers are bound.

Since `Unbound` is not bound, the line

```xkb
            map[Bound+Unbound] = Level1;
```

should be ignored.

But xkbcommon does not ignore it and maps Shift+a to Level1.

wismill:
Good catch! Although with #450 landing soon, this will work as expected. So we should just wait for #450 to fix it.

wismill:
#450 is still pending, so let’s fix this now.

PR #745, opened 2025-05-05
state: Fix virtual modifiers with non-real mod mapping

wismill:
Currently there are 2 issues with the handling of virtual modifiers in the keyboard state:
1. We assume that the input modifiers masks encode the indexes of all the modifiers of the keymap, but this is true only for the *real* modifiers (at least in xkbcommon and X11). Indeed, since the virtual modifiers *indexes* are implementation-specific, the input modifier masks merely *encode* the modifiers via their *mapping*.

   Consider the following keymap:

   ```c
   xkb_keymap {
     xkb_compat { virtual_modifiers M1 = 0x100; };
     xkb_types  { virtual_modifiers M2 = 0x200; };
   };
   ```

   Now to illustrate, consider the following 2 implementation variants of libxkbcommon (assuming indexes 0-7 are the usual real modifiers):
   1. Process `xkb_compat` then `xkb_types`. M1 and M2 have the respective indexes 8 and 9 and map to themselves (with the current assumption about mask denotation).
   2. Process `xkb_types` then `xkb_compat`. M1 and M2 have the respective indexes 9 and 8 and map to each other.

   With the current `xkb_state_update_mask`, implementation 2 will swap M1 and M2 (compared to impl. 1) at each update! Indeed, we can see that `xkb_state_serialize_mods` doesn’t roundtrip via `xkb_state_update_mask`.
2. We assume that modifier masks use only bits denoting modifiers in the keymap, but when parsing the keymap we accept explicit virtual modifiers mapping of arbitrary values.

   E.g. if `M1` is the only virtual modifier and it is defined by:

   ```c
   virtual_modifiers M1 = 0x80000000; // 1 << (32 - 1)
   ```

   then the 32th bit of a modifier mask input does *not* denote the 32th virtual modifier of the keymap, but merely the encoding of the mapping of `M1`.

   So when calling `xkb_state_update_mask`, we may discard some bits of the modifiers masks and end up with an incorrect state.

These 2 issues may break interoperability with other implementations of XKB (e.g. kbvm) and make pure virtual modifiers handling fragile.

We introduce the notion of *canonical state modifier mask*: the mask with the smallest population count that denotes all bits used to encode the modifiers in the keyboard state. It is equal to the bitwise OR of real modifiers mask and all the virtual modifiers mappings.

This commit fixes the 2 issues by making *weaker* assumptions about the input modifier masks:
1. Modifiers may map to arbitrary values, not only real modifiers.
2. Input modifier masks merely encode modifiers via their *mapping*:
   - *real* modifiers map to themselves;
   - *virtual* modifiers map to the bitwise OR of their *explicit* mapping (via `virtual_modifiers`) and their *implicit* mapping (via keys’ real and virtual modmaps).
   - modifiers indexes are implementation-specific.

   Since the implementation before this commit also resolved virtual modifiers to their mappings, we continue doing so, but using only the bits that are *not* set in the canonical state modifier mask, so that we enable roundtrip of `xkb_state_serialize_mods` via
   `xkb_state_update_mask`.
3. Input modifier masks do not denote modifiers indexes (apart from real modifiers), so it is safe to discard only the bits that are not set in the canonical state modifier mask.

This MR is necessary in order to implement #450.

@whot @bluetech @mahkoh

### DIFF
--- a/changes/api/+state-mod-mask-handling.breaking.md
+++ b/changes/api/+state-mod-mask-handling.breaking.md
@@ -1,0 +1,10 @@
+*Modifiers masks* handling has been refactored to properly handle virtual
+modifiers. Modifier masks are now always considered as an *opaque encoding* of
+the modifiers state:
+- Modifiers masks should not be interpreted by other means than the provided API.
+  In particular, one should not assume that modifiers masks always denote the
+  modifiers *indexes* of the keymap.
+- It enables using virtual modifiers with arbitrary mappings. E.g. one can now
+  reliably create virtual modifiers without relying on the legacy X11 mechanism,
+  that requires a careful use of keys’ real and virtual modmaps.
+- It enables *interoperability* with further implementations of XKB.
--- a/src/keymap-priv.c
+++ b/src/keymap-priv.c
@@ -36,6 +36,7 @@ update_builtin_keymap_fields(struct xkb_keymap *keymap)
         keymap->mods.mods[i].type = MOD_REAL;
     }
     keymap->mods.num_mods = ARRAY_SIZE(builtin_mods);
+    keymap->canonical_state_mask = MOD_REAL_MASK_ALL;
 }
 
 struct xkb_keymap *
--- a/src/keymap.h
+++ b/src/keymap.h
@@ -419,7 +419,37 @@ struct xkb_keymap {
     unsigned int num_sym_interprets;
     struct xkb_sym_interpret *sym_interprets;
 
+    /**
+     * Modifiers configuration.
+     * This is *internal* to the keymap; other implementations may use different
+     * virtual modifiers indexes. Ours depends on:
+     *   1. the order of the parsing of the keymap components;
+     *   2. the order of the virtual modifiers declarations;
+     */
     struct xkb_mod_set mods;
+    /**
+     * Modifier mask of the *canonical* state, i.e. the mask with the *smallest*
+     * population count that denotes all bits used to encode the modifiers in
+     * the keyboard state. It is equal to the bitwise OR of *real* modifiers and
+     * all *virtual* modifiers mappings.
+     *
+     * [WARNING] The bits that do not correspond to *real* modifiers should
+     * *not* be interpreted as corresponding to indexes of virtual modifiers of
+     * the keymap. Indeed, one may use explicit vmod mapping with an arbitrary
+     * value.
+     *
+     * E.g. if M1 is the only vmod and it is defined by:
+     *
+     *     virtual_modifiers M1=0x80000000; // 1 << (32 - 1)
+     *
+     * then the 32th bit of a modifier mask input does *not* denote the 32th
+     * virtual modifier of the keymap, but merely the encoding of the mapping of
+     * M1.
+     *
+     * In the API, any input mask should be preprocessed to resolve the bits
+     * that do not match the canonical mask.
+     */
+    xkb_mod_mask_t canonical_state_mask;
 
     /* This field has 2 uses:
      * • During parsing: Expected layouts count after RMLVO resolution, if any;
@@ -544,7 +574,7 @@ XkbWrapGroupIntoRange(int32_t group,
                       enum xkb_range_exceed_type out_of_range_group_action,
                       xkb_layout_index_t out_of_range_group_number);
 
-xkb_mod_mask_t
+XKB_EXPORT_PRIVATE xkb_mod_mask_t
 mod_mask_get_effective(struct xkb_keymap *keymap, xkb_mod_mask_t mods);
 
 struct xkb_level *
--- a/src/state.c
+++ b/src/state.c
@@ -885,6 +885,27 @@ xkb_state_update_key(struct xkb_state *state, xkb_keycode_t kc,
 }
 
 /**
+ * Compute the resolved effective mask of an arbitrary input.
+ *
+ * Contrary to `mod_mask_get_effective`, it resolves only modifiers not present
+ * in the canonical mask, so that it enables `xkb_state_serialize_mods` to
+ * round trip via `xkb_state_update_mask`.
+ */
+static inline xkb_mod_mask_t
+resolve_to_canonical_mods(struct xkb_keymap *keymap, xkb_mod_mask_t mods)
+{
+    return
+        /*
+         * Keep canonical modifier mask.
+         * It contains either real modifiers or canonical virtual modifiers.
+         */
+        (mods & keymap->canonical_state_mask) |
+        /* Resolve other modifiers */
+        mod_mask_get_effective(keymap,
+                               mods & ~keymap->canonical_state_mask);
+}
+
+/**
  * Updates the state from a set of explicit masks as gained from
  * xkb_state_serialize_mods and xkb_state_serialize_groups.  As noted in the
  * documentation for these functions in xkbcommon.h, this round-trip is
@@ -900,20 +921,10 @@ xkb_state_update_mask(struct xkb_state *state,
                       xkb_layout_index_t latched_group,
                       xkb_layout_index_t locked_group)
 {
-    struct state_components prev_components;
-    xkb_mod_mask_t mask;
+    const struct state_components prev_components = state->components;
 
-    prev_components = state->components;
-
-    /* Only include modifiers which exist in the keymap. */
-    mask = (xkb_mod_mask_t) ((UINT64_C(1) << xkb_keymap_num_mods(state->keymap))
-         - UINT64_C(1));
-
-    state->components.base_mods = base_mods & mask;
-    state->components.latched_mods = latched_mods & mask;
-    state->components.locked_mods = locked_mods & mask;
-
-    /* Make sure the mods are fully resolved - since we get arbitrary
+    /*
+     * Make sure the mods are fully resolved - since we get arbitrary
      * input, they might not be.
      *
      * It might seem more reasonable to do this only for components.mods
@@ -925,14 +936,13 @@ xkb_state_update_mask(struct xkb_state *state,
      * if a vmod is depressed, its mappings are depressed with it; so we're
      * expected to do the same here.  Also, LEDs (usually) look if a real
      * mod is locked, not just effective; otherwise it won't be lit.
-     *
-     * We OR here because mod_mask_get_effective() drops vmods. */
-    state->components.base_mods |=
-        mod_mask_get_effective(state->keymap, state->components.base_mods);
-    state->components.latched_mods |=
-        mod_mask_get_effective(state->keymap, state->components.latched_mods);
-    state->components.locked_mods |=
-        mod_mask_get_effective(state->keymap, state->components.locked_mods);
+     */
+    state->components.base_mods =
+        resolve_to_canonical_mods(state->keymap, base_mods);
+    state->components.latched_mods =
+        resolve_to_canonical_mods(state->keymap, latched_mods);
+    state->components.locked_mods =
+        resolve_to_canonical_mods(state->keymap, locked_mods);
 
     static_assert(XKB_MAX_GROUPS < INT32_MAX, "Max groups don't fit");
     state->components.base_group = (int32_t) base_group;
@@ -1247,13 +1257,12 @@ xkb_state_serialize_layout(struct xkb_state *state,
 xkb_mod_mask_t
 mod_mask_get_effective(struct xkb_keymap *keymap, xkb_mod_mask_t mods)
 {
+    /* Initialize the effective mask with its corresponding real mods. */
+    xkb_mod_mask_t mask = mods & MOD_REAL_MASK_ALL;
+
+    /* Resolve the virtual modifiers */
     const struct xkb_mod *mod;
     xkb_mod_index_t i;
-    xkb_mod_mask_t mask;
-
-    /* The effective mask is only real mods for now. */
-    mask = mods & MOD_REAL_MASK_ALL;
-
     xkb_vmods_enumerate(i, mod, &keymap->mods)
         if (mods & (UINT32_C(1) << i))
             mask |= mod->mapping;
@@ -1571,7 +1580,7 @@ xkb_state_mod_mask_remove_consumed(struct xkb_state *state, xkb_keycode_t kc,
     if (!key)
         return 0;
 
-    return mod_mask_get_effective(state->keymap, mask) &
+    return resolve_to_canonical_mods(state->keymap, mask) &
            ~key_get_consumed(state, key, XKB_CONSUMED_MODE_XKB);
 }
 
--- a/src/xkbcomp/keymap.c
+++ b/src/xkbcomp/keymap.c
@@ -327,6 +327,13 @@ UpdateDerivedKeymapFields(struct xkb_keymap *keymap)
             if (key->vmodmap & (UINT32_C(1) << i))
                 mod->mapping |= key->modmap;
 
+    /* Update the canonical modifiers state mask */
+    assert(keymap->canonical_state_mask == MOD_REAL_MASK_ALL);
+    xkb_mod_mask_t extra_canonical_mods = 0;
+    xkb_vmods_enumerate(i, mod, &keymap->mods)
+        extra_canonical_mods |= mod->mapping;
+    keymap->canonical_state_mask |= extra_canonical_mods;
+
     /* Now update the level masks for all the types to reflect the vmods. */
     for (i = 0; i < keymap->num_types; i++) {
         ComputeEffectiveMask(keymap, &keymap->types[i].mods);
--- a/test/modifiers.c
+++ b/test/modifiers.c
@@ -10,27 +10,21 @@
 
 #include "test.h"
 #include "keymap.h"
-
-/* Standard real modifiers indexes */
-#define ShiftIndex   0
-#define LockIndex    1
-#define ControlIndex 2
-#define Mod1Index    3
-#define Mod2Index    4
-#define Mod3Index    5
-#define Mod4Index    6
-#define Mod5Index    7
+#include "utils.h"
+#include "xkbcommon/xkbcommon.h"
 
 /* Standard real modifier masks */
-#define ShiftMask   (UINT32_C(1) << ShiftIndex)
-#define LockMask    (UINT32_C(1) << LockIndex)
-#define ControlMask (UINT32_C(1) << ControlIndex)
-#define Mod1Mask    (UINT32_C(1) << Mod1Index)
-#define Mod2Mask    (UINT32_C(1) << Mod2Index)
-#define Mod3Mask    (UINT32_C(1) << Mod3Index)
-#define Mod4Mask    (UINT32_C(1) << Mod4Index)
-#define Mod5Mask    (UINT32_C(1) << Mod5Index)
-#define NoModifier  0
+enum real_mod_mask {
+    ShiftMask   = (UINT32_C(1) << XKB_MOD_INDEX_SHIFT),
+    LockMask    = (UINT32_C(1) << XKB_MOD_INDEX_CAPS),
+    ControlMask = (UINT32_C(1) << XKB_MOD_INDEX_CTRL),
+    Mod1Mask    = (UINT32_C(1) << XKB_MOD_INDEX_MOD1),
+    Mod2Mask    = (UINT32_C(1) << XKB_MOD_INDEX_MOD2),
+    Mod3Mask    = (UINT32_C(1) << XKB_MOD_INDEX_MOD3),
+    Mod4Mask    = (UINT32_C(1) << XKB_MOD_INDEX_MOD4),
+    Mod5Mask    = (UINT32_C(1) << XKB_MOD_INDEX_MOD5),
+    NoModifier  = 0
+};
 
 static bool
 test_real_mod(struct xkb_keymap *keymap, const char* name,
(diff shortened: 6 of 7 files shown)


mahkoh:
KBVM doesn't have an equivalent of xkb_state_update_mask. KBVM also does not have a concept of virtual modifiers at runtime.

Modifiers are always passed in during lookup and are not adjusted modified. The key-type takes care of masking out unused bits.

During compilation, all given modifier masks are resolved to the final mod mask with the following function:

```rust
    fn get_effective(&self, mask: ModifierMask) -> ModifierMask {
        let mut to_clear = ModifierMask::NONE;
        let mut to_set = ModifierMask::NONE;
        for vmod in &self.virtual_mods {
            let vmod_mask = vmod.index.to_mask();
            if mask.contains(vmod_mask) {
                to_clear |= vmod_mask;
                to_set |= vmod.effective;
            }
        }
        (mask & !to_clear) | to_set
    }
```

This causes

```xkb
        virtual_modifiers X = 0xf0000000;

        key <a> {
            actions = [ SetMods(mods = 0x1100) ],
        };
```

to compile to

```xkb
        key <a> {
            actions = [ SetMods(mods = 0xf0001000) ]
        };
```

wismill:
> KBVM doesn't have an equivalent of xkb_state_update_mask.

@mahkoh How do you update the keyboard state when receiving the Wayland `modifiers` event?

> to compile to
> 
> ```
>         key <a> {
>             actions = [ SetMods(mods = 0xf0001000) ]
>         };
> ```

libxkbcommon would compile to:

```
        virtual_modifiers X = 0xf0000000;
        key <a> {
            // Internal
            actions = [ SetMods(mods = 0xf0000000) ]
            // Serialized
            actions = [ SetMods(mods = X) ]
        };
```

`0x1000` is dropped here because it is not a real modifier nor there is a modifier mapping to it. I think the only difference with kbvm is that libxkbcommon currently drop any bit with no associated virtual modifier at compile time. It is okay-ish because IRL, probably nobody would use the numeric values in the *source* keymap, apart from `virtual_modifiers`.

But we should not drop these bits for consistency. E.g. if one wanted to implement a keymap without declaring any virtual modifier, or… storing a secret state that cannot easily be queried 🕵️. This would be really niche though.

mahkoh:
>@mahkoh How do you update the keyboard state when receiving the Wayland modifiers event?

All client-side code is stateless. The group and modifiers are passed in as part of the lookup operation: https://docs.rs/kbvm/latest/kbvm/lookup/struct.LookupTable.html#method.lookup

wismill:
> All client-side code is stateless. 

@mahkoh Yes, this is a pity that we do not have a better split between server/client API in libxkbcommon. We do not miss that much API to get stateless too though: caps/ctrl transformation + LEDs.

Back on the topic: so with the current PR the only differences with kbvm in the modifiers handling are:
1. At keymap compilation libxkbcommon discard the bits in the modifiers masks that are not in the canonical state mask. Low to no impact expected IRL, see [this comment](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2851690017).
2. libxkbcommon processes extra bits by interpreting them as virtual mod masks and resolving to the corresponding mapping. This is a legacy behavior from previous version, so I kept it. It would have no impact if processing untampered mask from the server (assuming that the server already produces canonical state mask). However, it may be problematic if processing tampered data, if the extra bits are were not intended to encode libxkbcommon’s virtual modifiers.

@mahkoh do you agree?

mahkoh:
I believe that is correct.

>However, it may be problematic if processing tampered data, if the extra bits were not intended to encode libxkbcommon’s virtual modifiers.

AIUI GNOME stores additional data in the upper bytes of the modifiers. I don't know if those are exactly the virtual modifiers as understood by libxkbcommon.

wismill:
Note: once this MR is merged, we may be able to progress on this [Wayland issue](https://gitlab.freedesktop.org/wayland/wayland/-/issues/371), since most (all?) known XKB implementations would use the same modifier encoding.

wismill:
> AIUI GNOME stores additional data in the upper bytes of the modifiers. I don't know if those are exactly the virtual modifiers as understood by libxkbcommon.

It seems that they are the super, hyper and meta vmod masks from libxkbcommon:
- [Masks computation](https://gitlab.gnome.org/GNOME/mutter/-/blob/4d194b8e1da555c14e1dc092879e9580f25b6986/src/core/keybindings.c#L246-L279)
- [State alteration](https://gitlab.gnome.org/GNOME/mutter/-/blob/4d194b8e1da555c14e1dc092879e9580f25b6986/src/wayland/meta-wayland-keyboard.c#L330-L391)

This was introduced by [this commit](https://gitlab.gnome.org/GNOME/mutter/-/commit/0fa9751b3144aef5faaca1ad381b6f86b5f30994) to solve [this issue](https://bugzilla.gnome.org/show_bug.cgi?id=748526). It seems some hack to choose one vmod among vmods  with overlapping mappings.

Even if we do not interpret extra bits as [mentioned above](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2851848503), this can still have nasty effects, as illustrated in [the first issue in the MR description](https://github.com/xkbcommon/libxkbcommon/pull/745#issue-3039918896).

wismill:
I created #746 to fix [discarded extra bits in vmods masks](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2851848503), based on the current MR. I used a separate MR because I was not sure how to handle this issue, if at all.

wismill:
> > @mahkoh How do you update the keyboard state when receiving the Wayland modifiers event?
> 
> All client-side code is stateless. The group and modifiers are passed in as part of the lookup operation: https://docs.rs/kbvm/latest/kbvm/lookup/struct.LookupTable.html#method.lookup

@mahkoh The section [The common client pattern](https://docs.rs/kbvm/latest/kbvm/index.html#the-common-client-pattern) does seem to contradict your comment.

mahkoh:
It's a small helper for calculating the effective group and modifiers: https://docs.rs/kbvm/latest/src/kbvm/components.rs.html#101-104

wismill:
The is an issue if we drop support for the extra processing in `xkb_state_update_mask` (see [above](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2851848503)): we would loose the support for query the virtual modifier masks we just documented in #735! Agreed, this was more a hack that a real feature. I introduced a proper API in #748.

mahkoh:
I've been thinking that it might be better to keep numeric modifier specifications completely untouched. I.e.

```xkb
        virtual_modifiers X = 0xf0000000;

        key <a> {
            actions = [ SetMods(mods = 0x00001100) ],
        };
```

compiles to itself. Otherwise the assignment of indexes to virtual modifiers still leaks into the interpretation of modifiers. This would allow such specifications to be used unambiguously by every implementation without having to know which indexes are assigned to which virtual modifiers.

wismill:
@mahkoh xkbcommon would now leave your example untouched, but only because of the “extra bit” `0x1000`.

However, if we consider

```c
        virtual_modifiers X = 0xf0000000;

        key <a> {
            actions = [ SetMods(mods = 0x00000100) ],
        };
```

then xkbcommon currently serializes it to:

```c
        virtual_modifiers X = 0xf0000000;

        key <a> {
            actions = [ SetMods(mods = X) ],
        };
```

whereas you advocate that it should also be left untouched, did I get it right?

I do not think this is a good idea. The very motivation of virtual mods is to deal with the ambiguous interpretation of the real modifiers. So using numeric values here is similar to using real modifiers: interpretation becomes ambiguous. I would rather not encourage this practice.

Do you have a real use case with numeric masks?

mahkoh:
The serialization in your example implies that the implementation has assigned the 9-th bit to the modifier `X`. However, this assignment is implementation-specific. I believe this is the only remaining place where the parsing order of virtual modifiers leaks into the interpretation of keymaps.

mahkoh:
>Serializing using always numeric masks would achieve 100% interoperability, but at the cost of readability and associated difficulty in debugging.

I don't believe that is a big issue. Almost all keymaps only contain real modifiers and those you can still serialize using their well-known names.

>Tracking the original format (names or numeric) seems quite a big hammer. I am also running out of bits to encode extra info 😂 (that is, without increasing the memory footprint).

You would need to track expressions such as `shift + X` in their original form until the definitions of all virtual modifiers have been fully resolved. At that point you would replace the expression by `0xf0000001`.

wismill:
> I don't believe that is a big issue. Almost all keymaps only contain real modifiers and those you can still serialize using their well-known names.

It is if you develop a custom keymap or if you maintain a keyboard layout database such as xkeyboard-config. Having `SetMods(modifiers=Mod1)` is much less readable than `SetMods(modifiers=Alt)`, etc.

Not to mention that it would break all our golden tests.

> You would need to track expressions such as shift + X in their original form until the definitions of all virtual modifiers have been fully resolved.

Exactly. Not worth the effort.

Now, I knew that I already thought about using numeric mask a while ago: see #640. This would give the best of both worlds:
- exact and fast serializing *by default* by using numeric masks.
- human-friendly masks *on demand*, that we sure want to use in our tests. This would result in implementation-specific masks, as mentioned above. But this is perfectly legit in testing and debugging contexts.

mahkoh:
>It is if you develop a custom keymap or if you maintain a keyboard layout database such as xkeyboard-config. Having SetMods(modifiers=Mod1) is much less readable than SetMods(modifiers=Alt), etc.

Those keymaps are not the result of serialization. They are hand written.

What I'm suggesting is that you can still write `SetMods(modifiers=Alt)` and the behavior is as it has always been. Only if the input keymap already contains numeric notation, e.g. `SetMods(modifiers=0x100)`, the interpretation of those numbers should not be affected by the parsing order of virtual modifiers.

In fact, it would also be fine if xkbcommon continued serializing virtual modifier names. The only change I'm suggesting is that during the parsing process, if the source keymap contains a modifier specified as a number, that number is not affected by the virtual modifier parsing order.

>Exactly. Not worth the effort.

The effort seems rather negligible.

mahkoh:
Actually, you probably don't even need to keep the ast around. You just need two numbers: `u32 explicit_modifiers`, `u32 virtual_modifiers` where `explicit_modifiers` only contains the real modifiers and those specified via numbers and `virtual_modifiers` contains the virtual modifier **indices**. Later when you do the resolution from virtual modifier index to virtual modifier definition, you only look at the `virtual_modifiers` fields so that you don't accidentally translate values from the `explicit_modifiers`.

mahkoh:
I've implemented this here: https://github.com/mahkoh/kbvm/pull/44

wismill:
> > It is if you develop a custom keymap or if you maintain a keyboard layout database such as xkeyboard-config. Having SetMods(modifiers=Mod1) is much less readable than SetMods(modifiers=Alt), etc.
> 
> Those keymaps are not the result of serialization. They are hand written.

Yet *if* the compiler always serializes to numeric/real modifiers, it makes debugging difficult, hand written or not.

> > Exactly. Not worth the effort.
> 
> The effort seems rather negligible.

Then you are welcome to file a PR. This kind of comment is all but constructive.

> Actually, you probably don't even need to keep the ast around. You just need two numbers: `u32 explicit_modifiers`, `u32 virtual_modifiers` where `explicit_modifiers` only contains the real modifiers and those specified via numbers and `virtual_modifiers` contains the virtual modifier **indices**. Later when you do the resolution from virtual modifier index to virtual modifier definition, you only look at the `virtual_modifiers` fields so that you don't accidentally translate values from the `explicit_modifiers`.

You need to keep the AST around, because you cannot resolve nor store modifiers separately when using operations (apart from OR `+`), especially if mixing types: how would you handle e.g. `Alt - ~0x1` with the method you propose, if Alt is not yet resolved?

mahkoh:
>You need to keep the AST around, because you cannot resolve nor store modifiers separately when using operations (apart from OR +), especially if mixing types: how would you handle e.g. Alt - ~0x1 with the method you propose, if Alt is not yet resolved?

No existing XKB implementation supports this. For example, xkbcommon compiles

```xkb
        virtual_modifiers Alt = Mod1;
        key <a> {
            actions = [ SetMods(mods = Alt - Mod1) ]
        };
```

to

```xkb
        virtual_modifiers Alt = Mod1;
        key <a> {
            actions = [ SetMods(mods = Alt) ]
        };
```

And it makes sense, because `Alt - Mod1` = `0x100 - 0x08` = `0x100`.

wismill:
> No existing XKB implementation supports this.

What do you mean?

What I am trying to explain is that I do not see how you can achieve:

> during the parsing process, if the source keymap contains a modifier specified as a number, that number is not affected by the virtual modifier parsing order.

by just separating numeric masks from named modifiers as you suggested above, *before* vmod resolution and *without* keeping the AST around.

To me it seems mandatory to keep the relevant AST until the vmod are resolved. And this is not a “negligible” task, as you mentioned above.

mahkoh:
But I agree in principle that handling supporting such mixed operations is valuable. So I've changed my PR to use the "lazy evaluation" approach that I suggested originally. Surprisingly the entire diff is only +30 lines excluding test cases.

wismill:
So I gave it a second thought, and I came to the conclusion that *we are on the wrong track* here. My stance is therefore:

> [!IMPORTANT]
> Any field that uses virtual modifier names **is** a virtual modifier **mask**, denoting their *indices*.

Indeed:
- This is what the historical reference implementation, xkbcomp, implements. It is even really transparent about it: see *VMod* in e.g. `ExprResolveModMask(value, &rtrn, LookupVModMask, (XPointer) xkb)`.
- This is what the reference modern implementation, xkbcommon, does. And this implementation is the reference for the Wayland protocol.
- The scarce documentation never mentions numeric masks, but only describes the use of *virtual* modifiers.
- Numeric masks are interpreted as virtual mod masks in both xkbcomp and xkbcommon.

> [!IMPORTANT]
> Therefore, if there is no doc, the reference implementation *is* the specification.

Added to that:
- Interpreting virtual mod fields as mixed vmods + rmods complicates implementation. What you implemented in mahkoh/kbvm#44 is not exactly a trivial change, however clever or concise it is in Rust.
- Interpreting virtual mod fields as rmods masks is *lossy*, as I mention before. It makes debugging difficult or impossible, given that the original semantics (the virtual mods) are lost.

> [!WARNING]
> @mahkoh I think you read `virtual_modifiers X = 0x100;` as a variable `X` to substitute wherever `X` appears in the code, but the analysis above shows that this is not the original intent.

All this mess is happening because of one root issue: *virtual mods indexes are implementation-specific*.

So, what I propose is that instead of changing the semantics, we should simply *specify the computation of virtual modifiers indexes*. My first take on this is obviously to follow what xkbcomp and xkbcommon do:

> [!TIP]
> 1. Virtual modifiers names are assigned to virtual mods (indexes) following their order of appearance in a keymap component, if that name was not previously assigned.
> 2. The order of the relevant keymap components used for virtual mods assignment is:
>    1. key types
>    2. compatibility
>    3. symbols
> 3. A virtual modifier mask denotes virtual modifiers indexes.

Now, it suffices to declare all virtual modifiers in `xkb_types` (or if empty, whatever non-empty section afterwards, in the order specified above) in their ascending indexes order to be compatible with xkbcommon. If you wish to always use numeric masks, you simply have to use the *virtual* mod masks in the relevant places, not the *real* mod masks.

wismill:
@mahkoh If we agree on interpreting only as virtual masks, then let’s discuss the vmod index specification in #755. If not, let’s continue discussion here, to keep the info in one thread.

mahkoh:
>This is what the historical reference implementation, xkbcomp, implements. It is even really transparent about it: see VMod in e.g. ExprResolveModMask(value, &rtrn, LookupVModMask, (XPointer) xkb).

I don't believe references to xkbcomp are useful here. Statements such as

```xkb
        virtual_modifiers V2 = 0x100;
```

do not make sense in xkbcomp since xkbcomp automatically truncates the RHS to the lower 8 bits. If you want to support things such as pure virtual modifiers then xkbcomp cannot serve as a guide.

>Numeric masks are interpreted as virtual mod masks in both xkbcomp and xkbcommon.

I don't believe there was any intentional design behind this. The original developers simply likely never expected the source file to contain any explicit numbers larger than 255 (in this context) and therefore thought that they could use the unused upper bytes as a transport for virtual modifier indices.

If they had not re-used one of the existing expression evaluation functions, they might as well have truncated all explicit numeric values that appear in all mod positions to the lower 8 bits, to be consistent with the statement above.

>Interpreting virtual mod fields as mixed vmods + rmods complicates implementation.

As you've said yourself above, you expected something like

```xkb
virtual_modifiers Alt; // implicitly set to Mod1 via modmap elsewhere
key <a> { [ SetMods(mods = Alt - Mod1) ] };
```

to produce `SetMods(mods = None)`. Yet neither xkbcomp nor xkbcommon do this. To fix this, I believe you would have to all of the work anyway by keeping the expression structure around until much later into the keymap evaluation (at least until all implicit vmod definitions are resolved). Do you no longer believe that `Alt - Mod1` should behave like this?

>Interpreting virtual mod fields as rmods masks is lossy, as I mention before.

In which ways is it lossy? Is it not exactly the opposite since it also coincidentally enables you to solve the `Alt - Mod1` issue?

> It makes debugging difficult or impossible, given that the original semantics (the virtual mods) are lost.

I don't believe anything is lost since you're storing strictly more information. During formatting, you can continue to format the virtual modifier names if you choose to do so.

>@mahkoh I think you read virtual_modifiers X = 0x100; as a variable X to substitute wherever X appears in the code, but the analysis above shows that this is not the original intent.

The original intent was, if we're going by xkbcomp, that the RHS is truncated to the lower 8 bits. I believe xkbcommon has already decided to disregard the original intent.

>So, what I propose is that instead of changing the semantics, we should simply specify the computation of virtual modifiers indexes. My first take on this is obviously to follow what xkbcomp and xkbcommon do:

I disagree that implementation details such as virtual modifier indices should be elevated when there is no precedent for this except ancient bugs in implementations that do not support any of the relevant features such as pure virtual modifiers or multi-byte modifier masks.

mahkoh:
In current xkbcommon master, interpreting numbers as virtual modifier indices also has the following negative side effect:

```xkb
xkb_keymap {
    xkb_keycodes {
        <shift> = 50;
        <a> = 38;
    };

    xkb_types {
        type "X" {
            modifiers = 0x100;
            map[0x100] = Level2;
        };
    };

    xkb_compat { };

    xkb_symbols {
        key <shift> {
            [ Shift_L ],
            [ SetMods(mods = 0x100) ]
        };
        key <a> {
            [ a, A ]
        };
    };
};
```

This works as one might expect: `Shift+A` produces the uppercase `A`.

However, if you now add `virtual_modifiers A;` anywhere (possibly in an include), `Shift+A` no longer produces the uppercase `A`.

wismill:
> > This is what the historical reference implementation, xkbcomp, implements. It is even really transparent about it: see VMod in e.g. ExprResolveModMask(value, &rtrn, LookupVModMask, (XPointer) xkb).
> 
> I don't believe references to xkbcomp are useful here. Statements such as
> 
> ```
>         virtual_modifiers V2 = 0x100;
> ```
> do not make sense in xkbcomp since xkbcomp automatically truncates the RHS to the lower 8 bits. If you want to support things such as pure virtual modifiers then xkbcomp cannot serve as a guide.

You missed the point: the snippet I quoted is *not* about `virtual_modifiers` but about all the fields where one can use a virtual modifier name. xkbcomp *does* resolve those field to a (virtual) modifier mask.

> > Numeric masks are interpreted as virtual mod masks in both xkbcomp and xkbcommon.
> 
> I don't believe there was any intentional design behind this. The original developers simply likely never expected the source file to contain any explicit numbers larger than 255 (in this context) and therefore thought that they could use the unused upper bytes as a transport for virtual modifier indices.

Or: it treats numeric values the same way it treat names, for *consistency*.

> > Interpreting virtual mod fields as mixed vmods + rmods complicates implementation.
> 
> As you've said yourself above, you expected something like
> 
> ```
> virtual_modifiers Alt; // implicitly set to Mod1 via modmap elsewhere
> key <a> { [ SetMods(mods = Alt - Mod1) ] };
> ```
> to produce `SetMods(mods = None)`.

I never wrote this! What [I wrote](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2858942707) was simply:

> how would you handle e.g. `Alt - ~0x1` with the method you propose, if Alt is not yet resolved?

> Yet neither xkbcomp nor xkbcommon do this.
> […]
> Do you no longer believe that `Alt - Mod1` should behave like this?

I do not and I did not before. If `Alt` has index 3, I would expect: `(1<<3) & ~0x8 = 0x100 = Alt`.

> > Interpreting virtual mod fields as rmods masks is lossy, as I mention before.
> 
> In which ways is it lossy? Is it not exactly the opposite since it also coincidentally enables you to solve the `Alt - Mod1` issue?
> 
> > It makes debugging difficult or impossible, given that the original semantics (the virtual mods) are lost.
> 
> I don't believe anything is lost since you're storing strictly more information. During formatting, you can continue to format the virtual modifier names if you choose to do so.

It seems I do not express myself clearly. If you have the following snippet:
```c
type "X" {
  modifiers = Alt;
  map[Alt] = 2;
};
```

and `Alt` resolves to the usual `Mod1`, you propose that 

```c
type "X" {
  map[Mod1] = 2;
};
```

is strictly equivalent, but it is not, at least not from the *semantics* point of view. I feel that you focus solely on the keyboard state and consider the first syntax as a useless historically artifact. While your version is equivalent from the point of view of an application that care only about the keyboard state, it does *loss* the semantics, which you *cannot retrieve* if any other virtual modifier resolves to `Mod1` too. And semantics *are* important if debugging or wanted to modify the keymap.

I mean, the whole point of XKB was to improve the core X protocol; one of the improvement was to avoid using directly `Mod1-Mod5` because they lack semantics. So going real modifiers-only would be a regression in the keymap would be a regression.

> [!NOTE]
> Do not get me wrong: I wish we never had to deal with virtual modifiers in the first place.
> E.g. #450 and #757 are intents to move past the implicit vmod mappings.

> > I think you read `virtual_modifiers X = 0x100;` as a variable `X` to substitute wherever `X` appears in the code, but the analysis above shows that this is not the original intent.
> 
> The original intent was, if we're going by xkbcomp, that the RHS is truncated to the lower 8 bits. I believe xkbcommon has already decided to disregard the original intent.

You miss the point by bringing xkbcomp and focusing on truncation. The point is that it is possible to have a direct explicit mapping to whatever the compiler supports as modifier encoding: 8bits for xkbcomp, 32 for libxkbcommon. Then everywhere else, `X` does not denote its encoding but *itself*, using its unambiguous mask based on its index.

This is what xkbcomp and libxkbcommon do and have done so forever.

> > So, what I propose is that instead of changing the semantics, we should simply specify the computation of virtual modifiers indexes. My first take on this is obviously to follow what xkbcomp and xkbcommon do:
> 
> I disagree that implementation details such as virtual modifier indices should be elevated when there is no precedent for this except ancient bugs in implementations that do not support any of the relevant features such as pure virtual modifiers or multi-byte modifier masks.

I am not a fan of it either, but this is the only *backward-compatible* and *pragmatic* solution I can think of. Also, is it really just “implementation details” and “there is no precedent” when there is e.g. [`XkbGetVirtualMods`](https://www.x.org/releases/current/doc/man/man3/XkbGetVirtualMods.3.xhtml) and xkbcommon API which *explicitly exposes the vmod indices*?


> In current xkbcommon master, interpreting numbers as virtual modifier indices also has the following negative side effect:
> ```c
> xkb_keymap {
>     xkb_keycodes {
>         <shift> = 50;
>         <a> = 38;
>     };
> 
>     xkb_types {
>         type "X" {
>             modifiers = 0x100;
>             map[0x100] = Level2;
>         };
>     };
> 
>     xkb_compat { };
> 
>     xkb_symbols {
>         key <shift> {
>             [ Shift_L ],
>             [ SetMods(mods = 0x100) ]
>         };
>         key <a> {
>             type = "X",
>             [ a, A ]
>         };
>     };
> };
> ```
> 
> This works as one might expect: `Shift+A` produces the uppercase `A`.
> 
> However, if you now add `virtual_modifiers X;` anywhere (possibly in an include), `Shift+A` no longer produces the uppercase `A`.

1. I see *no inconsistency*: without declaring `X`, it interpret the bit in the numeric mask as an *anonymous* virtual modifier.
2. When declaring `X`, we are back to the *classical behavior*: it is neither explicitly nor implicitly mapped, so it mapped to 0, which makes the type drop the entry. You actually reported this in #578 (thanks!), which I fixed in #749.
3. What use case is this?
   - I advocate to follow the *standard practice* of *never using a numeric value*. E.g. xkeyboard-config would never use such config.
   - I never saw explicit numeric values anywhere else than in kbvm. Yet kbvm would always produce explicit vmod mappings.
   - So this would only be a hand-written keymap. Who would do that? And if using a numeric value, that would be an advanced user, as this is documented nowhere.

In short: I see only a *consistent* edge-case.

wismill:
> If that is your meaning and you want to treat virtual modifiers as distinct from each other and what they map to, then the answer is obvious. `Alt - ~0x1` = `Alt - ~Shift` = `None`. This behavior is not affected by anything that I have suggested.

Well with my meaning yes, but you took my answer out of context: the question was about *your* suggestion of implication, with the “negligible” effort you mentioned.

> But I don't believe we will come to an agreement here.

Well, we will not if you disregard most of my comments, in particular those about the semantics.

mahkoh:
>Well with my meaning yes, but you took my answer out of context: the question was about your suggestion of implication, with the “negligible” effort you mentioned.

If you want to apply your interpretation of `Alt - XYZ`, then you don't need to keep the AST around which would make the whole thing very simple even if xkbcommon. Keeping the AST around is only necessary if you want to operate on the resolved values, which you don't.

>Well, we will not if you continuously disregard my comments, in particular those about the semantics.

You have claimed that I'm saying that `SetMods(mods = Alt)` and `SetMods(mods = Mod1)` should be treated identically. But I've never suggested this and I've already said as much. Everything I've suggested only affects statements containing modifiers encoded as number literals. Neither of these two statements contain number literals.

wismill:
> If you want to apply your interpretation of `Alt - XYZ`, then you don't need to keep the AST around which would make the whole thing very simple even if xkbcommon. Keeping the AST around is only necessary if you want to operate on the resolved values, which you don't.

I am glad that you realized that, as this is what I tried to explain you. Let’s me put this again in [context](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2858942707) where it started:

> > Actually, you probably don't even need to keep the ast around. You just need two numbers: `u32 explicit_modifiers`, `u32 virtual_modifiers` where `explicit_modifiers` only contains the real modifiers and those specified via numbers and `virtual_modifiers` contains the virtual modifier **indices**. Later when you do the resolution from virtual modifier index to virtual modifier definition, you only look at the `virtual_modifiers` fields so that you don't accidentally translate values from the `explicit_modifiers`.
> 
> You need to keep the AST around, because you cannot resolve nor store modifiers separately when using operations (apart from OR `+`), especially if mixing types: how would you handle e.g. `Alt - ~0x1` with the method you propose, if Alt is not yet resolved?

---

> You have claimed that I'm saying that `SetMods(mods = Alt)` and `SetMods(mods = Mod1)` should be treated identically. But I've never suggested this and I've already said as much.

No, what I claim is you treat the `mods` field of `SetMods(mods = Alt)` as being a real modifier mask and that is resolved to `SetMods(mods = Mod1)` after vmod mappings are resolved. Thus kbvm serializes `SetMods(mods = Alt)` to `SetMods(mods = Mod1)` and the original expression cannot be retrieved if another vmod maps to `Mod1`, e.g. `Meta`. This is the lossy part that you continue to disregard, because to your eyes apps will work identically so why bother?

Then there is the issue with operations on masks, e.g. for the `mods` field of `SetMods(mods=…)`. I claim this is a non-issue if treating the field as a *virtual* mod mask, but you insist to interpret the field as a *real* mask (well, “extended” real, aka mod encoding), which thus requires special processing, while in my view:
1. this interpretation is wrong: see evidences in xkbcomp and xkbcommon implementations + XKB philosophy;
2. it should always be interpreted as a virtual mod mask, numeric values included;
3. thus no special handling is required.

I do not see what it would bring us to *break* the historical behavior.

PR #746, opened 2025-05-06
Do not discard extra bits in vmod masks

wismill:
Since we accept numeric values for the vmod mask in the keymap, we may have extra bits set that encode *no* real/virtual modifier. Keep them unchanged for consistency.

E.g. the following keymap:

```c
xkb_keymap {
    xkb_keycodes { <a> = 38; };
    xkb_symbols {
        virtual_modifiers X = 0xf0000000;
        key <a> { [ SetMods(mods = 0x00001100) ] };
    };
};
```

would compile to:

```c
xkb_keymap {
    xkb_keycodes { <a> = 38; };
    xkb_symbols {
        virtual_modifiers X = 0xf0000000;
        // Internal state
        key <a> { [ SetMods(mods = 0xf0001000) ] };
        // Serializatiion
        key <a> { [ SetMods(mods = 0x00001100) ] };
    };
};
```

See also [this discussion](https://github.com/xkbcommon/libxkbcommon/pull/745#issuecomment-2851848503).

TODO:
- [ ] changelog? it’s very niche
- [x] rebase on #745

### DIFF
--- a/src/xkbcomp/keymap.c
+++ b/src/xkbcomp/keymap.c
@@ -30,10 +30,18 @@ has_unbound_vmods(struct xkb_keymap *keymap, xkb_mod_mask_t mask)
     return false;
 }
 
-static void
+static inline void
 ComputeEffectiveMask(struct xkb_keymap *keymap, struct xkb_mods *mods)
 {
-    mods->mask = mod_mask_get_effective(keymap, mods->mods);
+    /*
+     * Since we accept numeric values for the vmod mask in the keymap, we may
+     * have extra bits set that encode no real/virtual modifier. Keep them
+     * unchanged for consistency.
+     */
+    const xkb_mod_mask_t unknown_mods = ~(xkb_mod_mask_t)
+        ((UINT64_C(1) << keymap->mods.num_mods) - UINT64_C(1));
+    mods->mask = mod_mask_get_effective(keymap, mods->mods)
+               | (mods->mods & unknown_mods);
 }
 
 static void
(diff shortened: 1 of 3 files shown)


wismill:
(note to self)

I found it awkward to use random bits, until I read the XKB protocol again:

> XKB provides a set of sixteen named virtual modifiers

([source](https://www.x.org/releases/current/doc/kbproto/xkbproto.html#:~:text=XKB%20provides%20a%20set%20of%20sixteen%20named%20virtual%20modifiers))

I think the phrasing is wrong and should be understood as:

> XKB provides a set of sixteen virtual modifiers *that can be optionally named*.

The names are not defined in the protocol but rather in each keymap; xkeyboard-config being the reference source. One should be able to use any virtual modifier, even if it is *unnamed*.

> [!IMPORTANT]
> Those random bits are in fact just *unnamed __extended__ real* modifiers, in X11’s terminology. Just the same as one can use unnamed vmods in X11.

---

I keep being puzzled by:
1. Using “modifier” for both human-friendly modifiers (i.e. *virtual* mods) and their encoding (i.e. *real* mods).
2. We have 8 named real modifiers, but the rest is unnamed.
3. `xkb_mod_mask_t` being used for both real and virtual modifiers.

At some point we should abandon the real/virtual terminology, it is just confusing. But what then?
- *canonical modifier* sounds like a nice replacement for real modifier: i.e. it maps to itself.
- *modifier alias* sounds like a good candidate for virtual modifier, although it still suggests a 1:1 relation. 
- “Named canonical modifier mask” would be a good description of what a virtual mod is, but it’s a bit too long 😅.

PR #527, opened 2024-10-11
Actions/locks xkb extension v3

wismill:
This is a refactor of #485, rebased on master and without exposing the new format yet, but using a hidden `XKB_EXPERIMENTAL_LOCKS` environment variable to activate it, until we agree on an API (see #484).

In order to fix #420, one must:
- Set the env variable `XKB_EXPERIMENTAL_LOCKS` to something, e.g. `true`.
- Modifiy `/usr/share/xkb/symbols/group` (preferably by overriding it via `~/.config/xkb`), that is use multiple actions thanks to #487.

```diff
partial modifier_keys
xkb_symbols "lctrl_lshift_toggle" {
-    key <LFSH> {[  Shift_L,  ISO_Next_Group  ], type[group1] = "PC_CONTROL_LEVEL2" };
+    key <LFSH> {[  Shift_L,  {Shift_L, ISO_Next_Group}  ], type[group1] = "PC_CONTROL_LEVEL2" };
-    key <LCTL> {[  Control_L,  ISO_Next_Group  ]};
+    key <LCTL> {[  Control_L,  {Control_L, ISO_Next_Group}  ]};
};

partial modifier_keys
xkb_symbols "lctrl_lshift_toggle_rev" {
-    key <LFSH> {[  Shift_L,  ISO_Prev_Group  ], type[group1] = "PC_CONTROL_LEVEL2" };
+    key <LFSH> {[  Shift_L,  {Shift_L,ISO_Prev_Group}  ], type[group1] = "PC_CONTROL_LEVEL2" };
-    key <LCTL> {[  Control_L,  ISO_Prev_Group  ]};
+    key <LCTL> {[  Control_L,  {Control_L,ISO_Prev_Group}  ]};
};

partial modifier_keys
xkb_symbols "rctrl_rshift_toggle" {
-    key <RTSH> {[  Shift_R,  ISO_Next_Group  ], type[group1] = "PC_CONTROL_LEVEL2" };
+    key <RTSH> {[  Shift_R,  {Shift_R,ISO_Next_Group}  ], type[group1] = "PC_CONTROL_LEVEL2" };
-    key <RCTL> {[  Control_R,  ISO_Next_Group  ]};
+    key <RCTL> {[  Control_R,  {Control_R,ISO_Next_Group}  ]};
};
```

PR #754, opened 2025-05-07
Introduce keymap text format v2

wismill:
I think it is due time that we add support for new keymap formats, in order to solve 20-years old issues. This PR introduces the new keymap text format but with no new feature; they will be added in follow-up PR.

- Add new keymap format `XKB_KEYMAP_FORMAT_TEXT_V2`
- Add `xkb_keymap_new_from_names2()`
- Deprecate `xkb_keymap_new_from_names()`
- Make `xkb_keymap_new_from_names()` use `XKB_KEYMAP_FORMAT_TEXT_V2` internally
- Enable tools to use an explicit keymap format for parsing and serialing
- Internal API to query and parse supported keymap formats
- Add initial documention for XKB_KEYMAP_FORMAT_TEXT_V2

Foreseen applications are #515 and #527.

@bluetech @whot


wismill:
Let’s take the plunge!

PR #757, opened 2025-05-08
state: support pure virtual modifiers V2

### LINKED ISSUE #755: Specify virtual modifier indexes assignment
I propose to *specify the computation of virtual modifiers indexes* in order to improve interoperability and to solve the issue with numeric masks. My first take on this is obviously to follow what xkbcomp and xkbcommon do:

> [!IMPORTANT]
> 1. Virtual modifiers names are assigned to virtual mods (indexes) following their order of appearance in a keymap component, if that name was not previously assigned.
> 2. Indexes are assigned in ascending order, from 0.
> 3. If an index is already assigned to a different name, then the next free index is used.
> 4. The order of the relevant keymap components used for virtual mods assignment is:
>    1. key types
>    2. compatibility
>    3. symbols
> 5. A virtual modifier mask denotes virtual modifiers indexes.

Now, it suffices to declare all virtual modifiers in `xkb_types` (or if empty, whatever non-empty section afterwards, in the order specified above) in their ascending indexes order to be compatible with xkbcommon. If one wish to always use numeric masks, they simply have to use the *virtual* mod masks in the relevant places, not the *real* mod masks.

_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/issues/745#issuecomment-2861779818_

### LINKED ISSUE #802: `self` and `auto` special` modifier encoding values
What about introducing the `self`[^1] special value to use more “explicit” mapping, rather than mapping auto-magically any unmapped modifier?

```c
virtual_modifiers M = self;
``` 

Alternatively, `auto` could be used to set to `self` only if the modifier is not bound implicitly (`modmap`).

[^1]: Or `canonical`; naming subject to bikeshedding

_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/issues/757#issuecomment-2889804452_

wismill:
Up-to-date alternative design to #450. Pure virtual modifiers are already supported, albeit requiring an explicit mapping. This PR only introduces assigning automatically a canonical mapping to vmod that remain unmapped after the usual vmod resolution.

I do not think we want this by default, at least not in the v1 keymap format, as it introduces breakage. However it could be activated by default in the v2 keymap format.

TODO:
- [x] Ensure automated canonical mapping is made explicit at serialization
- [x] fix #755
- [x] guard with v2 keymap format check
- [x] update the tests
- [x] update description
- [x] Update the commit messages to reflect current implementation
- [x] Document new feature
- [ ] Document [new terminology](https://github.com/xkbcommon/libxkbcommon/pull/450#issuecomment-2497364099)
- [x] Changelog

Closes #447
Closes #450
Closes #802

---

Traditionally, *virtual* modifiers were merely name aliases for *real* modifiers (X *core* modifiers), e.g. `NumLock` was usually mapped to `Mod2` (see `modifier_map` statement). Virtual modifiers that were never mapped to a real ones had no effect on the keymap state.

xkbcommon already supports the concept of “pure” virtual modifiers, i.e. virtual modifiers that are *encoded* using the full 32-bit range, not just the first 8 bits corresponding to the real modifiers.

But until this commit, one had to declare such mapping *explicitly*: e.g. `virtual_modifiers M = 0x100;`. This has at least two drawbacks:
- Numerical values may look quite arbitrary and are not user-friendly. It’s OK in the resulting compiled keymap, but it requires careful sync between sections when developing KcCGST files.
- If the modifier is *also* mapped *implicitly* using the traditional `vmodmap`/`modifier_map`, then both mappings are OR-combined.

This patch enables to automatically map unmapped virtual modifiers to their *canonical* mapping, i.e. themselves: their corresponding virtual and real modifier masks are identical: `1u << mod_index`.

Since this feature is incompatible with X11, this is guarded by requiring at least keymap text format **v2**.

Note that for now, canonical virtual modifiers cannot be used in an interpret action’s `AnyOf()`. An interpret action for a canonical virtual modifier must be `AnyOfOrNone()` to take effect:

```c
    virtual_modifiers APureMod, …;

    interpret a+AnyOfOrNone(all) {
      virtualModifier= APureMod;
      action= SetMods(modifiers=APureMod);
    };
```

The above adds a virtual modifier `APureMod` for keysym `a`. It will be canonical iff it is not mapped implicitly.


wismill:
What about introducing the `self` special value to use more “explicit” mapping, rather than mapping auto-magically any unmapped modifier?

```c
virtual_modifiers M = self;
```

PR #763, opened 2025-05-09
actions: Add VoidAction(), mirroring NoSymbol/VoidSymbol.

### LINKED ISSUE #622: Add VoidAction() to mirror NoSymbol/VoidSymbol
_Originally posted by @wismill in https://github.com/xkbcommon/libxkbcommon/issues/620#issuecomment-2618155046_
            
I thought about introducing `VoidAction()`, to mirror `NoSymbol/VoidSymbol`. Because currently we also do not have an easy way to erase previous explicit actions when using RMLVO:
- `NoAction()` does not overwrite the previous action.
- `replace key` works only in the same file or after a child file.
- `SetControls()` kind of work if you want to break latches. But it’s ugly.

The no/void qualifiers are not very clear, but at least we would be consistent with the keysyms.

wismill:
Added `VoidAction()` action to match the keysym pair `NoSymbol` / `VoidSymbol`.

It enables erasing a previous action and breaks latches.

This is a libxkbcommon extension. When serializing it will be converted to `LockControls(controls=none,affect=neither)` for backward compatibility. We cannot serialize it to `NoAction()`, as it would be dropped in e.g. the context of multiple actions.

Fixes #622

@mahkoh @Jules-Bertholet

### DIFF
--- a/changes/api/622.feature.md
+++ b/changes/api/622.feature.md
@@ -1,0 +1,5 @@
+Added `VoidAction()` action to match the keysym pair `NoSymbol`/`VoidSymbol`.
+It enables erasing a previous action and breaks latches.
+
+This is a libxkbcommon extension. When serializing it will be converted to
+`LockControls(controls=none,affect=neither)` for backward compatibility.
--- a/doc/keymap-format-text-v1.md
+++ b/doc/keymap-format-text-v1.md
@@ -2555,16 +2555,17 @@ In order to define and use a modifier, one must:
 
 The following table provide an overview of the available actions:
 
-| Category | Action              | Alias            | Description                        |
+| Category | Action name         | Alias            | Description                        |
 |----------|---------------------|------------------|------------------------------------|
-|          | `NoAction`          |                  | Do nothing.                        |
-| [Modifier action] | `SetMods`           |                  | Modifies the _depressed_ modifiers |
+| [Ineffectual action] | [`NoAction`][no-action]|   | **Default action**: *implicitly* do nothing. |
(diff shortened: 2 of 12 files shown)


changes/api/622.feature.md
```diff
@@ -0,0 +1,5 @@
+Added `VoidAction()` action to match the keysym pair `NoSymbol`/`VoidSymbol`.
+It enables erasing a previous action and breaks latches.
+
+This is a libxkbcommon extension. When serializing it will be converted to
+`LockControls(controls=none,affect=neither)` for backward compatibility.
```
mahkoh, line 5: Does LockControls also break latches?

wismill: [It does](https://gitlab.freedesktop.org/xorg/proto/xorgproto/-/blob/1914233e662d23ffb3812b80fadd0bbd064ad91c/include/X11/extensions/XKB.h#L508-L512). See also our own `xkb_action_breaks_latch()`.

mahkoh: In that case, AIUI, VoidAction is just a nicer alias for the LockControls action. If so, this LGTM.

wismill: Exactly. This brings better semantics and we can make it a proper independent action later. I chose `LockControls(controls=none,affect=neither)` because it seems the only portable action we support that would cause no side effect other than breaking latches. `SetControls()` may be fine too, but the `affect=neither` field of `LockControls` seems  to better ensure it is a dummy action.

mahkoh:
https://github.com/mahkoh/kbvm/pull/45

PR #778, opened 2025-05-22
build: Require Bison ≥ 3.6

bgilbert:
Older versions don't support `%define parse.error detailed`.

Fixes: a3f1a9d3ba ("xkbcomp/parser: enable Bison detailed syntax error")


wismill:
Thanks! I added a changelog entry to highlight this. Note: Bison 3.6 was released in Mai 2020, so it should affect nobody.

PR #794, opened 2025-06-12
Actions: Locks XKB extension v4

### LINKED ISSUE #372: CapsLock change upon press instead of release option.
Hey, could we add an option to change the CapsLock button to release on button press instead of button release?

The current behaviour leads to accidental uppercase letters after trying to switch from upper to lower case when writing fast:

"THe elephant was very thirsty. UNfortunately there was no water around."

Currently there are a couple of scripts out there that fix this by setting CapsLock as a Ctrl Modifier and some other stuff (as seen below), but afaik this does not work for NixOS at the moment. At least not by changing the configuration.nix (or maybe I'm too dumb, but then so are probably 90% of the other casual Windows users). I already went to the NixOS discord and they could only point me towards xkb.
As Linux grows in popularity more, I assume more people will start switching from Windows to Linux and encounter this problem. This inconvenience has been around for forever, there are threads that are like 10 years old. 
I think an official option to change this would be greatly appreciated instead of using weird system hacks and automated scripts.

This is the script that would supposedly fix this on arch/ubuntu btw:
```

// This changes the <CAPS> key to become a Control modifier,
// but it will still produce the Caps_Lock keysym.
hidden partial modifier_keys
xkb_symbols "ctrl_modifier" {
    key <CAPS> {
        type = "ALPHABETIC",
        repeat = No,
        symbols[Group1] = [ Caps_Lock, Caps_Lock ],
        actions[Group1] = [ LockMods(modifiers = Lock), LockMods(modifiers = Shift + Lock, affect = unlock) ]
    };
};
```
Taken from https://github.com/rafaeloledo/linux-capslock-delay-fix .

Thanks in advance. Oh and before anyone starts the usual discussion about just switching to Shift: 

"Sean Wrona, one of the world's fastest typists, uses Caps Lock for capital letters. He often types at 250+ wpm.
From his [website](http://seanwrona.com/typing.php):
'I recommend using caps lock instead of shift to type capital letters to allow more flexibility in the hand that you would normally use shift with.' "

### LINKED ISSUE #420: keyboard layout change to CTRL-Shift/ALT-Shift breaks other shortcuts starting with CTRL-Shift/ALT-Shift
Trying to move from Windows to Linux & KDE (Gnome is affected as well). Found a ground braking misbehavior, making the move for me impossible. I used to change layouts with CTRL-Shift on Windows for 30 years as many other users.

Many multi language users suffers from this issue.

STEPS TO REPRODUCE
1. Add keyboard layouts: English, Estonian, Russian (it should work with only two added languages, but I have three of them)
2. Set keyboard language switcher to CTRL-Shift  

Try to:
a) in Firefox close a tab and reopen it with CTRL-Shift-T shortcut
b) in Konsole try to copy a text with CTRL-Shift-C shortcut
c) any CTRL-Shift-X combinations in any program that supports any of such shortcuts

OBSERVED RESULT
a) opens a new tab
b) sends CTRL-C to terminate a command
c) any CTRL-Shift-X will not work

EXPECTED RESULT
a) reopens a closed tab
b) copies selected text
c) shortcuts works as designed by the program

THE PROBLEM ROOT
Pressing CTRL-Shift immediately changes the language, so the pressed letter does not do anything. Why? Of XBD standard. XBM triggers on pressing of a button, not on a release.

More information:
A 20 years old bug for X11: https://bugs.freedesktop.org/show_bug.cgi?id=865
KDE discussion: https://bugs.kde.org/show_bug.cgi?id=383113

There is a patched X11 available, but since all moves to Wayland, it is not an option anymore:
https://www.shellhacks.com/ctrl-shift-key-not-working-ubuntu-linux-mint/

---

EDIT 2024-06-17:

Now I have the exact guide to workaround this problem in KDE 6.1 and newer (release is scheduled for next week, 18.06.2024).

1. Open Settings -> Keyboard -> Layouts
2. Delete the assigned MAIN shortcut (CTRL-Shift or ALT-Shift), you may have to uncheck it in "Key Bindings" and apply.
3. Assign Ctrl+Shift as ALTERNATIVE shortcut and apply.

Now it works as desired!

wismill:
# Context

This is an alternative (and final version) of #527 (previous attempts: #485 and #446).

> [!IMPORTANT]
> This is the fix of the dreaded “*need to kick hotkeys on release, not press*”.

# Changes

## New parameter `lockOnRelease` for `LockGroup()`

It enables to use e.g. the combination `Control + Shift` *alone* to switch layouts, while keeping the use of `Control + Shift + other key` (typically for keyboard shortcuts).

The specification of `LockGroup()` is changed to:
- On key *press*:
  - If `lockOnRelease` is set, then key press has no effect.
  - Otherwise:
    - if the `group` is absolute, key press sets the *locked* keyboard group to `group`;
    - otherwise, key press adds `group` to the *locked* keyboard group.

  In either case, the resulting *locked* and *effective* group is brought back into range depending on the value of the `GroupsWrap` control for the keyboard.
- On key *release*:
  - If `lockOnRelease` is not set, then key release has no effect.
  - Otherwise, if any other key was *pressed* after the locking key, then key release has no effect.
  - Otherwise, it has the same effect than a key press *without* `lockOnRelease` set.

This is really useful for people coming from other platforms, such as Windows.

> [!TIP]
> It fixes a [20-year old issue] inherited from the X11 ecosystem, by extending the [XKB protocol key actions].

> [!IMPORTANT]
> As it is incompatible with X11, this feature is available only using the keymap text format **v2**.

[20-year old issue]: https://gitlab.freedesktop.org/xorg/xserver/-/issues/258

## New paramter `unlockOnPress` for `LockMods()`

It enables e.g. to deactivate CapsLock on press rather than on release, as in other platforms such as Windows.

- On key *press*:
  - If `unlockOnPress` is true and some of the target modifiers were *locked* before the key press, then unlock them if `noUnlock` false.
  - Otherwise:
    - add target modifiers to *depressed* modifiers;
    - if `noLock` is false, add target modifiers to the *locked* modifiers.
- On key *release*:
  - If `unlockOnPress` is true and triggered unlocking on key press, do nothing.
  - Otherwise:
    - remove modifiers from the *depressed* modifiers, if no other key that affect the same modifiers is down;
    - if `noUnlock` is false and if any target modifiers was locked before the key press, *unlock* them.

> [!TIP]
> It fixes a [12-year old issue] inherited from the X11 ecosystem, by extending the [XKB protocol key actions].

> [!IMPORTANT]
> As it is incompatible with X11, this feature is available only using the keymap text format **v2**.

[12-year old issue]: https://gitlab.freedesktop.org/xorg/xserver/-/issues/312
[XKB protocol key actions]: https://www.x.org/releases/current/doc/kbproto/xkbproto.html#Key_Actions

# Bugs fixed

Fixes #372
Fixes #420

# Superseded MRs

Closes #527
Closes #485
Closes #446

@whot @bluetech @fooishbar @mahkoh


wismill:
> > As it is incompatible with X11, this feature is available only using the keymap text format v2.
> 
> How is this versioning scheme going to work? Are you planning to create a new version every few months?

@mahkoh Good question! No, I plan to use v2 as a *living standard* for now, based on the current syntax. It’s clearly not ideal, but I just have not enough resources to design the formal spec of a new format.

I have a few more ideas I would like to implement in 1.11. Please feel free to share your ideas as well!

What is sure is that xkeyboard-config will now target specific ranges of versions of libxkbcommon, and it will bump its major version each time it starts to use a feature requiring a new lower bound on the libxkbcommon version.

mahkoh:
>@mahkoh Good question! No, I plan to use v2 as a living standard for now, based on the current syntax.

I don't know if you've changed this, but before, the parser would outright abort if it encountered an unknown field. How does this work with a living standard?

wismill:
> I don't know if you've changed this, but before, the parser would outright abort if it encountered an unknown field. 

It is unchanged.

> How does this work with a living standard?

For now there is no plan for such forward-compatibility support. I count on distro keeping xkeyboard-config and libxkbcommon in sync.

The main issue is Flatpak & co. But for now Compositors are serializing to v1 anyway, while they can parse using v2. Fallback should not be a pb (client are not expected to run the actions) except for the extended layout indexes.

PR #796, opened 2025-06-14
Add support for layout-specific options

### LINKED ISSUE #500: RMLVO: support for applying option on specific layout index

wismill:
# Rules: Add support for layout-specific options

Enabled specifying a layout index for each option, so that it applies only if the layout matches. The layout index is specified by appending immediately after the option name the `!` specifier delimiter and then the layout index, in decimal form and 1-indexed.

Note that `!` was chosen instead of the usual `:` specifier delimiter, because some options contains `:`, e.g. `grp:menu_toggle`. `!` *cannot* clash with component names, because `!` is a token in the rules files and thus cannot be used as in component names. It is also vaguely similar to `:`, compared to e.g. `@` or `#`.

Example: given the following rules:

    ! layout[any]   option  =  symbol
      *             opt1    =  +s1:%i
      l2            opt2    =  +s2:%i

it may result in the following configurations:

| Layout   | Option   | Symbols      |
| -------- | -------- | ------------ |
| `l1`     | `opt1`   | `+s1:1`      |
| `l2`     | `opt1`   | `+s1:1`      |
| `l1`     | `opt2`   | ``           |
| `l2`     | `opt2`   | `+s2:1`      |
| `l1,l2`  | `opt1`   | `+s1:1+s1:2` |
| `l1,l2`  | `opt1!1` | `+s1:1`      |
| `l1,l2`  | `opt1!2` | `+s1:2`      |
| `l1,l2`  | `opt2`   | `+s2:2`      |
| `l1,l2`  | `opt2!1` | ``           |
| `l1,l2`  | `opt2!2` | `+s2:2`      |

# RMLVO builder: Add layout-specific option

Modified the signature of `xkb_rmlvo_builder_append_layout()` to accept an optional array of options.

We can do this because the function has not been yet published in a release.

# Registry: Add `rxkb_option_is_layout_specific()`

Enable to query if an option accepts layout index specifiers to restrict its application to the corresponding layouts.

Fixes #500


wismill:
@whot proposed to add a proper [builder API for RMLVO values](https://github.com/xkbcommon/libxkbcommon/issues/500#issuecomment-2309247372) too.

wismill:
The regsitry DTD is updated with:

```diff
  <!ELEMENT option (configItem)>
  <!ELEMENT configItem (name, shortDescription?, description?, vendor?, countryList?, languageList?, hwList?)>
+ <!ATTLIST configItem layout-specific (true|false) "false">
  <!ATTLIST configItem popularity (standard|exotic) #IMPLIED>
```

I wonder if we should introduce a child tag instead, so that it’s easier to extend the feature (e.g. layout range, etc.).

CC @mahkoh

wismill:
[Sergey agrees](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/merge_requests/830#note_2958904) with the simple `layout-specific` attribute in the registry DTD.

wismill:
For now tools will support this feature using the `!` layout index specifier.

The [new CLI arguments suggested](https://github.com/xkbcommon/libxkbcommon/issues/500#issuecomment-2309282184) by @whot would require further work:

```bash
xkbcli compile-keymap --layout-1 us --variant-1 dvorak --option-1=banana --option-1=orange
```

PR #803, opened 2025-06-18
xkbcli-interactive-wayland improvements

wismill:
- Fix memory leak. I am not sure this is the right fix. I am puzzled by how it works.
- Added window decoration, if supported. It’s not much, but it much more user-friendly than a blank square. I am not sure if it is supported by all DE though, e.g. Gnome use to be quite loud against it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -577,7 +577,7 @@ if build_tools
     # Tools: interactive-wayland
     if get_option('enable-wayland')
         wayland_client_dep = dependency('wayland-client', version: '>=1.2.0', required: false)
-        wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.12', required: false)
+        wayland_protocols_dep = dependency('wayland-protocols', version: '>=1.15', required: false)
         wayland_scanner_dep = dependency('wayland-scanner', required: false, native: true)
         if not wayland_client_dep.found() or not wayland_protocols_dep.found() or not wayland_scanner_dep.found()
             error('''The Wayland xkbcli programs require wayland-client and wayland-protocols which were not found.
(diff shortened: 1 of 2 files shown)


whot:
> Fix memory leak. I am not sure this is the right fix. I am puzzled by how it works.

fix LGTM (though I haven't written that part of wayland code before). The api doc for `wl_surface_attach` explains it though but basically: the buffer is used until you either get the `wl_buffer.release` event from the compositor or you replace the buffer yourself with another attach. In both cases can you free that buffer, but not before.

> gnome use to be quite loud against it.

GNOME won't support SSD so you'll get the same as before, just a undecorated window. Meh?

meson.build
```diff
@@ -601,11 +601,17 @@ You can disable the Wayland xkbcli programs with -Denable-wayland=false.''')
             wayland_scanner_code_gen.process(xdg_shell_xml),
             wayland_scanner_client_header_gen.process(xdg_shell_xml),
         ]
+        xdg_decoration_xml = wayland_protocols_datadir/'unstable/xdg-decoration/xdg-decoration-unstable-v1.xml'
+        xdg_decoration_sources = [
+            wayland_scanner_code_gen.process(xdg_decoration_xml),
+            wayland_scanner_client_header_gen.process(xdg_decoration_xml),
+        ]
```
whot, line 608: Fairly recent additions but just fyi: the [meson wayland module](https://mesonbuild.com/Wayland-module.html) is now stable so long term we should switch to using that.

whot (approved):

meson.build
wismill: Good point, create #804 for long-term.

What is our policy related to min meson requirement? I have a pending patch to bump it to 1.1 (April 2023). I used a similar period (least 2 years old) than the previous bump.

PR #812, opened 2025-06-22
test: Add example for CapsLock breaking latches

wismill:
Consider the following use case: if `Caps_Lock` is on the second level of some key, and `Shift` is latched, pressing the key locks `Caps` while also breaking the `Shift` latch, ensuring that the next character is properly uppercase.

This MR illustrates how to fix the issue presented in #51 but using the current features: multiple actions per level and `VoidAction()`.

The advantage compared to #51 and #569 is that the behavior is *opt-in*: it only depends on the keymap and avoid special handling in the code.

CC @Jules-Bertholet 

TODO:
- [ ] Write doc about this feature. Maybe inaugurate an “advanced keymap features” page, else add an entry in the FAQ.


Jules-Bertholet:
I don’t think this should be opt-in. The current behavior has no advantages for users. “Do the right thing” should be the default, not something that can be achieved only with manual tweaking of configuration files, that the overwhelming majority of users won’t do. (I’m thinking forward to a future where #596 is fixed, and users can enable latching on any layout just by flipping a switch in their DE.)

Also, this only works for changes to locks triggered by `LockMods()`. `latchToLock`/`clearLocks` should also break latches, but this trick doesn’t allow accomplishing that.

wismill:
> “Do the right thing” should be the default, 

@Jules-Bertholet “the road to hell is paved with good intentions”. So, are you are the righteous person who knows and understands *every* use case out there?

I am more humble and know that every time there is a new feature, it may enable new use cases but also introduce new edge cases. So I prefer to be conservative and develop general solutions once the problem is well identified and understood, rather than playing a whack-a-mole game with the code.

The solution proposed in this MR does solve the issue, using existing features, thus avoiding new edge cases.

I also added a new use case for the German E1 layout.

> not something that can be achieved only with manual tweaking of configuration files, that the overwhelming majority of users won’t do. (I’m thinking forward to a future where #596 is fixed, and users can enable latching on any layout just by flipping a switch in their DE.)

Your assumption seems that this feature will never be shipped by xkeyboard-config, which is wrong. 

> Also, this only works for changes to locks triggered by `LockMods()`. `latchToLock`/`clearLocks` should also break latches, but this trick doesn’t allow accomplishing that.

It does not make sense for *this* use case: one *cannot* trigger the latch-to-lock behavior of Shift!

Please provide a realistic use case where latch-to-lock should break a latch.

Jules-Bertholet:
> Your assumption seems that this feature will never be shipped by xkeyboard-config, which is wrong.

The issue here would be that this would make the `xkeyboard-config` layouts no longer compatible with X11/xkbcomp. But I suppose you have a plan for that?

> Please provide a realistic use case where latch-to-lock should break a latch.

There isn’t one (AFAIK), other than consistency. But consistency matters: it make the rules easier to understand for users.

> I also added a new use case for the German E1 layout.

This isn’t *quite* as powerful as #569, which only breaks latches for the modifiers whose latches need to be broken. This could have an effect if people want to set keyboard shortcuts involving the latched layer. But this is probably a really rare use-case.

wismill:
> > Your assumption seems that this feature will never be shipped by xkeyboard-config, which is wrong.
> 
> This would make the `xkeyboard-config` layouts no longer compatible with X11/xkbcomp. But I suppose you must have a plan for that?

[Indeed](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/blob/master/VERSIONING.md).

> > Please provide a realistic use case where latch-to-lock should break a latch.
> 
> There isn’t one (AFAIK), other than consistency. But consistency matters: it make the rules easier to understand for users.

If there is no current use case, then there is also no need for users to understand the *full* specification of latches in order to use the known use cases. Motivated users can still check the doc.

> > I also added a new use case for the German E1 layout.
> 
> This isn’t _quite_ as powerful as #569, which only breaks latches for the modifiers whose latches need to be broken. This could have an effect if people want to set keyboard shortcuts involving the latched layer. But that is probably a really rare use-case.

Shortcuts using the 5th level? Various DE and apps already struggle to map non-ASCII characters, even on the 1st level.

Jules-Bertholet:
> If there is no current use case, then there is also no need for users to understand the full specification of latches in order to use the known use cases. Motivated users can still check the doc.

I am thinking more of recovery from accidental key presses, versus deliberate use.

wismill:
> > If there is no current use case, then there is also no need for users to understand the full specification of latches in order to use the known use cases. Motivated users can still check the doc.
> 
> I am thinking more of recovery from accidental key presses, versus deliberate use.

This is a good use case, but if you want to cancel a latch like this, just type the *same* latch key, no? It is already possible. Why would you use another key for that?

wismill:
> > just type the same latch key, no?
> 
> Not if it’s a latch-to-lock key.
> 

So you wrote that latch-to-lock should break latches in order to enable:

> recovery from accidental key presses, versus deliberate use.

among other use cases already covered by this MR.

In the case of the accidental latch, if you use latch-to-lock: you obviously *lock* a modifier. Then:
1. Either it is the same modifier, it breaks the latch and then another press does unlock it, as I wrote above: “press the same latch key”.
2. Else it is another modifier, and then applying you proposal effectively have the accidental latch broken but *also* have a possibly *unwanted* locked modifier.

I do not see how locking a modifier to cancel a *different* latch (case 2) is any more intuitive than to just locking and unlocking the *same* accidental latch (case 1). 

For the matter, simply pressing any key that breaks a latch is also a good solution, let us not forget that. It’s also very likely to be what users will do by accident, without thinking, unless very carefully using *search-and-peck* typing.

So I am not against the feature per see, but I need to see a compelling use case.

> But, anyway, my point is not so much “this would provide useful capabilities to users” but “this would be less surprising for users”. I.e. Principle of Least Astonishment.

While I agree this principle is a good directive, I do not see where it applies in your example.

---

In fact, I think that the latch-to-lock feature may actually *not* be what people want: it may be more useful to have it deactivated and *make a second tap on the same latch modifier actually break the pending latch* the most useful behavior. 

Apart from CapsLock and NumLock, it does not seem that usual to lock modifiers. This is a wild guess though, so I would not change the current keymap unless checking it with knowledgeable accessibility people.

Jules-Bertholet:
> Way too much niche 

I agree that it’s niche. My point is not “users will demand this feature because of this use-case”, but instead “users who find themselves in this unusual situation may be surprised by unintuitive behavior”. It’s a minor point, and I totally understand if you think it is outweighed by e.g. implementation complexity or X11 compatibility.

wismill:
> users who find themselves in this unusual situation may be surprised by unintuitive behavior

As written before, I have yet to see real-life situations where this is unintuitive. 

Any accidental press on a latch will be unintuitive to fix, if you ask me, unless there are indicators of some sort to show the modifiers state. Without them, I will be in the “[wat?](https://www.destroyallsoftware.com/talks/wat)” group, just next the “what the f… with my mods?!” user group.

> But everything I have said about latched/locked mods applies to groups as well.

I agree that group latch should behave similarly to mods ones. Yet the same comments apply.  Imagine the following use case, a variant from the initial one: 

```c
key <LFSH> { [ ISO_Level2_Latch, ISO_Group_Latch ] };
```

What should we do here? It seems unlikely to (always) want to keep the `Shift` modifier latched after the pressing the group latch, isn’t it?

My take is that we can apply the same fix than the modifiers:

```c
key <LFSH> {
  // I used a group latch for a change, but a group lock is a legit use case too.
  [ ISO_Level2_Latch, ISO_Group_Latch ],
  [ LatchMods(mods=Shift), {VoidAction(), LatchGroup(group=+1)} ]
};
```

Only that contrary to modifiers, it seems very, *very* unlikely to use simultaneous group latches. And using a modifier latch-to-lock to cancel a group latch seems… all but intuitive.

---

Also, I realize that I have not presented a use case where having locks breaking latch is unwanted.

1. Imagine one uses a layout with accentuated letters on the 3rd level. They want to type ALL CAPS (or simply use CapsLock instead of Shift because some users prefer it). Should they first lock Caps or should they latch Level3? I do not think we have to decide for them: the order should not matter here.
2. Imagine one has 2 groups, `xx` and `yy`, the later being currently locked. The user want to type in the `xx` group starting with an upper case. Should we force them to first switch the group, then latch the Shift modifier? Or could we just let them press the corresponding keys in whatever order, especially if those are close to each other?

These use cases seem realistic to me. *If* the order of the keys would matter, then I claim that it would actually break the *Principle of Least Astonishment*, because there is no requirement in those use cases.

There are most probably a lot more of use cases than what we discussed here or in the various issues/MR. I would rather fix the keymaps locally using the `VoidAction()` feature[^1] than implementing a *general* locks-break-latches feature.

[^1]: It’s not a trick: I added it in part due to your feedback that there were no easy way to break latches 😄. The other part of the rationale was to mirror the `VoidSymbol` feature for actions.

Jules-Bertholet:
> These use cases seem realistic to me.

Those don’t match my intuition, I would expect to have to lock before latching. But if my intuition is not universal, that certainly weakens the case for my proposed changes.

> use CapsLock instead of Shift because some users prefer it

I expect most such users would actually prefer a latching Shift, but either don’t know about it or don’t want to deal with the hassle of configuring it.

wismill:
> > These use cases seem realistic to me.
> 
> Those don’t match my intuition, I would expect to have to lock before latching. But if my intuition is not universal among latching users, that certainly weakens the case for my proposed changes.

Better be humble by default when dealing with UX 😉

> 
> > use CapsLock instead of Shift because some users prefer it
> 
> I expect most such users would actually prefer a latching Shift, but either don’t know about it

You would be surprised: see #372.

>  or don’t want to deal with the hassle of configuring it.

Excluding users, is that it? This is not the policy of this project nor of xkeyboard-config.

Jules-Bertholet:
> Excluding users, is that it? This is not the policy of this project nor of xkeyboard-config.

To be clear, “I don’t want to deal with the hassle” is perfectly valid/sensible. Especially as a user may switch between many devices and platforms. It’s certainly something we need to account for, I don’t disagree.

PR #816, opened 2025-06-25
State: support sticky keys

### LINKED ISSUE #596: Missing Sticky Keys support

wismill:
- Added API to change global key controls of the keyboard state:
  - new enumeration `xkb_keyboard_controls`
  - new function `xkb_state_update_controls()`
  - new function `xkb_state_serialize_controls()`
  - new member `XKB_STATE_CONTROLS` in the `xkb_state_component`
    enumeration.
- Added support for `SetControls()` and `LockControls()`
- Added support for Sticky Keys

> [!important]
> I do not plan to add support for global key controls into Wayland. This is a compositor-only feature.

Fixes #596

TODO:
- [x] Split the MR: general controls support + sticky keys. See #917.
- [x] Fix LatchToLock, which is not a control but a control option.
- [x] Implement SetControls()
- [x] Fix clearing all latches and locks on sticky keys disablement.
- [x] Fix X11 API requiring access to private `state::controls` fields for initialization/update.
- [x] Add `--controls` arg to all relevant tools.
- [x] Support `XkbAX_TwoKeys` control? See #918 for a dedicated discussion.
- [x] Changelog
- [ ] Doc: describe controls, their support compared to Xorg, etc.
- [x] Add basic tests
- [x] Add tests with latch-to-lock
- [ ] Do the Wayland compositors implement sticky key differently than XKB? If so we probably need a flag to customize the behavior. See [this discussion](https://github.com/xkbcommon/libxkbcommon/pull/611#issuecomment-2613006951). Better use dedicated issue & MR.

---

It may not be that difficult to support sticky keys after all, so optimistically planning it for ~1.11~ 1.14.


bluetech:
Once this is ready for testing, I'd enjoy testing it in e.g. interactive-evdev if you can write a little "guide" on how to do it, like the xkeyboard-config options and key combinations and such.

wismill:
- Rebased on #913. 
- Changed general state flags to be specific accessibility flags, so that we may add other kind of flags.
- Added latch-to-lock test
- Fixed Controls handling: private vs. public API
- Some refactoring

wismill:
@bluetech this is very close to be ready, now that there are sufficient tests. I will probably split the MR into, moving the introduction of Controls handling a dedicated MR.

An important TODO is to check if we need further flags to have sticky keys behave like in the current Wayland compositors, which may have adopted non-XKB behavior. But a follow-up MR may be more adapted.

There are still some minor TODOs.

Things I am not sure about:
- Implement `XkbAX_TwoKeys`: I find the feature debatable and would like to confirm its utility with experts. I will probably leave it for a further MR, if relevant.

wismill:
Rebased on #917, so that the current MR only deal with sticky-keys.

PR #825, opened 2025-07-15
Add xkb_rmlvo_builder_ref()

wismill:


src/rmlvo.c
```diff
@@ -136,6 +136,13 @@ xkb_rmlvo_builder_append_option(struct xkb_rmlvo_builder *rmlvo,
     return true;
 }
 
+struct xkb_rmlvo_builder *
+xkb_rmlvo_builder_ref(struct xkb_rmlvo_builder *rmlvo)
+{
+    rmlvo->refcnt++;
```
whot, line 142: best to add an `assert(rmlvo->refcnt > 0)`. Right now (while buggy) this would typically work:
```
xkb_rmlvo_builder_unref(builder); /* last ref, rmlvo->refcnt set to 0 */
xkb_rmlvo_builder_ref(builder); /* memory has been freed but refcnt is still on 0 */
```
we might as well blow up in that case.

And speaking of which, adding an `assert(rmlvo->refcnt > 0);` to unref would provide the same basic security.

whot (approved):

src/rmlvo.c
wismill: Added the assertions, which are also a way to document the role of the `refcnt` member. I added them also for all the other ref functions for consistency.

But we will never manage to evaluate them anyway: the object is always freed just after reaching `refcnt = 0`, so any use will always lead to a use-after-free error.

PR #838, opened 2025-08-07
`xkbcli how-to-type`: Add Compose support

### LINKED ISSUE #361: Determining required dead keys

wismill:
- Enable Compose support by default, disable it with `--disable-compose`
- Some filters are used to avoid combinatorial explosion

I do not think we need an API, since the filters used may depend on the use case. But a JSON/YAML output would be a good idea.

Fixes #361


johngebbie:
I haven't got round to trying to implement something for my case with this yet, but it looks good, and I just want to say I appreciate your work. Thank you!

PR #848, opened 2025-08-16
Compose: custom locale fallback

wismill:
Previously, loading a Compose file based on the locale would fail if the locale is not in the X11 Compose locale registry. While there are workarounds (e.g. `~/.XCompose` file or `$XCOMPOSEFILE`), it does not work if the corresponding file has `include "%L"`.

This commit adds the fallback `en_US.UTF-8` in case the locale is installed but not registered in the X11 Compose locale registry.

> [!NOTE]
> Ideally we should have a mechanism to extend the Compose locale registry at the *system* level. Mechanisms at the *user* level (e.g. custom Compose file, environment variable) are deemed sufficient.

Unfortunately is not possible to test it in our test suite. One can still check it works following these instructions:

1. Create a custom locale, e.g. `en_XX.UTF-8`. `glibc-i18ndata` or similar package may be required to get the required files in `/usr/share/i18n/`.

   `sudo localedef -i en_US -f UTF-8 en_XX.UTF-8`
2. `xkbcli compile-compose --verbose --locale en_XX.UTF-8`

Closes #318

@JSkold @bluetech


wismill:
> I could imagine more elaborate fallbacks (like matching language first) but very sure it's not worth it.

@bluetech I agree, but I did not found a function in stdlib/glibc to manipulate the locale. While the format XPG is simple enough, I think it is best to avoid messing with locales manually.

PR #863, opened 2025-10-06
Fix invalid macro in xkbcli-how-to-type.1

pbrisbin:
`.Be` is not a valid macro. I believe the intent was to end the list opened with `.Bl`. The macro for that is `.El`, as used elsewhere.

https://mandoc.bsd.lv/man/mdoc.7.html#Displays_and_lists


pbrisbin:
Thanks for the quick merge!

PR #876, opened 2025-10-17
context: Add support for XKB extensions directories

wismill:
*(EDIT: 2025-10-30 this is the updated proposal; see the original at the end of the description)*

libxkbcommon introduced flexibility by enabling multiple XKB lookup paths. However, there is no easy way to distribute packages that provide XKB data: such packages usually resort to modify the root XKB directory, which is very fragile.

This can be solved by adopting the `.d` directory approach used in e.g. `/etc`.

Added another XKB path lookup mechanism: the XKB extensions directories. There are 2 root extensions directories, both optional:
- `XKB_CONFIG_UNVERSIONED_EXTENSIONS_PATH`, defaulting to
  `<xkeyboard-config-root>` - version suffix + `.d`, e.g.
  `/usr/share/xkeyboard-config.d`. It is intended as the default
  extensions path.
- `XKB_CONFIG_VERSIONED_EXTENSIONS_PATH`, defaulting to
  `<xkeyboard-config-root>` + `.d`, e.g. `/usr/share/xkeyboard-config-2.d`.
  It is intended for backward compatibility.

`XKB_CONFIG_VERSIONED_EXTENSIONS_PATH` has higher priority than `XKB_CONFIG_UNVERSIONED_EXTENSIONS_PATH`.

When appending the default include paths, after the extra path but before the canonical XKB root, for each directory above: its subdirectories are appended in lexicographic order to the include paths; unversioned subdirectories are appended only if there is no corresponding versioned subdirectory.

Example: given the following setup:
- `XKEYBOARD_CONFIG_ROOT` = `/usr/share/xkeyboard-config-2`
- `/usr/share/xkeyboard-config-2.d` tree:
  - `p1` (installed by package p1)
    - `rules/evdev.xml`
    - `symbols/a`
  - `p2` (installed by package p2)
    - `rules/evdev.xml`
    - `symbols/b`
- `/usr/share/xkeyboard-config.d` tree:
  - `p1` (installed by package p1)
    - `rules/evdev.xml`
    - `symbols/a`
  - `p3` (installed by package p3)
    - `rules/evdev.xml`
    - `symbols/c`

then the include path list may be:
1. `~/.config/xkb`
2. `~/.xkb`
3. `/etc/xkb`
4. `/usr/share/xkeyboard-config-2.d/p1`
5. `/usr/share/xkeyboard-config-2.d/p2`
6. `/usr/share/xkeyboard-config.d/p3`
7. `/usr/share/xkeyboard-config`

In this example:
- Package p1 distributes a layout with specific compatibility files for
  xkeyboard-config v2 and fallback files for v3+. In this case the
  versioned files have priority, so the unversioned
  `/usr/share/xkeyboard-config.d/p1` path is not included.
- Package p2 distributes only a layout specific to xkeyboard-config v2.
- Package p3 distribute a layout for any xkeyboard-config version.

---

TODO:
- [x] Implement the registry bits + tests
- [x] Rebase on #903
- [ ] Check support for OS other that Linux

---

<details><summary>Initial proposal</summary>

libxkbcommon introduced flexibility by enabling multiple XKB lookup paths. However, there is no easy way to distribute packages that provide XKB data: such packages usually resort to modify the root XKB directory, which is not good.

This can be solved by adopting the `.d` directory approach used in e.g. `/etc`.

Added another XKB path lookup: each subdirectory of the XKB extensions directory, listed in lexicographic order. The extensions directory is defined by the environment variable `XKB_CONFIG_EXTENSIONS_PATH` or defaults to the system XKB extension directory, defined at library configuration time (usually `/usr/share/xkeyboard-config-<VERSION>.d`).

---

The default directory naming is subject to bikeshedding. I would like to:
- Keep it open to other XKB implementations.
- Related to the xkeyboard-config file it may import. I introduce the `xkb_extensions_path` variable in the pkg-config file, in hope that packages distributing keyboard layouts use it for their installation directory, e.g. `<xkb_extensions_path>/<package_name>`.

</details>

@whot @bluetech @mahkoh

### DIFF
--- a/changes/build/+extensions-dirs-pkg-config.feature.md
+++ b/changes/build/+extensions-dirs-pkg-config.feature.md
@@ -1,0 +1,8 @@
+Added the following variables to the pkg-config file, in order to expose the
+default XKB directories:
+- `xkb_root`
+- `xkb_unversioned_extensions_path`
+- `xkb_versioned_extensions_path`
+- `xkb_extra_path`
+
+See @ref packaging-keyboard-layouts "" for further details.
--- a/meson.build
+++ b/meson.build
@@ -433,12 +433,23 @@ dep_libxkbcommon = declare_dependency(
     include_directories: include_directories('include'),
 )
 meson.override_dependency('xkbcommon', dep_libxkbcommon)
+pkgconfig_variables = [
+    f'xkb_root=@XKB_CONFIG_ROOT@',
+    f'xkb_extra_path=@XKB_CONFIG_EXTRA_PATH@',
+]
+if has_extensions_directories
+    pkgconfig_variables += [
+        f'xkb_unversioned_extensions_path=@XKB_CONFIG_UNVERSIONED_EXTENSIONS_PATH@',
+        f'xkb_versioned_extensions_path=@XKB_CONFIG_VERSIONED_EXTENSIONS_PATH@',
+    ]
+endif
 pkgconfig.generate(
     libxkbcommon,
     name: 'xkbcommon',
     filebase: 'xkbcommon',
     version: meson.project_version(),
     description: 'XKB API common to servers and clients',
+    variables: pkgconfig_variables,
 )
 
 


mahkoh:
There is no package in the arch repos that has files in those directories except xkeyboard-config.

wismill:
To develop further the motivation of the MR: there are some layouts that are popular enough to have some community, yet is not in xkeyboard-config for some reason (motivation, project policy, etc.). So it does make sense to have a dedicated package rather than requiring each user to modify `~/.config/xkb` or `/etc/xkb`.

This is not currently possible without using ugly hacks in the xkeyboard-config data directory, thus this MR.

wismill:
> It means that all of these AUR packages will break every time xkbcommon switches to a new directory 

@mahkoh It’s a good point. My initial idea was to ensure that layout packages are using the xkeyboard-config they are designed for, so placing them in a versioned dir that was retrieved from either the xkeyboard-config or libxkbcommon seemed a good idea… but only if the layout packages are rebuilt each time their dependency is modified, which is obviously not automatic in packages not managed by distros!

> (unless all older directories are also added at the end of the include path?).

No, the point is (was) to keep *by default* `XKB_CONFIG_EXTENSIONS_PATH` = `XKB_CONFIG_ROOT` + `.d` for consistency.

What about `/usr/xkb.d`? The `.d` is to differentiate with the approach of `XKB_CONFIG_EXTRA_PATH` which default to `/etc/xkb` and expect `/etc/xkb/symbols`, …, while `/usr/xkb.d` would expect `/usr/xkb.d/<some package>/symbols`, …. 

`/usr/xkb/extensions` would be fine too. 

Anyway those are default values and before further bikeshedding it would be great to agree on the concept @mahkoh.

heftig:
Depending on how breaking the changes from `xkeyboard-config-2` to `xkeyboard-config-3` (and beyond) look like, it might be a bad idea to provide a version-independent extension dir. If "extension broken" is much worse than "extension missing", extension packages will just have to supply config for each of the versions they support.

Is `/usr/xkb.d` a typo? I would have expected `/usr/share/xkb.d`, or better, `/usr/share/xkeyboard-config.d` if you want a version-independent dir.

wismill:
> Is `/usr/xkb.d` a typo? I would have expected `/usr/share/xkb.d`, or better, `/usr/share/xkeyboard-config.d` if you want a version-independent dir.

It is! Fixed, thanks

wismill:
I added a discussion #878 too. I will invite some maintainers of keyboard layout packages.

mahkoh:
>@mahkoh Please check the xkeyboard-config and Xorg project git history and trackers.

You can just tell me.

>I already explained this to you, yet you keep spreading FUD. V2 format is a superset of the V1 format, so it won’t suddenly stop working.

You have explained that all of the existing AUR packages will stop working without manual intervention:

>>(unless all older directories are also added at the end of the include path?).
>
>No

I was quoting @heftig who said that that and doing the same again next time might be advisable.

>I have no idea what you are talking about.

Apple requires app developers to always be on their toes and recompile their applications to comply with the latest apple innovations or else they will be removed from the app store.

In this comparison you are apple and app developers are people who provide keymaps in any way other than xkeyboard-config.

>And it is. Or you have an example where it is not?

If at any point xkbcommon stops adding /usr/share/X11/xkb to its default include path, all external keymaps installed there will stop working. As I understand it, doing so is the plan. This will break existing setups.

I had not considered the existence of such maps installed outside of $HOME before so it is good that you have brought them up. Adding a new directory for extension files has no bearing on this since existing setups don't use it.

Now that you are aware of it, maybe it is a good idea to rethink this plan if keeping existing setups working is the first priority. The current plan seems to have been based on the assumption that all keymaps and xkbcommon are distributed in lockstep.

Maybe always including all older directories but not allowing components in older directories to overwrite components from newer directories would be the solution. I have not given it much thought. 

>And? What’s your point? We do not fix decade-old bugs that require format change? We wait 10 years before any change?

As I've already said, keymaps should be boring with existing setups being the first priority. If that meant waiting 10 years, then waiting 10 years would be the correct choice.

But it doesn't. If all changes are only additions of new syntax, other languages add new syntax all the time without incrementing their major version. Take Java for example which has added new syntax to the language and new constructs to its bytecode for 30 years. And similarly other languages have the ability to detect the capabilities of the execution environment at the source code level and change their behavior accordingly. 

I have already granted that one breaking change with a new directory tree might be necessary because xkb was not designed for this. The primary objective of the next major version should then be to add this capability.

However, even that might not be necessary. In the past, multiple keysym per level were added as new syntax. I assume this was done by modifying xkbcomp in sync so that xkbcomp does not outright break when it encounters this syntax. The same could be done now but in a more coarse manner: Add a new syntax construct add the top of the file that xkbcomp recognizes and makes it ignore that file if it appears in the include path. That would be the last required change in xkbcomp. All existing xkeyboard-config files would have to be distributed forever in legacy versions for X11, but they could live in the same directories as the new files.

Since includes are based on file names, it would have to be designed. For example, newer xkb implementations could be taught that if files, "x" and "x.compat" exist, they will use the compat file when the "x" include is requested. The plain "x" file would contain only legacy syntax for xkbcomp and "x.compat" would be the new file. Such a scheme would also fail grecefully because the legacy file would still provide the existing functionality.

That is just one possibility. I don't know how much effort has been put into engineering schemes that have not creating a new directory tree as an absolute requirement.

>It is super hard to get the community involved.

I don't think this kind of work is really something that would be made easier with more cooks in the kitchen. And I wouldn't expect the average keyboard user to get excited about designing a language extension that allows features to be added in a backwards compatible way in the future.

However, it seems like a paradox. If the community does not care, then why is there a need to add new syntax to xkeyboard-config in the first place? If xkbcommon has the capability, such users that do care can always store keymaps with the new syntax under $HOME.

As an aside, some users have asked for more support of existing xkb language feature in xkbcommon and those users could be helped without changing any syntax. How the numbers compare I do not know.

My guess would be that all of these issues have been open for decades because the need is really not that big. Not big enough to not be able to wait another few years at least. Until such a time when a solution has been engineered that will make the next xkb version the last one until it is replaced by a wholly new language.

>So please help and make proposals like your previous contributions rather than spread FUD.

I find it unfortunate that you have positioned yourself as the person in control of a decades old standard and react this way to pushback.

wismill:
Rebased and updated with my latest proposal.

Issue #914, opened 2025-11-06
Windows-style `LevelThree` as `Ctrl+Alt`


wismill:
On Windows the *default* third-level modifier is actually `Ctrl+Alt`, while Linux has a *dedicated* modifier: `LevelThree`.

While most people would probably use only the `AltGr` key as third-level modifier, some prefer to use the key combination `Ctrl+Alt`, either because they do have *symmetric* `Alt` keys or because they simply expect `AltGr` and `Ctrl+Alt` key combos to be strictly equivalent.

See some use cases on the xkeyboard-config project:
- [xkeyboard-config#33](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/33)
- [xkeyboard-config#482](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/482)
- [xkeyboard-config#555](https://gitlab.freedesktop.org/xkeyboard-config/xkeyboard-config/-/issues/555)

PR #922, opened 2025-11-11
Add state event API

wismill:
Add the **state event API**, which is required to implement some XKB features, e.g. `RedirectKey()` (see #919) and overlays, but also new features such as Windows-style 3rd level choser `Ctrl+Alt`, see #914.

New API:
- `struct xkb_state_machine`:
  - `xkb_state_machine_new()`
  - `xkb_state_machine_ref()`
  - `xkb_state_machine_unref()`
  - `xkb_state_machine_get_keymap()`
  - `xkb_state_machine_update_latched_locked()`
  - `xkb_state_machine_update_controls()`
  - `xkb_state_machine_update_key()`
- `struct xkb_event_iterator`:
  - `xkb_event_iterator_new()`
  - `xkb_event_iterator_destroy()`
  - `xkb_event_iterator_next()`
- `struct xkb_event`:
  - `xkb_event_get_type()`
  - `xkb_event_get_keycode()`
  - `xkb_event_get_changed_components()`
  - `xkb_event_serialize_controls()`
  - `xkb_event_serialize_mods()`
  - `xkb_event_serialize_layout()`

This is making the API much larger, but it seems unavoidable.

> [!note]
> The change set seems huge, but it is mostly due to tools, doc, and test updates.
>
> The new API implementation itself is relatively small, because we reuse most of the current `xkb_state` API.

---

TODO:
- [ ] Validate design
- [ ] Improve naming?

@bluetech @whot

### DIFF
--- a/changes/tools/+state-event-api.feature.md
+++ b/changes/tools/+state-event-api.feature.md
@@ -1,2 +1,2 @@
-`xkbcli interactive-{evdev,wayland}`: Added support for the new state event API.
+`xkbcli interactive-{evdev,wayland,x11}`: Added support for the new state event API.
 The legacy state API can be used instead with by passing `--legacy-state-api`.
--- a/tools/interactive-x11.c
+++ b/tools/interactive-x11.c
@@ -48,6 +48,8 @@ struct keyboard {
 
     struct xkb_keymap *keymap;
     struct xkb_state *state;
+    struct xkb_state_machine *state_machine;
+    struct xkb_event_iterator *state_events;
     struct xkb_compose_state *compose_state;
     int32_t device_id;
 };
@@ -58,6 +60,7 @@ static enum xkb_keymap_format keymap_format = DEFAULT_OUTPUT_KEYMAP_FORMAT;
 static enum xkb_keymap_serialize_flags serialize_flags =
     (enum xkb_keymap_serialize_flags) DEFAULT_KEYMAP_SERIALIZE_FLAGS;
 #else
+static bool use_events_api = true;
 static enum print_state_options print_options = DEFAULT_PRINT_OPTIONS;
 static bool report_state_changes = true;
 static bool use_local_state = false;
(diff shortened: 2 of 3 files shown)


wismill:
Rebased + renamed `xkb_state_event*` to simply `xkb_event*`.

wismill:
@bluetech @whot I am going to merge this MR to enable to work on further MR that depend on it.

> [!note]
> It would still be great to get some feedback in the aftermath, because this is quite an important change in the API!

wismill:
@bluetech Thanks for the feedback. I answer some of the points now, and will file a MR soon to address the others.

> I am curious myself, what is the reason for adding a new API, instead of keep adding to `xkb_state`. Is it a functional requirement, or just to have a clean separation between client and server?
> 

A clean separation:

- Ensures that there is no risk to mix the keyboard state at the *current*
  processed event and the state of the machine that populated the corresponding
  event queue, which may have been e.g. updated since the event generation.
- Add further features without modifying `xkb_state`, which is already
  bloated for *clients* applications.

> 
> What is the purpose of `xkb_state_update_from_event`? If `xkb_state_machine` is what servers should now use, why add a new server method to the client `xkb_state`?

It is for the case that the server needs to query the keymap and thus use the `xkb_state` API.

> 
> Similarly, I wonder if we should now revert the `xkb_state_options` etc. APIs, since they are not yet released but are meant for server applications, which should now use `xkb_state_machine`? I think it makes sense to reserve new functionality to the new API.

I thought about it too, but I remained hesitant since the `xkb_state_options` API was added just before the current MR. Should I delete `xkb_state_update_controls()` too?

> 
> I think the name `xkb_event_iterator` is a bit misleading -- my expectation for an iterator is it goes over some fixed thing, but here new events are always arriving. A name that would match my expectations more would be `xkb_event_stream` or `xkb_event_queue`.

It does iterate over a *fixed* list of events: the call to `xkb_state_machine_update_key()` *resets* the given iterator and adds the events corresponding to the input key event.

I chose this for simplicity to avoid to impose a specific implementation of a stream and avoid handling async operations.

The design space is vast for streams: it would be great to leave the decision to the user. Anyway, have you something in mind in order to improve the API?

> 
> In the wayland server example, in the event type switch-case default case, you wrote "". That make me think about the forward-compatibility behavior of adding new event types. Will new event types be added freely, or only on new keymap versions? 

I would say freely, but I was hoping to address the last X11 actions types in this release, so that new event types should not be necessary for a great while.

> If freely, should we guarantee that new event types won't break compat and are safe to ignore (i.e. should _not_ report unhandled event), just possibly missing some new ancillary functionality?

I would say they are safe to ignore, except if the server uses a `xkb_state` which must be kept in sync. Calling `xkb_state_update_from_event()` at each event should be enough in this case.

> 
> I wonder if `xkb_state_machine` doesn't need APIs to get keysym, utf8, layout, etc? Is the assumption that a server doesn't need those? Or if it does, should maintain an `xkb_state` as well?

I thought that maintaining a `xkb_state` would avoid to duplicate the whole `xkb_state` API for `xkb_state_machine`.

> 
> What is the reason to have `--legacy-state-api` in the tools? I guess for testing and example code?

Mainly testing, yes. Examples should be updated to the (now default) event API.

> 
> And as a suggestion, I think it would be good to release an rc1 version for 1.14.0, and try to get a compositor to use the new API, as an exercise for the new API.

Good idea. I may try to tweak Weston and motivate the KDE team.

PR #929, opened 2025-11-18
Shortcut tweak v3

### LINKED ISSUE #753: Persian sub layout English is not matching the variant
Hi!
I using us colemak mod-dh keyboard layout and Persian as my second layout but the problem is the layout of the short cuts or other things in Persian is not matching like ctrl+z is b in colemak mod dh but it is the qwerty on Persian.

I'm open to contribute just give me an hint

wismill:

Tweak keyboard shortcuts by adding the option to switch to a different group when some modifiers are active (typically `Control`, `Alt` or `Super`). The target group can be set for each group, although the most common use case would probably be the same target for all layouts.

It works by specifying:
- A *modifier mask* to watch, typically `Control`, `Alt` and `Super`, using `xkb_state_machine_options_shortcuts_update_mods()`.
- A *mapping*: source layout → target shortcuts layout, using `xkb_state_machine_options_shortcuts_set_mapping()`.

Then whenever the active modifiers contain *some* of the modifiers of the mask defined hereinabove, the active layout is switched to the target layout defined in the mapping, if any, otherwise it is left unchanged.

Also added the `--shortcuts-*` options to the interactive tools, e.g.:

```shell
xkbcli interactive-evdev --layout fr,us --shortcuts-mask "Control+Alt+Super" --shortcuts-mapping 1:2
```

Fixes #753
Closes #739

TODO:
- [x] Ensure the tweak is applied in all relevant code paths
- [x] Improve doc (FAQ, etc.)
- [x] Remove environment variable support

---

This is a refactor of #739, which uses an obsolete approach with Wayland issues.

### DIFF
--- a/tools/interactive-x11.c
+++ b/tools/interactive-x11.c
@@ -67,6 +67,7 @@ static bool use_local_state = false;
 static struct xkb_state_machine_options * state_machine_options = NULL;
 static enum xkb_keyboard_controls kbd_controls_affect = XKB_KEYBOARD_CONTROL_NONE;
 static enum xkb_keyboard_controls kbd_controls_values = XKB_KEYBOARD_CONTROL_NONE;
+static const char *raw_shortcuts_mask = NULL;
 static struct xkb_keymap *custom_keymap = NULL;
 #endif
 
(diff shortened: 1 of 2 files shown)


wismill:
To test it:

```shell
xkbcli interactive-evdev --layout fr,us --shortcuts-mask "Control+Alt+Super" --shortcuts-mapping 1:2
```

bam80:
@wismill thank you for your work on it, I didn't follow closely, though.

> The target layout can be set individually
for each source layout, although the most common use case would probably
be the same target for all layouts.

Do we really need such a complication? I can't imagine any real use-cases for even more layouts just for shotcuts than the user already has. Moreover, only one of them is usually useful, as you already mentioned.

PR #935, opened 2025-11-20
Add xkb_get_version()

### LINKED ISSUE #932: Add version query method (`xkb_get_version`)
There's currently no way to use multiple versions of libxkbcommon with standard dynamic linking. You're forced to use dlopen if you're going to support both Arch (on version 1.13.0 atm) and Ubuntu (on version 1.7.0).

This is very inconvenient because now you're forced to manually load all the symbols and check whether they exist to figure out which version you're on. There's versions that don't differ in exposed symbols though, so you're actually supposed to query the system package database (?) to deal with breaking changes. Or just assume you don't have access to functionality newer than 1.7.0.

In my case (Rust bindings) I want to bind `xkb_keymap_new_from_names2`, but return a `VersionNotSupported` error if the specified format is too new and fall back on `xkb_keymap_new_from_names`.

## Solution

Provide a global `xkb_get_version` function that returns a pointer to a static version triplet (e.g. `[uint8; 3]`) which would allow the libraries to query the installed version of the library on the system and change some logic based on that.

wismill:
Enable to apps to query the version of the libxkbcommon library actually in use and adapt their behavior: e.g. supported flags and available symbols, etc.

Fixes #932

---

TODO:
- [ ] Better commit message
- [ ] Documentation with examples (do and don’t)
- [ ] Changelog
- [ ] Add a flag arg to enable future developments?

### DIFF
--- a/include/xkbcommon/xkbcommon.h
+++ b/include/xkbcommon/xkbcommon.h
@@ -314,6 +314,16 @@ typedef uint32_t xkb_led_mask_t;
 #define xkb_keycode_is_legal_x11(key) ((key) >= 8 && (key) <= 255)
 
 /**
+ * Get the version of libxkbcommon in use.
+ *
+ * @returns a triplet representing MAJOR.MINOR.PATCH library version.
+ *
+ * @since 1.14.0
+ */
+XKB_EXPORT const uint8_t *
+xkb_get_version(void);
+
+/**
  * @defgroup rules-api Rules
  * Utility functions related to *rules*, whose purpose is introduced in:
  * @ref xkb-the-config "".
--- a/meson.build
+++ b/meson.build
@@ -324,6 +324,15 @@ yacc_gen = generator(
     output: ['@BASENAME@.c', '@BASENAME@.h'],
     arguments: ['--defines=@OUTPUT1@', '-o', '@OUTPUT0@', '-p', '_xkbcommon_', '@INPUT@'],
 )
+version_file = configure_file(
+    input: 'src/version.c.in',
+    output: 'version.c',
+    configuration: {
+        'MAJOR': xkbcommon_project_version[0],
+        'MINOR': xkbcommon_project_version[1],
+        'PATCH': xkbcommon_project_version[2],
+    },
+)
 libxkbcommon_sources = [
     'src/compose/constants.h',
     'src/compose/parser.c',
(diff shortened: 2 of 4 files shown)


Caellian:
I will go with symbol testing in the end. I see `xkb_get_version` as something useful only if functions like `xkb_keymap_is_supported_format` are missing for functionality that can't be probed through symbols or require e.g. constructing context+keymap+state to check whether something is supported.

Wayland allows requesting specific protocol versions, so I guess this would simplify some cases the same way it helps with protocol versioning in Wl. But I agree now that it's bad to rely solely on version to handle version differences in the library because that requires the app/bindings/wrapper to stay on top of all the changes to this library.

There's potential for misuse, but it could also pad mistakes. I suggest you mention in the docs "prefer other means (e.g. checking symbols) instead of relying on this value if possible".

whot:
Given that these are all my fault :) 

It's a mix and largely depends on how much effort I want to spend on versions that *don't* have those features. In the above examples: mtdev 1.1 is 15 years old and had some significant API changes back then. libwacom 0.27 is 8 years old and the only new required features are behind a function check. lua multiple versions is just a no-go, too much effort.

One counter-example for taking the checks to quite some extreme: https://gitlab.freedesktop.org/xorg/driver/xf86-input-libinput/-/blob/master/meson.build?ref_type=heads#L48.

Version checks are still useful for a minimum supported version - it's anything after that minimum that should be dynamic is where they tend to break down (esp. where only a single feature is required).

Caellian:
Alternatively, there could be `**char xkb_get_runtime_features()` that provides a list of supported function names, enum values, capabilities, as a list of static strings. Something like `"S xkb_rmlvo_builder", "F xkb_components_names_from_rules", ...`.

This would provide what dlopen use needs and would steer people away from testing version numbers.

I personally prefer this (though it increases the cost of checks), I just didn't propose it at first because it seemed like a large maintenance burden to impose. Given that #942 already uses clang to do something similar for enum values, I don't think it would be much more work to do something similar for exported functions and struct names. You'd only need to manually maintain a list of features that can't be guessed through function/struct availability (e.g. text v2 support if there wasn't already an enum value).

_Maybe_ use gperf to provide `bool xkb_has_runtime_feature("S xkb_rmlvo_builder")` to reduce runtime cost of string comparison.

PR #938, opened 2025-11-25
Make use of `counted_by` gcc/clang attribute

bluetech:
Since I've read about `counted_by` I wanted to try it, it seems to work as expected!

There are some preparatory commits, and the last commit contains the details.

@wismill I'll leave the decision whether to merge it to you. The main catch is that it requires some care that the len is always valid even during initialization (see commit for details). I see that CI already has a run with `-fsanitize=undefined`, so at least covered code should be safe (I fixed what it caught already), but uncovered code might not be.


wismill:
Rebased after merging the independent commits in #939.

wismill:
@bluetech that seems a very interesting feature. Are there any relevant runtime costs though?

PR #965, opened 2026-02-07
test: retain const qualifier from pointer

heitbaum:
Since glibc-2.43:

For ISO C23, the functions bsearch, memchr, strchr, strpbrk, strrchr, strstr, wcschr, wcspbrk, wcsrchr, wcsstr and wmemchr that return pointers into their input arrays now have definitions as macros that return a pointer to a const-qualified type when the in put argument is a pointer to a const-qualified type.

https://lists.gnu.org/archive/html/info-gnu/2026-01/msg00005.html

Resolves the following warnings:
    ../test/utils-text.c: In function 'uncomment':
    ../test/utils-text.c:82:18: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
       82 |     char *next = strstr(start, prefix);
          |                  ^~~~~~
    ../test/utils-text.c:94:14: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
       94 |         next = strchr(start, 0x0a);
          |              ^
    ../test/utils-text.c: In function 'split_lines':
    ../test/utils-text.c:122:14: warning: assignment discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
      122 |         next = strchr(start, 0x0a);
          |              ^

### DIFF
--- a/test/utils-text.c
+++ b/test/utils-text.c
@@ -80,7 +80,7 @@ uncomment(const char *input, size_t input_length, const char *prefix)
     const char *start = input;
     const char *end = input + input_length;
 
-    char *next = strstr(start, prefix);
+    const char *next = strstr(start, prefix);
     size_t count;
     while (start < end && next != NULL) {
         count = (size_t)(next - start);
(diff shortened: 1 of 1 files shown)


wismill:
Thanks!

PR #1003, opened 2026-03-24
build: include pre-generated parser files and manual regeneration target

wismill:
This change introduces pre-generated `parser.c` and `parser.h` files into the source tree and adds a manual mechanism to regenerate them.

- Added a `yacc-wrapper` script to ensure Bison runs from the project root. This guarantees that `#line`` directives in generated files use relative paths, keeping the source tree sanitized and portable.
- Implement a 'regenerate' custom target (not built by default) to allow developers to manually sync the parser after a `parser.y` update.
- Include the initial commit of the generated `parser.c`` and `parser.h` to allow out-of-the-box compilation without a yacc dependency.
- Do not regenerate the parser files in CI, except for some Linux setup.

This approach prevents “Bison lock-in” while maintaining the ability to update the grammar as needed.

Fixes #997

@whot @bluetech @vazub

TODO:
- [x] Add changelog entry
- [x] Add license for code from parser.y
- [x] Remove bison installation in CI

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -71,15 +71,25 @@ jobs:
         run: |
           # -gdwarf-4 - see https://github.com/llvm/llvm-project/issues/56550.
           CFLAGS="-gdwarf-4 -fno-omit-frame-pointer $CFLAGS_EXTRA" \
-              meson setup -Dbuildtype=debugoptimized -Denable-docs=true -Denable-cool-uris=true \
-                          $MESON_EXTRA_FLAGS \
+              meson setup -Dbuildtype=debugoptimized \
+                          -Denable-docs=true \
+                          -Denable-cool-uris=true \
+                          -Denable-parser-generation=$MESON_PARSER_GENERATION \
+                          $MESON_SANITIZE_FLAGS \
                           build
         env:
           CC: ${{ matrix.compiler }}
           # Use _FORTIFY_SOURCE only if not using sanitizers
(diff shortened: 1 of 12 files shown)


wismill:
@whot @bluetech a quick review would be greatly appreciated 🙂, since it is a significant change of the build.

bluetech:
I have a hazy memory that we used to do this once (we also used to use flex back then), then removed it. I believe Debian at least has a policy of generating the files itself (to provide the ability to modify the source code in its preferred form), so the ability will need to be maintained at least.

Overall this adds complexity for a rather thin benefit, as bison is pretty close to ubiquitous. But if you want to help @vazub avoid bison, that's OK. Ideally byacc would add the detailed error feature, then we could just restore byacc support...

wismill:
> I believe Debian at least has a policy of generating the files itself (to provide the ability to modify the source code in its preferred form), so the ability will need to be maintained at least.

It seems Fedora has some policy about it too. So let’s be a good FLOSS citizen and keep bison requirement *by default* (see hereinafter).

@bluetech By the same reasoning it seems that we should also make the build depend on `gperf` for `keywords.c`.

wismill:
I added the build option ~`enable-parser-generation`~ `enable-parser-auto-generation`:
- If true (default), the build requires bison and auto-generates the parser files.
- Otherwise bison is optional and the build uses pre-generated files.

There is also a manual target `src/xkbcomp/update-parser` to regenerate bundled parser files, which depends on bison availability, but not of the value of `enable-parser-generation`.

wismill:
Rebased, changed the meson option name and added a check in the CI to ensure pre-generated files are up-to-date.

PR #1006, opened 2026-03-25
doc: Use libxkbcommon for explicit library references

### LINKED ISSUE #953: Package name: xkbcommon versus libxkbcommon, including the default archive distributed by the github project here
Hey folks,

This is not a very important topic, but nonetheless I wanted to raise it. I will explain why in a moment.

The issue is:

We can download a stable tarball release, e. g. a .tar.gz file, such as this one here:

https://github.com/xkbcommon/libxkbcommon/archive/refs/tags/xkbcommon-1.12.0.tar.gz
wget https://github.com/xkbcommon/libxkbcommon/archive/refs/tags/xkbcommon-1.12.0.tar.gz

Now, when we extract it, we actually get this directory:

libxkbcommon-xkbcommon-1.12.0

This name is wrong, in my opinion. The best assumption is that the name of the tarball at hand,
becomes the very same name as the extracted directory. So the correct name should be:

    xkbcommon-1.12.0

I assume github does this automatically, but I think it is also possible to change it; I know that
other projects follow this convention of package_name.tar.xz -> package_name, so I think
github prepends the project name to the tarball. I am not sure where this can be changed,
but since other projects manage this, I think it is possible and should be changed, for
simplicity and consistency reasons.

But this is the small thing; I don't think I would have created an issue for this (or perhaps I
did in the past, I am losing track of things I did...)

The thing to me is ... what is the name of the project? Is it "libxkbcommon" or "xkbcommon"?

I believe the old name was xkbcommon. Wasn't it hosted at https://www.x.org/releases/individual/?
I am not entirely certain anymore. The reason I assumed the name was xkbcommon was because
I used to track this program in the past, with a set of scripts; but it seems the name is indeed
libxkbcommon now, so I need to rename it locally. Which of course is trivial; the confusion I have
is more what the name really is. Right now I'll go with "libxkbcommon", but even this is not 
100% consistent with the current README you guys are using. Let me give one example.

You folks write in the README this here:

"An project’s homepage can be found at https://xkbcommon.org/.

xkbcommon is maintained in git at: https://github.com/xkbcommon/libxkbcommon"

So this is interesting, because you write "xkbcommon is maintained" rather than
"libxkbcommon is maintained"; yet the URL ends with "libxkbcommon".

The URL at https://github.com/xkbcommon/ only has two entries - one is for
the homepage, the other one is for "libxkbcommon". I guess the name of the
library is "libxkbcommon", but the project name is xkbcommon. But would it
then not make more sense to write e. g.:

"libxkbcommon is maintained in git at: https://github.com/xkbcommon/libxkbcommon"

?

Anyway. I just wanted to point this out. I am fine either way, but the less confusion
the better, IMO. I think I first tracked xbkcommon many years ago, so I am not
always 100% up to date with upstream changes. Cheers!

wismill:
# Context

“xkbcommon” and “libxkbcommon” are both used throughout the code base to refer loosely to the library, which may bring some confusion (see #953).

The meta-project is *xkbcommon* and has currently two projects:
- 3 libraries: 
  - *libxkbcommon* (the (core) “xkbcommon library”). This is the historical project.
  - *libxkbcommon-x11*
  - *libxkbregistry*
- a CLI tool: *xkbcli*

I am also thinking to have some `libxkbcommon-<lang>` “official” bindings.

# Changes

- Add a prose naming convention guide. The gist is:
 
  > Default to **xkbcommon**. Use a **lib…** artifact name only when one of these *triggers* applies:
  > 1. **Disambiguation:** a sibling library is actually named in the same context.
  > 2. **Artifact specificity:** the statement is about a compiled artifact,
       `-dev` package, or link flag — something you could point to a file for.
  >
  >     The `lib` prefix marks the artifact's canonical/package identity matching its `.pc` name, not a literal on-disk filename: MSVC builds, for example, produce `.lib` files without the `lib` prefix.
  >
  > Sub-libraries **xkbcommon-x11**, **xkbregistry**  use their bare name in *all* prose; add `lib` only for trigger 2.
- Apply the guide to the code base.

Fixes #953

---

> [!caution]
> @fooishbar @bluetech @whot Is the naming distinction between project and lib actually correct? Digging in the git log it seems _this confusion has been present since the early stages of the project_.

I mean, we (devs) all understand the terms depending on their context, but it matters for third-party and search referencing.


whot:
I think @fooishbar might have more info, in my head it was always libxbkcommon and i have to shift gear to think of it as xkbcommon - though that's the URL etc.

wismill:
> I’d been working on a library called [xkbcommon](https://xkbcommon.org/).

etc. Source: https://www.fooishbar.org/blog/xkbcommon-intro/

fooishbar:
I’d always personally thought of the name as xkbcommon, but a) I sort of inherited it from Dan Nicholson, and b) this project hasn’t been my project for over 10 years now!

wismill:
I went ahead and mass-renamed libxkbcommon → xkbcommon, libxkbcommon-x11 → xkbcommon-x11, and libxkbregistry → xkbregistry, with a few exceptions for disambiguation and artifact names.

Please note that the new prose naming conventions are designed to enhance the project’s public communication. Let’s stay disciplined with external materials (release notes, documentation, and announcements), but feel free to use whichever naming you are used to in other contexts.

PR #1017, opened 2026-04-16
registry: Add DTD error logging for libxml2 < 2.14

wismill:
Fixed missing logging of libxml2 errors when validating the DTD for libxml2 < 2.14.

Relates-to: https://bugs.kde.org/show_bug.cgi?id=518963

@ariasuni

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -34,7 +34,7 @@ jobs:
         shell: cmd
         run: |
           :: Jinja2 is needed for merge-modes tests
-          python -m pip install --upgrade meson Jinja2
+          python -m pip install --upgrade "meson!=1.11.0" Jinja2
           choco install -y winflexbison3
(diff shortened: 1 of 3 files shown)


ariasuni:
I have the problem you’re referring to with libxml2 2.15.2-1 (Arch Linux).

wismill:
@ariasuni Have you recompiled libxkbcommon with the libxml2-2.15 headers? Fix requires recompilation to take effect.

wismill:
@ariasuni OK I found the issue that triggers on your system and not mine: it is about include paths in the header probing.

The `libxml2-devel` package of openSUSE Tumbleweed provides the symlink `/usr/include/libxml` → `/usr/include/libxml2/libxml` while ArchLinux (and Ubuntu) do not.

Fixed by specifying dependency properly in the header probing in `meson.build`.

PR #1068, opened 2026-07-26
doc: Create predefined keysyms page

wismill:
Since the keysyms header has a fixed syntax, use a Python script to dynamically transform the C header into a Doxygen-annotated file.


wismill:
New documentation pages are live:
- [keysym encoding](https://xkbcommon.org/doc/1.14.0/group__keysym-encoding.html)
- [predefined keysyms](https://xkbcommon.org/doc/1.14.0/group__predefined-keysyms.html)

PR #1075, opened 2026-08-05
Typo "Real ... " -> "Virtual modifier map" Update keymap-text-format-…

krei-se:
…v1-v2.md

"Real modifier map" was copypaste error twice


wismill:
Good catch, thanks!

krei-se:
Good software and documentation, thanks!

PR #1077, opened 2026-08-12
license: Complete and correct SPDX notices

ytausch:
## Summary

- add the ISC license notice from `src/utils-checked-arithmetic.h` to the top-level `LICENSE` file
- identify the Digital Equipment Corporation notice as SPDX `HPND` in `src/atom.c` and `LICENSE`
- simplify the file-level expression to `MIT-open-group AND HPND`

## Rationale

`src/utils-checked-arithmetic.h` was added in fb9fec18 with an ISC SPDX identifier and the full ISC notice. The top-level `LICENSE` says it lists all license statements in the source tree, but it was not updated when that file was introduced.

The Digital Equipment Corporation notice matches the [SPDX HPND](https://spdx.org/licenses/HPND.html) template: the copyright holder is replaceable, the no-endorsement clause is permitted, and the suitability / as-is paragraph is optional. Therefore the custom `LicenseRef-digital-equipment-corporation` identifier is unnecessary.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -164,8 +164,8 @@ OR PERFORMANCE OF THIS SOFTWARE.
 
 
 -------------------------------------------------------------------------------
-SPDX full name: <none>
-SPDX short identifier: LicenseRef-digital-equipment-corporation
+SPDX full name: Historical Permission Notice and Disclaimer
+SPDX short identifier: HPND
 License text:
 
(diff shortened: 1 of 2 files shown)


ytausch:
@wismill gentle ping

bluetech (approved):
Looks right, thanks!

